PR #1, opened 2018-10-04
Make IRB::VERSION available for library users

stomar:
`IRB::VERSION` is defined and used in the gemspec, but not available for irb users. The PR makes it available in irb sessions.

_Edit:_ The build fail seems to be unrelated.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -19,6 +19,8 @@ require "irb/ruby-lex"
 require "irb/input-method"
 require "irb/locale"
 
+require "irb/version"
+
 # IRB stands for "interactive Ruby" and is a tool to interactively execute Ruby
 # expressions read from the standard input.
 #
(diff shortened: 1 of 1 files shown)


marcandre:
Seems like a good idea. Would be a good idea to replace those `require` with `require_relative` too.

stomar:
I believe `require` should be preferred, because `require_relative` doesn‘t properly detect multiple requires, afaik. Also, there is no need for it, because the lib dir is in the load path anyway for an installed gem.

marcandre:
AFAIK, `require_relative` does detect multiple requires.
It's also faster because it doesn't have to lookup the load path...

stomar:
Honestly, I do not _know_... possibly/probably(?) outdated information.

But anyway I think this would best be considered in a different PR or on Redmine. Your suggestion would affect most of the standard libraries and extensions in Ruby trunk, I guess.

marcandre:
> Your suggestion would apply to most of the standard libraries and extensions in Ruby trunk, I guess.

Sure, I'll have a look.

marcandre:
You probably want to remove the useless `require 'irb/lib'` on line 357

marcandre:
@stomar Maybe [this is what you vaguely remembered](https://github.com/rubygems/rubygems/pull/2425#issuecomment-427535667) :-)

stomar:
@marcandre Thanks for the reference, seems I got it mixed up :)

PR #5, opened 2018-12-26
Remove development files in gem releases

colby-swandale:
This PR is like https://github.com/ruby/fileutils/pull/27 and removes files that are mostly used for development.


colby-swandale:
i will look into why the test suite is failing

PR #8, opened 2018-12-27
Fix breaking specs

colby-swandale:
the test suite is currently only working on Ruby 2.6 because of a
new method on Binding that older versions of Ruby do not have.

This PR fixes the test suite by checking for the method and using the
previous method if it's not available.


colby-swandale:
I will investigate the ruby-head error

colby-swandale:
This should be good to merge now.

PR #10, opened 2019-01-01
fix breaking workspace test on ruby 2.3

colby-swandale:
The test suite is currently not passing on Ruby 2.3 due to `Tempfile.create` requiring an argument. See https://travis-ci.com/ruby/irb/jobs/167607309 for the log, Later Ruby versions do not require any arguments.

This PR fixes the outstanding issue and should have tests passing on all versions 🎉.


yuki24:
Thanks!

PR #11, opened 2019-03-20
Don't echo results of assignment expressions

onlynone:
This is a proof of concept for something I'd like to see added to irb. The ability for the interpreter to not echo results of assignment expressions.

I'm guessing this exact patch wouldn't be accepted because:

1. You won't want to pull in another gem, and
2. You might want the semantics of the `IRB.conf[:ECHO]` setting to stay the same (ie `true` always echos and `false` never echos)

As to 1: I'm guessing irb and ruby have other ways of doing expression parsing that maybe could be hooked into instead of using the `parser` gem. I'm just not very familiar with those possible methods, and I found `parser` easy to use to express the functionality I was looking for.

For 2: Would a third value other than `true` or `false` for that setting be acceptable? Like `:non_assignment_expressions`? Or maybe another setting altogether: `IRB.conf[:NO_ECHO_ON_ASSIGNMENT]`? So that only when `IRB.conf[:ECHO]` and `IRB.conf[:NO_ECHO_ON_ASSIGNMENT]` are `true` would the advanced echo/noecho logic be used.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,5 @@ source "https://rubygems.org"
 git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
 
 gemspec
+
+gem "parser", "~> 2.6"
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -392,16 +392,27 @@ module IRB
       # whether to echo this line or not
       if @original_echo
         node = Parser::CurrentRuby.parse(line)
-        if node.type.to_s.end_with?('asgn')
-          # Node types that end with 'asgn':
-          # lvasgn: local variable assignment: "foo = bar"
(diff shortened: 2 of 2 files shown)


onlynone:
See https://stackoverflow.com/questions/55128369/prevent-ruby-rails-from-printing-an-object-in-the-console-in-certain-situations for a justification.

hsbt:
We couldn't use the 3rd party gem like `parser`. Because irb is the part of a standard library of ruby language.

onlynone:
Hi @hsbt. I understand not being able to use that third party dependency. That was just a starting point so I could show the intended behavior. I was hoping someone more knowledgeable about ruby and irb internals would be able to point me to another, more appropriate, internal, AST parser that I could use instead of the `parser` gem.

I was hoping that my first implementation would spark discussion of how to properly do this.

PR #12, opened 2019-03-21
Dont echo assignment expressions

onlynone:
Same idea as #11 but it uses ~`RubyVM::AbstractSyntaxTree`~ `ripper` rather than the `parser` gem. ~It would be nice to have a solution that's back-portable to ruby versions prior to 2.6 though.~ I'm open to any ideas.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -10,6 +10,7 @@
 #
 #
 require "e2mmap"
+require "ripper"
 
 require "irb/init"
 require "irb/context"
@@ -410,6 +411,35 @@ module IRB
   end
 
   class Irb
+    ASSIGNMENT_NODE_TYPES = [
+      # Local, instance, global, class, constant, instance, and index assignment:
+      #   "foo = bar",
+      #   "@foo = bar",
+      #   "$foo = bar",
+      #   "@@foo = bar",
+      #   "::Foo = bar",
+      #   "a::Foo = bar",
+      #   "Foo = bar"
+      #   "foo.bar = 1"
+      #   "foo[1] = bar"
+      :assign,
+
+      # Operation assignment:
+      #   "foo += bar"
+      #   "foo -= bar"
+      #   "foo ||= bar"
+      #   "foo &&= bar"
+      :opassign,
+
+      # Multiple assignment:
+      #   "foo, bar = 1, 2
+      :massign,
+    ]
+    # Note: instance and index assignment expressions could also be written like:
+    # "foo.bar=(1)" and "foo.[]=(1, bar)", when expressed that way, the former
+    # be parsed as :assign and echo will be suppressed, but the latter is
+    # parsed as a :method_add_arg and the output won't be suppressed
+
     # Creates a new irb session
     def initialize(workspace = nil, input_method = nil, output_method = nil)
       @context = Context.new(self, workspace, input_method, output_method)
@@ -494,7 +524,7 @@ module IRB
           begin
             line.untaint
             @context.evaluate(line, line_no, exception: exc)
-            output_value if @context.echo?
+            output_value if @context.echo? && (@context.echo_on_assignment? || !assignment_expression?(line))
           rescue Interrupt => exc
           rescue SystemExit, SignalException
             raise
@@ -703,6 +733,18 @@ module IRB
         end
       end
       format("#<%s: %s>", self.class, ary.join(", "))
+    end
+
+    def assignment_expression?(line)
+      # Try to parse the line and check if the last of possibly multiple
+      # expressions is an assignment type.
+
+      # If the expression is invalid, Ripper.sexp should return nil which will
+      # result in false being returned. Any valid expression should return an
+      # s-expression where the second selement of the top level array is an
+      # array of parsed expressions. The first element of each expression is the
+      # expression's type.
+      ASSIGNMENT_NODE_TYPES.include?(Ripper.sexp(line)&.dig(1,-1,0))
     end
 
     ATTR_TTY = "\e[%sm"
--- a/test/irb/test_raise_no_backtrace_exception.rb
+++ b/test/irb/test_raise_no_backtrace_exception.rb
@@ -7,6 +7,7 @@ module TestIRB
       bundle_exec = ENV.key?('BUNDLE_GEMFILE') ? ['-rbundler/setup'] : []
       assert_in_out_err(bundle_exec + %w[-rirb -W0 -e IRB.start(__FILE__) -- -f --], <<-IRB, /Exception: foo/, [])
       e = Exception.new("foo")
+      puts e.inspect
       def e.backtrace; nil; end
       raise e
 IRB
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -121,6 +121,11 @@ module IRB
       if @echo.nil?
         @echo = true
       end
+
+      @echo_on_assignment = IRB.conf[:ECHO_ON_ASSIGNMENT]
+      if @echo_on_assignment.nil?
+        @echo_on_assignment = false
+      end
     end
 
     # The top-level workspace, see WorkSpace#main
@@ -236,6 +241,15 @@ module IRB
     #     puts "omg"
     #     # omg
     attr_accessor :echo
+    # Whether to echo for assignment expressions
+    #
+    # Uses IRB.conf[:ECHO_ON_ASSIGNMENT] if available, or defaults to +false+.
+    #
+    #     a = "omg"
+    #     IRB.CurrentContext.echo_on_assignment = true
+    #     a = "omg"
+    #     #=> omg
+    attr_accessor :echo_on_assignment
     # Whether verbose messages are displayed or not.
     #
     # A copy of the default <code>IRB.conf[:VERBOSE]</code>
@@ -261,6 +275,7 @@ module IRB
     alias ignore_sigint? ignore_sigint
     alias ignore_eof? ignore_eof
     alias echo? echo
+    alias echo_on_assignment? echo_on_assignment
 
     # Returns whether messages are displayed or not.
     def verbose?
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -51,6 +51,7 @@ module IRB # :nodoc:
     @CONF[:IGNORE_SIGINT] = true
     @CONF[:IGNORE_EOF] = false
     @CONF[:ECHO] = nil
+    @CONF[:ECHO_ON_ASSIGNMENT] = nil
     @CONF[:VERBOSE] = nil
 
     @CONF[:EVAL_HISTORY] = nil
@@ -172,6 +173,10 @@ module IRB # :nodoc:
         @CONF[:ECHO] = true
       when "--noecho"
         @CONF[:ECHO] = false
+      when "--echo-on-assignment"
+        @CONF[:ECHO_ON_ASSIGNMENT] = true
+      when "--noecho-on-assignment"
+        @CONF[:ECHO_ON_ASSIGNMENT] = false
       when "--verbose"
         @CONF[:VERBOSE] = true
       when "--noverbose"
--- a/test/irb/test_context.rb
+++ b/test/irb/test_context.rb
@@ -26,6 +26,10 @@ module TestIRB
       def encoding
         Encoding.default_external
       end
+
+      def reset
+        @line_no = 0
+      end
     end
 
     def setup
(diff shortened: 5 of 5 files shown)


ioquatix:
I think this is a really awesome idea.

onlynone:
@hsbt Is this a better implementation than #11 ? I guessed on what the new configuration setting should be called (`echo_on_assignment` which defaults to `false`), but I'd still like to hear any feedback on it.

I also wrote it so that the echoing isn't suppressed when running ruby < 2.6.0 because `RubyVM::AbstractSyntaxTree` isn't available until then. If there's another library or technique I could use to figure out the expression type that would work for earlier ruby versions, I'd love to know.

Could we do something like not require the `parser` gem, but if it's available and `RubyVM::AbstractSyntaxTree` is not, then try and use it?

eregon:
Anything under `RubyVM` is not portable and CRuby-specific.
Would using Ripper works? That's a fairly portable standard library.

onlynone:
@eregon Thanks for the tip about Ripper. Just updated the patch to use Ripper instead and removed the restriction for 2.6.0.

hsbt:
@onlynone Can you add test for this change?

lib/irb.rb
```diff
@@ -702,6 +732,14 @@ def inspect
       format("#<%s: %s>", self.class, ary.join(", "))
     end
 
+    def is_assignment_expression?(line)
```
nobu: `is_` seems redundant with `?` suffix.

lib/irb.rb
```diff
@@ -702,6 +732,14 @@ def inspect
...
+      s = Ripper.sexp(line)
+      return false if !s.is_a?(Array)
+      return false if !s[1].is_a?(Array)
+      return false if !s[1][0].is_a?(Array)
+      ASSIGNMENT_NODE_TYPES.include?(s[1][0][0])
```
nobu: It doesn't work for multi-statement, e.g.,`nil; a = 0`.
And you can use `Array#dig`:
```ruby
ASSIGNMENT_NODE_TYPES.include?(Ripper.sexp(line)&.dig(1, -1, 0))
```

onlynone: I realize it doesn't work for something like `nil; a = 0`. I wasn't necessarily trying to solve it for all cases, just most.

The only problem with [`Array.dig`](https://ruby-doc.org/core-2.3.0/Array.html#method-i-dig) is that if any intermediate element isn't an array (or has its own `dig` method) it raises an error. I wasn't sure if I could rely on the object structure returned from `Ripper.sexp` always being an array whose second element is an array whose first element is an array

onlynone:
> @onlynone Can you add test for this change?

Tests added.

lib/irb.rb
onlynone: removed `is_`

lib/irb.rb
onlynone: The logic now looks at the last expression in the array. I still wasn't 100% confident that I could rely on everything being arrays, so I didn't change to `dig`. If it's a guarantee that any non-nil value returned by `Ripper.sexp` can be dug with 3 integer arguments, then I'd make the change.

nobu: An S-expression is a list in common, and a list is an Array in Ruby.
Even if it would change, it should be an array-like object.
Otherwise you can't access the contents using `[]`.

onlynone: The point is, I don't try to access it with `[]` until after I check that it's an array. I know that at the top level it's an array, and it's all nested arrays, but how do I know that I won't run into a terminal element at some point in the middle of the dig recursion? For example, this would fail: `[1, 2, 3].dig(1, -1, 0)`.

nobu: The outline of Ripper S-exp hasn't change for a long time.
If the structure changes, the code should have to change too.

Another way is massign.
```ruby
_, (*, (e,)) = Ripper.sexp(line)
ASSIGNMENT_NODE_TYPES.include?(e)
```

onlynone: > The outline of Ripper S-exp hasn't change for a long time.
If the structure changes, the code should have to change too.

Right, but I don't know what that structure is, and if the current structure is guaranteed to always have an array for the last element of the second element of the array returned by `Ripper.sexp` for every possible valid ruby expression. It holds for every valid expression I've tried, but I don't know if what I tried was exhaustive.

If you say it is guaranteed at the current time, then yeah, I think using `&.dig(1,-1,0)` is good.

onlynone: Using an `massign` seems more forgiving than `dig` when the structure doesn't match what one expects. It seems to assign `nil` to variables in cases that `dig` would raise errors. But the code looks more confusing.

If we can be certain that `Ripper.sexp(line)&.dig(1,-1,0)` will never raise an error, then I'm fine going with that. If it's possible that it might raise an error, I'd prefer to leave to code as is so it's more obvious what's going on rather than going with the `massign` expression.

stomar:
For `:opassign` echoing the assigned value might be useful, IMO.

```
# `option` might have been set earlier

>> option ||= "default"
=> "value"
```

lib/irb.rb
nobu: The results of ripper will be kept compatible as possible, but it is tough to guarantee.
Although this part would change rarely, I understand your concern.
Feel free to choice the method.

onlynone: I'll use the dig. I used a large Ruby code base as a source of valid Ruby expressions, and every one resulted in a sexp that matched the structure needed by that call to dig.

For the record, I wasn't worried about the sexp structure changing in later versions. I was worried that there might be valid expressions for which the *current* version of ripper would return structures that wouldn't work for the exact arguments to dig from above.

onlynone:
@stomar it sounds like you'd want to set `context.echo_on_assignment = true`:

```
>> context.echo_on_assignment = true
=> true
>> option ||= "default"
=> "default"
```

or `puts option ||= "default"`:

```
>> puts option ||= "default"
default
=> nil
```

or just type `option` at the prompt again:

```
>> option ||= "default"
>> option
=> "default"
```

onlynone:
Would it be better if the default for `echo_on_assignment` was `true` so that no behavior changes unless one explicitly sets `IRB.conf[:ECHO_ON_ASSIGNMENT] = false` or uses `--noecho-on-assignment` ? I certainly prefer to suppress the echo of assignment expressions, but I can understand if we don't want behavior to change by default.

onlynone:
@nobu just checking in to see if there's anything else that can be done.

nobu:
Seems @aycabta working on the overhaul of irb now, stay tuned plz.

onlynone:
Didn't @k0kubun just commit a bunch of stuff? Also, it looks like a lot of that came from ruby/ruby. Should I be making a PR there first?

k0kubun:
Making changes to ruby/ruby first was my mistake. You're filing PR to the perfect place.

I'm just a random contributor. Ask @aycabta about what's going on around irb, or @hsbt for how we should make changes for defaut gems (I commented it in the above line though).

onlynone:
@nobu It's almost been a month since you asked me to stay tuned, is there any update? @aycabta how's the overhaul going?

onlynone:
@nobu @hsbt are @aycabta's changes all set now? I've rebased on top of master and fixed conflicts. The only failing test seems unrelated to my changes.

onlynone:
Failing test was actually related... fixed and squashed all commits.

onlynone:
@nobu @hsbt and @aycabta, just checking in again. It seem like commits have been going in to master again. Any reason to keep holding off?

test/irb/test_raise_no_backtrace_exception.rb
```diff
@@ -7,6 +7,7 @@ def test_raise_exception
       bundle_exec = ENV.key?('BUNDLE_GEMFILE') ? ['-rbundler/setup'] : []
       assert_in_out_err(bundle_exec + %w[-rirb -W0 -e IRB.start(__FILE__) -- -f --], <<-IRB, /Exception: foo/, [])
       e = Exception.new("foo")
+      puts e.inspect
```
kares: :scissors:

onlynone: That output is what the test is testing for. Without the `puts` the test will fail.

lib/irb.rb
```diff
@@ -705,6 +735,18 @@ def inspect
...
+    def assignment_expression?(line)
+      # Try to parse the line and check if the last of possibly multiple
+      # expressions is an assignment type.
+
+      # If the expression is invalid, Ripper.sexp should return nil which will
+      # result in false being returned. Any valid expression should return an
+      # s-expression where the second selement of the top level array is an
+      # array of parsed expressions. The first element of each expression is the
+      # expression's type.
+      ASSIGNMENT_NODE_TYPES.include?(Ripper.sexp(line)&.dig(1,-1,0))
```
aycabta: Please use `Ripper.lex` because `Ripper.sexp` is not maintained.

nobu: I don't think that `Ripper.sexp` is less maintained than `Ripper.lex`, and vice versa.

JorgenIvarsson:
Hi! As a new-be on ruby on rails I feel that something has gone seriously wrong here. I have just searched the Internet, asked in forum for 24 hours, before finding the reason that irb does not echo the object I am creating, while following guides on learning Rails. After testing if this new behavior was caused by the switch from bash to zsh (on Mac), my rails configuration and so on, I finally got an answer on the Rails forum that from version 2.7, Ruby no longer echoes the return value when creating and assigning a new object. I also learned from another user that there is a configuration file for irb. And finally I found the source for this new behavior on this github page. 

As you might expect I do not like this change. For me, who is just learning both Ruby and Rails at the same time, I really appreciated the old behavior, when irb echoed the return value when I create a new object. 
I gess you pros have forgotten how it is to learn a new language and a new developer environment?
What makes me sad is that @onlynone in one of the posts above asks: 

_"Would it be better if the default for echo_on_assignment was true so that no behavior changes unless one explicitly sets IRB.conf[:ECHO_ON_ASSIGNMENT] = false or uses --noecho-on-assignment ? I certainly prefer to suppress the echo of assignment expressions, but I can understand if we don't want behavior to change by default."_ 

Spot on! As a new-be I can't argue against this new feature (from what I read you want Ruby more like Python-like, not echoing out hundreds of lines of code. My argument for letting the default value be `RB.conf[:ECHO_ON_ASSIGNMENT] = true` is that for a new-be like me, this output is very valuable. And it is really hard for a new-be to understand how to change this behavior in a configuration file that does not exist (I had to create a .irbrc file in my home directory). 
Most new-bees will never know that you actually can change the behavior of irb. 

But for a developer proficient in Ruby irritated over lines of echoed code, could probably with ease change the behavior to `RB.conf[:ECHO_ON_ASSIGNMENT] = false`, if he or she so preferred. 

So for me this is a pedagogical question. What should be the default behavior? And how will it effect the users? A new user just learning ruby? An experienced programmer? 

Please set the default value back to true! So you restore the previous behavior and help all new users that are learning Ruby (or Rails). I any user in time want to change the behavior of irb down the road, that person are experienced and most likely also know how to change these settings.

gdonald:
It's spelled "newbie".  https://www.merriam-webster.com/dictionary/newbie

JorgenIvarsson:
@gdonald Thank you Greg for correcting my spelling.

onlynone:
> My argument for letting the default value be RB.conf[:ECHO_ON_ASSIGNMENT] = true is that for a newbie like me, this output is very valuable. And it is really hard for a newbie to understand how to change this behavior in a configuration file that does not exist

If seeing the value from the return of a function is valuable, you don't have to look under the hood of irb settings to see it. You can just type the name of the variable you assigned the value at the prompt after the function call.

This change only comes into effect when you're doing an assignment like `a = new_thing()`. If you just type `new_thing()`, the returned value will still be echoed to the terminal. If you *are* executing an assignment statement, then by definition, you now have a reference to the thing you created, so you can just type `a` on the next line to see the value.

JorgenIvarsson:
> > My argument for letting the default value be RB.conf[:ECHO_ON_ASSIGNMENT] = true is that for a newbie like me, this output is very valuable. And it is really hard for a newbie to understand how to change this behavior in a configuration file that does not exist
> 
> If seeing the value from the return of a function is valuable, you don't have to look under the hood of irb settings to see it. You can just type the name of the variable you assigned the value at the prompt after the function call.
> 
> This change only comes into effect when you're doing an assignment like `a = new_thing()`. If you just type `new_thing()`, the returned value will still be echoed to the terminal. If you _are_ executing an assignment statement, then by definition, you now have a reference to the thing you created, so you can just type `a` on the next line to see the value.

Yes, I discovered that calls to the object echoed the value within minutes. The problem was that this behavior did not match the online guide I was following. This is a new behavior in Ruby from version 2.7. You have therefore, without warning, changed a behavior that has existed for many, many years. Of course, it can be argued that this changed behavior, in time, will be reflected in guides when they are being updated. 

I would like to argue that this change was not asked for. If I read the original case (https://github.com/ruby/irb/pull/11) there was a requested for a setting to be able to turn off the echo for assignments. That request is fulfilled by changing the configuration `IRB.conf[:ECHO_ON_ASSIGNMENT] = false.`

But instead of fulfilling that original request, you have changed a default behavior, so that all users of Ruby/Rails are affected by this new default value. I did not understand that there was such a request? If there is a general and strong desire to change this default setting in the ruby community, I must have missed it. If you can show that there is a strong request for this change of the default behavior, I will have to accept this change. 
But if there isn’t a widespread request for this change, I would argue that this change is a setback for mainly new users of Ruby/Rails.

Another aspect of this is consistency. An important mantra for Rails is "Sensible defaults". Due to the changes in Ruby, we now get an incomprehensible and inconsistent behavior in the IRB. Previously, the rule was that you always got a return value echoed. Now we have a situation where calling a variable echo the value. When calling “new “, you do not get an echo of the return value. But If you call destroy, you get an echo of the return value in irb. 

Again, I argue that as an experienced developer you can easily change these settings to your liking. But not so easy for the one who is just learning Rails. New developers will not know that they can change this behavior in a setting. As it now stands, this change doesn’t seem to be documented at all. 

So instead of fulfilling the original request (to be able to turn off the default behavior), you have changed the general behavior for all users of Ruby/Ralls. 
My humble opinion is that this was not the right way to do it. Please change it back.

mtgrosser:
Please change the default back to what it has (always?) been. This behaviour change is totally unexpected and irritating, and I'm saying this as a longtime Rubyist.

In Ruby, everything evaluates to something. Till now, the philosophy of IRB adhered exactly to that idea. Hiding the value of an assignment is not POLS.

aycabta:
I changed the behavior by https://github.com/ruby/irb/pull/129.

> Omit the results evaluated at assignment if they are too long.
> 
> The behavior of `ECHO_ON_ASSIGNMENT` being on by default is hard to understand, so I change it to off by default. Instead, we turn `OMIT_ON_ASSIGNMENT` on by default. The result is displayed on assignment, but it will always be short and within one line of the screen.

aycabta:
I created a concise Pull Request #130. (ref. https://github.com/ruby/irb/pull/129#issuecomment-694917004)

> Drop OMIT_ON_ASSIGNMENT and add :truncate option for ECHO_ON_ASSIGNMEN

And released 1.2.7.

PR #17, opened 2019-05-26
Add support for history with Reline backend

kou:


ioquatix:
Do you think it's good idea for `HISTORY` to be global state?

kou:
What do you mean?
Do you want to keep history for each session instead of sharing one history in all sessions?

ioquatix:
I just briefly looked at code and using a shared global state `HISTORY` gave me concern because it's mutable global state.

For example, it's tricky for unit test because there is global state. It might also break in multi-thread environment.

I just wondered if there is better approach.

kou:
I understand.
I think that we don't need to care about it for now.

Because the `HISTORY` in the original code is provided by readline (`Readline::HISTORY`). It accesses to the process global history in libreadline. So our history is needed to be process global.

For Reline backend, `HISTORY` is `Reline::HISTORY`. It's also the process global history.

If we need to reduce scope of `HISTORY`, we need to reduce scope of backend history (the history in libreadline and the history in Reline) at first.

PR #20, opened 2019-07-26
Use JRuby equivalent of RubyVM.compile

headius:
RubyVM is specific to CRuby and not supported on JRuby. This is
the equivalent operation.


headius:
Ideally there would be an `else` for this case that falls back on something other implementations can support like Ripper, but I don't know the Ripper API well enough to make that change.

aycabta:
Works fine! Thank you for your professional patch!

PR #23, opened 2019-08-25
Move private call without arguments inside method

jeremyevans:
This code did not have the desired effect.  I'm not sure if
irb_binding is supposed to be private or not.  If not, the private
call can just be removed.


aycabta:
I found the change at 1eb232cbfc19bfcedf05ad6d6d012e2907495322. Maybe, it was just a misstake.

PR #53, opened 2019-12-24
Fix typo

stomar:


stomar:
Or should I fix directly in Ruby upstream repository?

BTW. I had an older clone of this repo, but could not fast-forward to the current head. Seems nearly all commits since about 2001 have been rebased?

aycabta:
Good catch!

This repository was losing many commits from Ruby upstream repository. So I fixed it by rebasing. Sorry for the confusing situation.

Issue #55, opened 2019-12-26
Crashes with complex autoindent


aycabta:
Reported by @mame:

```ruby
(((
)
)
```

[Reported at #16453](https://bugs.ruby-lang.org/issues/16453) by @JunichiIto

```ruby
[<<FOO]
hello
FOO
```

BenKanouse:
I attempted to fix this issue here: https://github.com/ruby/irb/pull/56

Please let me know if you have any feedback.

Thanks!

BenKanouse:
I believe that this PR fixes both issues: https://github.com/ruby/irb/pull/58

It also adds test for both issues. I'm looking forward to any feedback. Thanks!

aycabta:
This is fixed by https://github.com/ruby/irb/pull/58.

Thank you, @BenKanouse!

PR #56, opened 2019-12-27
Fix for crashes with complex auto indent

BenKanouse:
Fix for https://github.com/ruby/irb/issues/55

The issue was that irb would crash if we used 3 braces in a row or used <<FOO syntax inside of braces.

Examples:

```
zerg:irb Ben$ irb
2.7.0 :001 > [[[
2.7.0 :002 >     ]
2.7.0 :003 >   ]
2.7.0 :004 > ]
```

```
zerg:irb Ben$ irb
2.7.0 :001"> [<<TEST]
2.7.0 :002"> foo
2.7.0 :003 > TEST
 => ["foo\n"]
```

This is my first time working on this project. I tried to make good commit messages to explain what I did.

Please let me know if there are any changes I should make, I would be happy to make them.

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -314,18 +314,26 @@ class RubyLex
 
   def check_newline_depth_difference
     depth_difference = 0
+    skip_til_newline = false
     @tokens.each_with_index do |t, index|
       case t[1]
       when :on_ignored_nl, :on_nl, :on_comment
+        skip_til_newline = false
         next
-      when :on_sp
+      when :on_sp || skip_til_newline
         next
       end
+
+      next if skip_til_newline
+
       case t[1]
       when :on_lbracket, :on_lbrace, :on_lparen
         depth_difference += 1
       when :on_rbracket, :on_rbrace, :on_rparen
         depth_difference -= 1
+      when :on_heredoc_beg
+        depth_difference = 0
+        skip_til_newline = true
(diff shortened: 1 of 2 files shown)


ioquatix:
Nice work.

aycabta:
Thank you for your contribution! Now, IRB doesn't have any tests, so your testing code is very great to be helpful.

The logic of `set_auto_indent` and `check_newline_depth_difference` what I wrote should process a complex code of #74 correctly. Unfortunately, this Pull Request breaks an autoindent behavior of #74.

Could you add a test for the complex code of #74 first by other Pull Request? I think the test needs a here-doc string literal that contains the complex code of #74 to read easily.

BenKanouse:
Thank you for the feedback, and good catch!

Here is a screenshot of this branch vs master with the code that you linked:
Screen Shot 2019-12-27 at 10 29 23 AM

Seems that this branch is adding some extra un-wanted spacing. As you suggested, I'll try to test that behavior on a new branch so that we do not accidentally break it.

Thanks again.

BenKanouse:
I created a pull request with tests for the code in https://github.com/ruby/reline/issues/74.

Link to pull request: https://github.com/ruby/irb/pull/57

I tested the changes in this pull request against those new tests and I have confirmed that this branch is adding too many spaces.

I'm closing the pull request since it doesn't fully address the problem.

PR #57, opened 2019-12-27
Add tests for RubyLex

BenKanouse:
The set_auto_indent method calculates the correct number of spaces for
indenting a line. We think there might be a few bugs in this method so
we are testing the current functionality to make sure nothing breaks
when we address those bugs.

Example test failure:

```
  1) Failure:
TestIRB::TestRubyLex#test_auto_indent [/Users/Ben/Projects/irb/test/irb/test_ruby_lex.rb:75]:
Calculated the wrong number of spaces for:
 def each_top_level_statement
  initialize_input
  catch(:TERM_INPUT) do
    loop do
      begin
        prompt
        unless l = lex
          throw :TERM_INPUT if @line == ''
        else
.
<10> expected but was
<12>.
```

### DIFF
--- a/test/irb/test_ruby_lex.rb
+++ b/test/irb/test_ruby_lex.rb
@@ -1,0 +1,79 @@
+require 'test/unit'
+
+module TestIRB
+  class TestRubyLex < Test::Unit::TestCase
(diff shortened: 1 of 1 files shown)


aycabta:
Awesome! Thank you! And I'll fix #55 later not to break these tests.

PR #58, opened 2019-12-30
Fix crashing when multiple braces per line

BenKanouse:
https://github.com/ruby/irb/issues/55

There are two issues in this commit.

Issue 1 multiple open braces on a line:

The first issue is that when we put multiple opening braces on a line the
spaces_of_nest array keeps getting '0' added to it. This means that when
we pop off of this array we are saying that we should be in position zero
for the next line. This is an issue because we don't always want to be
in position 0 after a closing brace.

Example:
```
[[[
]
]
]
```
In the above example the 'spaces_of_nest' array looks like this after
the first line is entered: [0,0,0]. We really want to be indented 4
spaces for the 1st closing brace 2 for the 2nd and 0 for the 3rd. i.e.
we want it to be: [0,2,4].

Issue 2:

We were subtracting from the depth in check_newline_depth_difference on
every open brace. This is the right thing to do if the opening and
closing brace are on the same line. For example in a method definition we
have an opening and closing parentheses we want to add 1 to our depth,
and then remove it.

```
def foo()
end
```

However this isn't the correct behavior when the brace spans multiple
lines. If a brace spans multiple lines we don't want to subtract from
check_newline_depth_difference and we want to treat the braces the same
way as we do `end` and allow check_corresponding_token_depth to pop the
correct depth.

Example of bad behavior:

```
def foo()
  [
  ]
puts 'bar'
end
```

Example of desired behavior:

```
def foo()
  [
  ]
  puts 'bar'
end
```

### DIFF
--- a/test/irb/test_ruby_lex.rb
+++ b/test/irb/test_ruby_lex.rb
@@ -90,5 +90,24 @@ module TestIRB
         assert_indenting(lines, row.new_line_spaces, true)
       end
     end
+
+    def test_multiple_braces_in_a_line
+      input_with_correct_indents = [
+        Row.new(%q([[[), nil, 6),
+        Row.new(%q(    ]), 4, 4),
+        Row.new(%q(  ]), 2, 2),
+        Row.new(%q(]), 0, 0),
+        Row.new(%q([<<FOO]), nil, 0),
+        Row.new(%q(hello), nil, 0),
+        Row.new(%q(FOO), nil, 0),
+      ]
+
+      lines = []
+      input_with_correct_indents.each do |row|
+        lines << row.content
+        assert_indenting(lines, row.current_line_spaces, false)
+        assert_indenting(lines, row.new_line_spaces, true)
+      end
+    end
   end
 end
--- a/test/irb/test_ruby_lex.rb
+++ b/test/irb/test_ruby_lex.rb
@@ -70,7 +70,22 @@ module TestIRB
       lines = []
       input_with_correct_indents.each do |row|
         lines << row.content
+        assert_indenting(lines, row.current_line_spaces, false)
+        assert_indenting(lines, row.new_line_spaces, true)
+      end
+    end
 
+    def test_braces_on_thier_own_line
+      input_with_correct_indents = [
+        Row.new(%q(if true), nil, 2),
+        Row.new(%q(  [), nil, 4),
+        Row.new(%q(  ]), 2, 2),
+        Row.new(%q(end), 0, 0),
+      ]
+
+      lines = []
+      input_with_correct_indents.each do |row|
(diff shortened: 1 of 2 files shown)


aycabta:
I mean, your way is great, and I think this Pull Request will improve if the modifications are split.

BenKanouse:
@aycabta Thank you for the review! I split up the first commit, and merged the tests I created in into the correct commit. I do think the change improved the PR.

Please let me know your thoughts.

test/irb/test_ruby_lex.rb
```diff
@@ -65,10 +65,17 @@ def test_auto_indent
         Row.new(%q(    end), 4, 4),
         Row.new(%q(  end), 2, 2),
         Row.new(%q(end), 0, 0),
+        Row.new(%q([[[), nil, 6),
+        Row.new(%q(    ]), 4, 4),
+        Row.new(%q(  ]), 2, 2),
+        Row.new(%q(]), 0, 0),
```
aycabta: Could you add a new test method? I think the test for this commit should be separated from `test_auto_indent`, like `test_multiple_braces_in_a_line`.

```ruby
    def test_multiple_braces_in_a_line
      input_with_correct_indents = [
        Row.new(%q([[[), nil, 6),
        Row.new(%q(    ]), 4, 4),
        Row.new(%q(  ]), 2, 2),
        Row.new(%q(]), 0, 0),
        Row.new(%q([<<FOO]), nil, 0),
        Row.new(%q(hello), nil, 0),
        Row.new(%q(FOO), nil, 0),
      ]

      lines = []
      input_with_correct_indents.each do |row|
        lines << row.content
        assert_indenting(lines, row.current_line_spaces, false)
        assert_indenting(lines, row.new_line_spaces, true)
      end
    end
```

test/irb/test_ruby_lex.rb
```diff
@@ -65,6 +65,10 @@ def test_auto_indent
         Row.new(%q(    end), 4, 4),
         Row.new(%q(  end), 2, 2),
         Row.new(%q(end), 0, 0),
+        Row.new(%q(if true), nil, 2),
+        Row.new(%q(  [), nil, 4),
+        Row.new(%q(  ]), 2, 2),
+        Row.new(%q(end), 0, 0),
```
aycabta: Also, this test should be separated into another test method.

test/irb/test_ruby_lex.rb
BenKanouse: Sounds good to me.

BenKanouse:
@aycabta I updated split the 1 test into 3. I agree that this makes the tests easier to understand.

Thank you for all of the feedback!

PR #93, opened 2020-03-28
Require jruby lib to get JRuby.compile_ir

headius:
For 9.3 we have made internal features of the JRuby module (like `compile_ir` accessible only if you `require 'jruby'`, so that's what this PR is for. The `compile_ir` call errors without this:

```
NoMethodError: undefined method `compile_ir' for JRuby:Module
Did you mean?  Complex
          check_code_block at /Users/headius/projects/jruby/lib/ruby/gems/shared/gems/irb-1.2.1/lib/irb/ruby-lex.rb:210
```

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -11,6 +11,7 @@
 #
 
 require "ripper"
+require "jruby" if RUBY_ENGINE == "jruby"
 
 # :stopdoc:
 class RubyLex


aycabta:
Thank you, this is an important thing.

headius:
Thank you! I look forward to a release (we do not commit irb sources to JRuby's repository now, and only use the gems).

headius:
@aycabta Can we get a release with this change? The gem that's out there no longer works on JRuby master. The fix here will work fine on all JRuby versions.

PR #105, opened 2020-05-18
respect NO_COLOR environment variable

jethrodaniel:
When `NO_COLOR` is set to any non-nil value, output is not colorized.

See https://no-color.org/

### DIFF
--- a/test/irb/test_init.rb
+++ b/test/irb/test_init.rb
@@ -60,6 +60,22 @@ module TestIRB
       ENV["IRBRC"] = backup_irbrc
     end
 
+    def test_no_color_environment_variable
+      orig = ENV['NO_COLOR']
+
+      assert IRB.conf[:USE_COLORIZE]
+
+      ENV['NO_COLOR'] = 'true'
+      IRB.setup(eval("__FILE__"))
+      refute IRB.conf[:USE_COLORIZE]
+
+      ENV['NO_COLOR'] = nil
+      IRB.setup(eval("__FILE__"))
+      assert IRB.conf[:USE_COLORIZE]
+    ensure
+      ENV['NO_COLOR'] = orig
+    end
+
     private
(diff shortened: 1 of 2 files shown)


nobu:
From that site:
> While some developers and users obviously prefer seeing these colors, many users don’t.

Just out of curiosity, how "many users" in percentage?

And are bold and underline included in `NO_COLOR`?

jethrodaniel:
> While some developers and users obviously prefer seeing these colors, many users don’t.

I expect this is a very small amount of people, myself not included. Maybe just one or two crazy people.

> And are bold and underline included in `NO_COLOR`?

My interpretation is that `NO_COLOR` would not prohibit bold, underline, flashing, etc. It should only prohibit color.

Since so few people use it, feel free to close if you think it will just add more maintenance burden.

I only added it after I saw the `--colorize` option, and thought, _does irb know about NO_COLOR?_.

test/irb/test_init.rb
```diff
@@ -60,6 +60,22 @@ def test_rc_file_in_subdir
...
+
+      assert IRB.conf[:USE_COLORIZE]
+
+      ENV['NO_COLOR'] = 'true'
+      IRB.setup(eval("__FILE__"))
+      refute IRB.conf[:USE_COLORIZE]
+
+      ENV['NO_COLOR'] = nil
+      IRB.setup(eval("__FILE__"))
+      assert IRB.conf[:USE_COLORIZE]
```
nobu, line 74: Why `eval` for `__FILE__` in the same context?

```suggestion
      IRB.setup(__FILE__)
      refute IRB.conf[:USE_COLORIZE]

      ENV['NO_COLOR'] = nil
      IRB.setup(__FILE__)
      assert IRB.conf[:USE_COLORIZE]
```

PR #106, opened 2020-06-01
Suppress incomplete coding/encoding magic comment error

aycabta:


lib/irb/color.rb
```diff
@@ -155,27 +155,29 @@ def scan(code, allow_last_error:)
         pos = [1, 0]
 
         verbose, $VERBOSE = $VERBOSE, nil
-        lexer = Ripper::Lexer.new(code)
```
nobu, line 158: How about `lexer = Ripper::Lexer.new(";\n#{code}", nil, 0)`?
Encoding pragma is effective only at the beginning.
And the semicolon and newline will be skipped because the position is before the initial `pos`.

aycabta: This Pull Request contains three fixes about tokenizing or compiling:

- https://github.com/ruby/irb/pull/106/files#diff-41331203a6a8fffee819125982796d0dL158
- https://github.com/ruby/irb/pull/106/files#diff-80fbffd8d3dfc72fb539e0c03a8f0581L79
- https://github.com/ruby/irb/pull/106/files#diff-80fbffd8d3dfc72fb539e0c03a8f0581L213

If the code, `";\n#{code}"` what suppresses the error and starts from line 0, I think it's complex a bit to appear three times. But I think your workaround is simpler than this. I'll consider the way later.

aycabta: https://github.com/ruby/irb/pull/108

PR #124, opened 2020-08-14
Add measure command

aycabta:
You can use `measure` command to check performance in IRB like below:

```
irb(main):001:0> 3
=> 3
irb(main):002:0> measure
TIME is added.
=> nil
irb(main):003:0> 3
processing time: 0.000058s
=> 3
irb(main):004:0> measure :off
=> nil
irb(main):005:0> 3
=> 3
```

You can set `measure :on` by `IRB.conf[:MEASURE] = true` in .irbrc, and, also, set custom performance check method:

```ruby
IRB.conf[:MEASURE_PROC][:CUSTOM] = proc { |context, code, line_no, &block|
  time = Time.now
  result = block.()
  now = Time.now
  puts 'custom processing time: %fs' % (Time.now - time) if IRB.conf[:MEASURE]
  result
}
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -538,7 +538,23 @@ module IRB
         signal_status(:IN_EVAL) do
           begin
             line.untaint if RUBY_VERSION < '2.7'
-            @context.evaluate(line, line_no, exception: exc)
+            if IRB.conf[:MEASURE] && IRB.conf[:MEASURE_CALLBACKS].empty?
+              IRB.set_measure_callback
+            end
+            if IRB.conf[:MEASURE] && !IRB.conf[:MEASURE_CALLBACKS].empty?
+              result = nil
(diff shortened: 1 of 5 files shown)


ko1:
`MEASURE_MODE` like `IRB.conf[:PROMPT_MODE] = :DEFAULT` ? (nil by default disables the measurement)

ko1:
or it can be stack-able (measure time, memory, ...) different from prompt.

aycabta:
Most of the measurements will affect each other. Also, if IRB makes multiple executions for each measuring, code with side effects will not be measured correctly, and execution time will be longer. Therefore, I don't think the stackable way is possible.

aycabta:
I thought about it again, and I think stackable is possible, so I implemented it.

PR #129, opened 2020-09-03
Add OMIT_ON_ASSIGNMENT

aycabta:
Omit the results evaluated at assignment if they are too long.

The behavior of `ECHO_ON_ASSIGNMENT` being on by default is hard to understand, so I change it to off by default. Instead, we turn `OMIT_ON_ASSIGNMENT` on by default. The result is displayed on assignment, but it will always be short and within one line of the screen.

ref. https://github.com/ruby/irb/pull/12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,6 @@ rvm:
 before_install:
   - yes | gem update --system
 
-script: bundle exec rake test
+script:
+  - bundle install
+  - bundle exec rake test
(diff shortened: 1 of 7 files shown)


onlynone:
It might be confusing having the options `ECHO`, `ECHO_ON_ASSIGNMENT`, and `OMIT_ON_ASSIGNMENT`. Would it be simpler to let `ECHO_ON_ASSIGNMENT` by more than just a boolean? Maybe it could be `false`, `true`, or `:truncate`, and it would default to `:truncate`. That way we wouldn't need the extra option.

aycabta:
That’s exactly it! I created a new Pull Request https://github.com/ruby/irb/pull/130.

PR #131, opened 2020-09-22
support more body argument for oneliner method definition

no6v:
```ruby
def i() = 1
def s() = :s
def t() = Time.now
```

are all valid, but

```ruby
$ bin/console -f
irb(main):001:0> def i() = 1
=> :i
irb(main):002:1> def s() = :s
irb(main):003:1*   ^C
irb(main):002:1> def t() = Time.now
irb(main):003:1*   ^C
irb(main):002:0> 
```

It seems that a method body does not always contain `Ripper::EXPR_END`.


aycabta:
Thank you, great!

Issue #133, opened 2020-11-05
Latest IRB fails on TruffleRuby because it uses RubyVM


eregon:
```
$ ruby-build truffleruby-dev ~/.rubies/truffleruby-dev 
$ chruby truffleruby-dev 
$ gem i irb
$ irb
/home/eregon/.rubies/truffleruby-dev/lib/gems/gems/irb-1.2.7/lib/irb/ruby-lex.rb:229:in `const_missing': uninitialized constant RubyLex::RubyVM (NameError)
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/irb-1.2.7/lib/irb/ruby-lex.rb:229:in `block in check_code_block'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/irb-1.2.7/lib/irb/ruby-lex.rb:36:in `compile_with_errors_suppressed'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/irb-1.2.7/lib/irb/ruby-lex.rb:228:in `check_code_block'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/irb-1.2.7/lib/irb/ruby-lex.rb:129:in `check_state'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/irb-1.2.7/lib/irb/ruby-lex.rb:65:in `block (2 levels) in set_input'
	from <internal:core> core/array.rb:311:in `each_index'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/irb-1.2.7/lib/irb/ruby-lex.rb:63:in `block in set_input'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/reline-0.1.7/lib/reline/line_editor.rb:79:in `call'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/reline-0.1.7/lib/reline/line_editor.rb:79:in `check_multiline_prompt'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/reline-0.1.7/lib/reline/line_editor.rb:330:in `rerender'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/reline-0.1.7/lib/reline.rb:235:in `inner_readline'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/reline-0.1.7/lib/reline.rb:175:in `readmultiline'
[...]

PR #134, opened 2020-11-05
Add a fallback for check_code_block that does not depend on implementation-private APIs

eregon:
* Fixes https://github.com/ruby/irb/issues/133

I also suggest to add TruffleRuby in CI to notice this earlier: https://github.com/ruby/irb/pull/135

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -222,11 +222,16 @@ class RubyLex
     begin # check if parser error are available
       verbose, $VERBOSE = $VERBOSE, nil
       case RUBY_ENGINE
+      when 'ruby'
+        self.class.compile_with_errors_suppressed(code) do |inner_code, line_no|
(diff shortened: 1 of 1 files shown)


aycabta (approved):
Good!

lib/irb/ruby-lex.rb
```diff
@@ -222,11 +222,16 @@ def check_code_block(code)
...
+          RubyVM::InstructionSequence.compile(inner_code, nil, nil, line_no)
+        end
       when 'jruby'
         JRuby.compile_ir(code)
       else
-        self.class.compile_with_errors_suppressed(code) do |inner_code, line_no|
-          RubyVM::InstructionSequence.compile(inner_code, nil, nil, line_no)
+        catch(:valid) do
+          eval("BEGIN { throw :valid, true }\n#{code}")
+          false
```
eregon, line 234: A note mostly from myself, the code is from https://github.com/ruby/irb/issues/133#issue-737109995

PR #138, opened 2020-11-20
Change context-mode's default to new mode 4.

marcandre:
This new mode uses a copy of the TOPLEVEL_BINDING. This is compatible with refinements (contrary to mode 3), while keeping nested IRB sessions separate.

Current default makes it impossible to call `using SomeModule` from the top level.
Other modes work:

```sh
irb --single-irb # => works (shared context for nested sessions)
irb --context 0 # => works
irb --context 1 # => works
irb --context 2 # => works
irb --context 3 # (current default) => does not work
irb --context 4 # (proposed default) => works
```

I believe the default mode can not be made to work as it is by design that one can not call `using` from inside a method.

This PR proposes a 4th mode that seems the most versatile and least error prone: `TOPLEVEL_BINDING.dup`.

I am surprised that it isn't used already, I was expecting this to be the default. Either there is some limitation that I am not aware of, or else maybe it is for historical reason that is not used?

Fixes [redmine bug #9580](https://bugs.ruby-lang.org/issues/9580)

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -108,7 +108,7 @@ module IRB # :nodoc:
     @CONF[:PROMPT_MODE] = (STDIN.tty? ? :DEFAULT : :NULL)
     @CONF[:AUTO_INDENT] = true
 
-    @CONF[:CONTEXT_MODE] = 3 # use binding in function on TOPLEVEL_BINDING
+    @CONF[:CONTEXT_MODE] = 4 # use a copy of TOPLEVEL_BINDING
     @CONF[:SINGLE_IRB] = false
 
     @CONF[:LC_MESSAGES] = Locale.new
--- a/lib/irb/workspace.rb
+++ b/lib/irb/workspace.rb
@@ -51,11 +51,13 @@ EOF
           end
           @binding = BINDING_QUEUE.pop
 
-        when 3	# binding in function on TOPLEVEL_BINDING(default)
+        when 3	# binding in function on TOPLEVEL_BINDING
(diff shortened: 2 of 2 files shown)


aycabta:
It's just for historical reason. Great, thank you!

marcandre:
Super, thanks!

marcandre:
Will you take care of copying this commit to main ruby?

aycabta:
@marcandre
> Will you take care of copying this commit to main ruby?

Done!

marcandre:
Great, thanks! 🎉

eregon:
I think this change might cause https://bugs.ruby-lang.org/issues/17623#note-3

marcandre:
@eregon indeed...

I only get `TOPLEVEL_BINDING.local_variables # => [:str, :version]` though...

eregon:
I think `bin/irb` in general might be generated, e.g., when doing `gem install irb`, so if we change it it should probably be done in RubyGems (the whole logic could be in a `-> { ... }.call` to avoid local vars).
I guess in general it could be useful if Ruby exposed (a frozen) `EMPTY_BINDING` or so, which could be a copy of TOPLEVEL_BINDING but before loading the main script (e.g., libraries required through `-rmylib` see TOPLEVEL_BINDING as empty).

marcandre:
I see the comment about them being generated, but I also notice that `bin/racc` has a different layout than the others so I'm no sure they are actually generated.

marcandre:
I created https://github.com/ruby/ruby/pull/4176

PR #139, opened 2020-11-21
workaround for lack of tokens from `Ripper.lex`

### LINKED ISSUE #38: `def foo(&nil);end` does not raise syntax error
Expected: raise syntax error like ruby command

```
% ruby -e 'def foo(&nil); end'
-e:1: syntax error, unexpected `nil', expecting local variable or method
def foo(&nil); end
```

Actual: continue to input with indents

```
irb(main):001:1> def foo(&nil); end
irb(main):002:1>   end
irb(main):003:1>     end
irb(main):004:1>
```

no6v:
* Fixes #38


aycabta:
Thank you!

PR #145, opened 2020-12-08
Enable deprecation warnings when using -W or -w

jeremyevans:
Requested in Ruby Bug 17377

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -131,7 +131,7 @@ module IRB # :nodoc:
         $DEBUG = true
         $VERBOSE = true
       when "-w"
-        $VERBOSE = true
+        Warning[:deprecated] = $VERBOSE = true
       when /^-W(.+)?/
         opt = $1 || argv.shift
         case opt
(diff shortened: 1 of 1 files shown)


lib/irb/init.rb
```diff
@@ -140,7 +140,7 @@ def IRB.parse_opts(argv: ::ARGV)
         when "1"
           $VERBOSE = false
         else
-          $VERBOSE = true
+          Warning[:deprecated] = $VERBOSE = true
```
JunichiIto, line 143: Don't you need to consider category?

```
$ ruby --help
  -W[level=2|:category]
                  set warning level; 0=silence, 1=medium, 2=verbose
```

jeremyevans: I don't think considering it is required for this commit. I'm not against it, but it's significantly more involved in terms of implementation, and I'm not attempting to replicate all aspects of the behavior of `ruby -W` in this commit, only trying to address what I see as the most important issue.  I'll leave it up to the `irb` maintainers whether replicating all behaviors of `ruby -w` and `ruby -W` for `irb -w` and and `irb -W` is worth the implementation cost.

PR #150, opened 2020-12-22
Make it possible to use measure even if it is installed as gem

elim:
Include the file that implements this command in the irb gem.


aycabta:
Long time no see, thanks for picking up on the details!

elim:
I'm glad you remember me! Thank you for your immediate response as well :)

PR #159, opened 2020-12-29
Enhance colored inspect output

k0kubun:
Kudos to pry.

## before
Screenshot from 2020-12-28 22-24-08

## after
Screenshot from 2020-12-28 22-24-30


k0kubun:
@aycabta checked the code, and said it's okay to switch the base implementation from inspect to pretty_inspect. Merging.

Issue #158, opened 2020-12-28
`while cond do` indentation detection is off


schneems:
## Description

IRB attempts to detect the correct indentation level based on keywords (while, until, for, do, etc.) and some keywords can occur on the same line. For instance (`while cond do`) needs to only be indented once instead of twice. This is handled here https://github.com/ruby/ruby/blob/06b44f819eb7b5ede1ff69cecb25682b56a1d60c/lib/irb/ruby-lex.rb#L362-L365

However, there are cases that accidentally trigger this code and do not provide correct indentation. Consider this method:

```ruby
def yield_if(block)
  yield if block.call
end
```

It can be called like this:

```ruby
yield_if ->{true} do
  puts "bad indent"
end
```

But if you try to type that in it will be forced to:

```ruby
yield_if ->{true} do
puts "bad indent"
end
```

## Result of irb_info

Here is a video of the indentation behavior:

https://user-images.githubusercontent.com/59744/103242427-be43ef80-491b-11eb-8ad1-e5cf5d000b48.mp4

Please past the result of `irb_info` command in IRB.

ex.)

```
$ irb
irb(main):001:0> irb_info
=>
Ruby version: 3.0.0
IRB version: irb 1.3.0 (2020-12-25)
InputMethod: ReidlineInputMethod with Reline 0.2.0
```

## Terminal Emulator

iterm 2

## Setting Files

```
$ cat ~/.irbrc
cat: /Users/rschneeman/.irbrc: No such file or directory
$ cat ~/.inputrc
cat: /Users/rschneeman/.inputrc: No such file or directory
```

schneems:
Wow. Epic fix 😍. Thanks for working on this!

PR #160, opened 2020-12-31
Handle indentations related to keyword "do" correctly

aycabta:
This fixes ruby/irb#158.


schneems:
Wow epic fix, thanks for working on this 😍

PR #164, opened 2021-01-04
Switch to use GitHub Actions

hsbt:


.github/workflows/test.yml
```diff
@@ -0,0 +1,24 @@
+name: build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: build (${{ matrix.ruby }} / ${{ matrix.os }})
+    strategy:
+      matrix:
+        ruby: [ 3.0, 2.7, 2.6, 2.5, head, truffleruby-head ]
```
mmizutani, line 10: As pointed out by aycabta [here](https://github.com/ruby/reline/pull/238#pullrequestreview-558916125), it would be necessary to quote `3.0` as `'3.0'` so that it won't be interpreted as `3`.

PR #165, opened 2021-01-04
Stringify when a non-object is passed to PP#text

aycabta:
If a nested object is passed to `#pp`, it may be sometimes passed to the `#text` method as an object without being stringified.

This is [fixed on the Ruby main repository](https://github.com/ruby/ruby/commit/433a3be86a811de0b4adbb92e054ee3a6fc6b4d8), but it was a bug of Ripper so still needs this workaround for using irb as a gem on Ruby 3.0.0 or earlier.

### DIFF
--- a/lib/irb/color_printer.rb
+++ b/lib/irb/color_printer.rb
@@ -10,7 +10,12 @@ module IRB
       out
     end
 
-    def text(str, width = str.length)
+    def text(str, width = nil)
+      unless str.is_a?(String)
+        str = str.inspect
+      end
+      width ||= str.length
+
       case str
       when /\A#</, '=', '>'
         super(Color.colorize(str, [:GREEN]), width)
--- a/test/irb/test_color.rb
+++ b/test/irb/test_color.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: false
 require 'test/unit'
 require 'irb/color'
+require 'irb/color_printer'
 require 'rubygems'
 require 'stringio'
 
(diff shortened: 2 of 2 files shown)


lib/irb/color_printer.rb
```diff
@@ -10,12 +10,17 @@ def self.pp(obj, out = $>, width = 79)
       out
     end
 
-    def text(str, width = str.length)
+    def text(str, width = nil)
       case str
       when /\A#</, '=', '>'
-        super(Color.colorize(str, [:GREEN]), width)
+        super(Color.colorize(str, [:GREEN]), str.length)
```
k0kubun, line 16: I think `width` should not be ignored when it's given. How about this?:

```rb
    def text(str, width = nil)
      unless str.is_a?(String)
        str = str.inspect
      end
      width ||= str.length

      case str
      when /\A#</, '=', '>'
        super(Color.colorize(str, [:GREEN]), width)
      else
        super(Color.colorize_code(str, ignore_error: true), width)
      end
    end
```

test/irb/test_color.rb
```diff
@@ -152,6 +153,19 @@ def test_colorize_code_complete_false
       end
     end
 
+    def test_color_printer
+      unless ripper_lexer_scan_supported?
+        skip 'Ripper::Lexer#scan is supported in Ruby 2.7+'
+      end
+      {
+        1 => "#{BLUE}#{BOLD}1#{CLEAR}",
```
k0kubun, line 161:
```suggestion
        1 => "#{BLUE}#{BOLD}1#{CLEAR}",
        Struct.new('IRBTestColorPrinter', :a).new('test') => "#{GREEN}#<struct Struct::IRBTestColorPrinter#{CLEAR} a#{GREEN}=#{CLEAR}#{RED}#{BOLD}\"#{CLEAR}#{RED}test#{CLEAR}#{RED}#{BOLD}\"#{CLEAR}#{GREEN}>#{CLEAR}",
```
let me add this too.

aycabta:
Fixed them.

k0kubun (approved):
:+1:

PR #168, opened 2021-01-06
do not escape a predicate method for doc namespace

### LINKED ISSUE #88: RDoc for predicate methods
I noticed that `RDoc` for predicate methods (i.e. end with `?`) are not shown.
I guess the root cause is that `?` is escaped.

```ruby
IRB::InputCompletor.retrieve_completion_data("1.odd?", bind: binding, doc_namespace: true)
# => "Integer.odd\\?"
```

no6v:
* Fixes #88

running the test in this PR with master, it fails as follow due to too much escaping `?`:

```
[15/67] TestIRB::TestCompletion#test_complete_predicate? = 0.00 s                        
  1) Failure:
TestIRB::TestCompletion#test_complete_predicate? [/var/tmp/git/ruby/irb/test/irb/test_completion.rb:56]:
<"Integer.positive?"> expected but was
<"Integer.positive\\?">.
```

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -47,7 +47,7 @@ module IRB
       when /^((["'`]).*\2)\.([^.]*)$/
         # String
         receiver = $1
-        message = Regexp.quote($3)
+        message = $3
 
         candidates = String.instance_methods.collect{|m| m.to_s}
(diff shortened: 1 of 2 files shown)


aycabta:
Great, thank you!

PR #174, opened 2021-01-19
Use a real screen size for pp by default

k0kubun:
This tries to mitigate the situation around https://bugs.ruby-lang.org/issues/17557 a little bit.

It seems this API is fast enough, but is that always true?

```rb
irb(main)[02:0]> measure 'Reline.get_screen_size.last'
processing time: 0.000165s
=> nil
```


aycabta (approved):
For example, getting the cursor position is slow, but getting the screen size is not so slow. I don't think it's a problem.

PR #177, opened 2021-01-22
Bug17564 exit after source

no6v:
[Bug #17564] https://bugs.ruby-lang.org/issues/17564

Although some documents in particular `IRB::FileInputMethod` should be updated, would you please confirm this?


no6v:
Hmm, the test results are different from my local environment, fixing...

PR #179, opened 2021-01-27
Actions - add a few Windows jobs

MSP-Greg:
Also, add matrix notes, ruby/setup-ruby install Bundler by default

std-lib repos should test on Windows rather than break ruby/ruby...

macOS-11.0 is better, but still troublesome.


hsbt:
@MSP-Greg I'll close this because this is steal.

PR #180, opened 2021-01-27
Fix IrbLoader on Windows

MSP-Greg:
`File.absolute_path?` is only available with Ruby 2.7 and later, hence, a bit messy...


MSP-Greg:
Closing.  See https://github.com/ruby/ruby/pull/4127 & https://github.com/ruby/ruby/commit/f6387ae0737e

PR #202, opened 2021-03-16
process multi-line pastes as a single entity

ccutrer:
this allows pasting leading-dot chained methods correctly:

```ruby
class A
  def a; self; end
  def b; true; end
end

a = A.new

a
 .a
 .b
```

will properly return `true` instead of erroring on the `.a` line:

```
irb(main):001:1*     class A
irb(main):002:1*       def a; self; end
irb(main):003:0>     end
irb(main):004:0*
irb(main):005:0>     a = A.new
irb(main):006:0*
irb(main):007:0>     a
irb(main):008:0>      .a
irb(main):009:0>      .a
=> #<A:0x00007f984211fbe8>
```

I'd be happy to write a test, if you could give me some pointers of how to do so. The only thing I can figure to do is to shell out to `irb` itself, since this is dependent in Reline reading from stdin. Possibly even requiring a pty.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -7,4 +7,11 @@ Rake::TestTask.new(:test) do |t|
   t.test_files = FileList["test/**/test_*.rb"]
 end
 
+Rake::TestTask.new(:test_yamatanooroti) do |t|
+  t.libs << 'test'
+  t.libs << 'lib'
+  #t.loader = :direct
+  t.pattern = 'test/irb/yamatanooroti/test_*.rb'
+end
+
 task :default => :test
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -315,7 +315,13 @@ module IRB
       Reline.output = @stdout
       Reline.prompt_proc = @prompt_proc
       Reline.auto_indent_proc = @auto_indent_proc if @auto_indent_proc
-      if l = readmultiline(@prompt, false, &@check_termination_proc)
+
+      l = readmultiline(@prompt, false) do |line|
+        next false if Reline::IOGate.in_pasting?
+        @check_termination_proc.call(line)
+      end
+
+      if l
         HISTORY.push(l) if !l.empty?
         @line[@line_no += 1] = l + "\n"
       else
--- a/test/irb/yamatanooroti/test_rendering.rb
+++ b/test/irb/yamatanooroti/test_rendering.rb
@@ -1,0 +1,54 @@
+require 'irb'
+
+begin
+  require 'yamatanooroti'
+
+  class IRB::TestRendering < Yamatanooroti::TestCase
+    def setup
+      @pwd = Dir.pwd
+      suffix = '%010d' % Random.rand(0..65535)
+      @tmpdir = File.join(File.expand_path(Dir.tmpdir), "test_irb_#{$$}_#{suffix}")
+      begin
+        Dir.mkdir(@tmpdir)
+      rescue Errno::EEXIST
+        FileUtils.rm_rf(@tmpdir)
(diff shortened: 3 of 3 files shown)


marcandre:
Looks like a good addition 👍

aycabta:
I think it's a really great patch too, and a lot of people will be happy with it.

On the other hand, Reline uses "whether the buffer has received a series of keystrokes" as a criterion to determine whether it is pasting or not. So, if there is a delay in I/O due to a slow machine or over the network, it will assume that everything is being pasted. Therefore, with this implementation alone, there is a possibility that the code that should be evaluated immediately (e.g. `a = 1`) will be followed by a new line and will not be executed forever.

We are currently writing an additional patch, so please wait a little longer.

aycabta:
I added a test example to this branch, but it uses yamatanooroti gem, which has crash problems in Ruby 3.0.0. I added a test example to this branch, but it uses yamatanooroti gem, which has crash problems in Ruby 3.0.0. We are investigating the cause, but for now, please use Ruby 2.7.2 and install `gem install yamatanooroti vterm`, then try `rake test_yamatanooroti`. You should see the test run. You can use this to add a test for your patch.

aycabta:
I've written a patch that evaluates top-level each statement at when multiple statements are pasted at once, https://github.com/ruby/irb/commit/d991d110e5c4de59a71935a4615578e4f9043e45.

If I paste below;

```ruby
class A
  def b; self; end
  def c; true; end
end

a = A.new

a
  .b
  # aaa
  .c

(a)
  &.b()


class A def b; self; end; def c; true; end; end;
a = A.new
a
  .b
  # aaa
  .c
(a)
  &.b()
```

must be below;

```


irb(main):001:1* class A
irb(main):002:1*   def b; self; end
irb(main):003:1*   def c; true; end
irb(main):004:0> end
=> :c
irb(main):005:0>
irb(main):006:0> a = A.new
=> #<A:0x0000562a844980e0>
irb(main):007:0>
irb(main):008:0> a
irb(main):009:0>   .b
irb(main):010:0>   # aaa
irb(main):011:0>   .c
=> true
irb(main):012:0>
irb(main):013:0> (a)
irb(main):014:0>   &.b()
=> #<A:0x0000562a844980e0>
irb(main):015:0>
irb(main):016:0>
irb(main):017:0> class A def b; self; end; def c; true; end; end;
=> :c
irb(main):018:0> a = A.new
=> #<A:0x0000562a83f73d80>
irb(main):019:0> a
irb(main):020:0>   .b
irb(main):021:0>   # aaa
irb(main):022:0>   .c
=> true
irb(main):023:0> (a)
irb(main):024:0>   &.b()
=> #<A:0x0000562a83f73d80>
irb(main):025:0>
```

I'll merge this Pull Request if you add a test. And I'll make a another Pull Request with the additional patch later.

aycabta:
@ccutrer Could you write a test with the https://github.com/ruby/irb/pull/202/commits/279155fceeca4cb055aadde934d5c6b27a9bb83d that I added to this branch? Then I'll merge.

I've written about how to use it before:

> Ruby 2.7.2 and install `gem install yamatanooroti vterm`, then try `rake test_yamatanooroti`. You should see the test run.

aycabta:
I and @mame fixed the gems and released new versions, so you can use Ruby 3.0.0 to write a test now.

aycabta:
I'm sorry, I added a test for multiline paste due to release schedule. I'll merge now. This feature is so great for many Ruby users. Thank you so much!

ccutrer:
Heh, no apologies necessary, for getting done what I didn't get around to. Thank you so much! I'm very much looking forward to this feature.

PR #203, opened 2021-03-18
Implement pry-like ls command

k0kubun:
I implemented pry's `ls` command for irb. This is useful when you want to inspect constants, methods, and variables of self or an object without typing a lot.

```rb
$ irb
irb(main)[01:0]> ls ERB.new('test')
ERB#methods:
  def_class         def_method        def_module        encoding          filename          filename=         lineno
  lineno=           location=         make_compiler     result            result_with_hash  run               set_eoutvar
  src
instance variables: @_init  @encoding  @filename  @frozen_string  @lineno  @src
```

### DIFF
--- a/lib/irb/cmd/ls.rb
+++ b/lib/irb/cmd/ls.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require_relative "nop"
+require_relative "../color"
+
+# :stopdoc:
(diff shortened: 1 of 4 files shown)


aycabta (changes requested):
Please don't contain merge commit and please use `git rebase origin/master`.

PR #204, opened 2021-03-18
Add whereami command

k0kubun:
This is another command in pry. When you execute a lot of expressions after your `binding.irb` invocation, sometimes the source code of your context may go out of your screen. If you use this `whereami` like gdb's `list`, you can show the code again.

```rb
$ ruby /tmp/a.rb

From: /tmp/a.rb @ line 3 :

    1: a = 1
    2: @b = 2
 => 3: binding.irb

irb(main)[01:0]> a
=> 1
irb(main)[02:0]> @b
=> 2
irb(main)[03:0]> whereami

From: /tmp/a.rb @ line 3 :

    1: a = 1
    2: @b = 2
 => 3: binding.irb

=> nil
```

### DIFF
--- a/lib/irb/cmd/whereami.rb
+++ b/lib/irb/cmd/whereami.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require_relative "nop"
+
+# :stopdoc:
+module IRB
+  module ExtendCommand
+    class Whereami < Nop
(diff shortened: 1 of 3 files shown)


aycabta (dismissed):
Please don't contain merge commit and please use `git rebase origin/master`.

PR #206, opened 2021-03-18
Avoid loading files' local variables [Bug #17623]

marcandre:
https://bugs.ruby-lang.org/issues/17623

### DIFF
--- a/exe/irb
+++ b/exe/irb
@@ -8,4 +8,5 @@
 
 require "irb"
 
+IRB::TOPLEVEL_BINDING = binding
 IRB.start(__FILE__)
--- a/bin/console
+++ b/bin/console
@@ -3,4 +3,5 @@
 require "bundler/setup"
 require_relative "../lib/irb"
 
+IRB::TOPLEVEL_BINDING = binding
 IRB.start(__FILE__)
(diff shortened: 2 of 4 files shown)


marcandre:
This also fixes the same issue with `irb --single` that must have existed for years

jeremyevans (approved):
I think this is a good way to fix the issue.

eregon:
This seems fine to me too given it assigns to `IRB::TOPLEVEL_BINDING` and does not override `::TOPLEVEL_BINDING`.

exe/irb
```diff
@@ -8,4 +8,5 @@
 
 require "irb"
 
+IRB::TOPLEVEL_BINDING = binding
```
aycabta, line 11: I believe that this is a good solution to fix the problem but this assignment is so far from lib/irb/workspace.rb. Would you move the assignment to the top-level of lib/irb/workspace.rb? And, `TOPLEVEL_BINDING` is confusing with the existing constant, so I think a different name would be better.

aycabta (changes requested):

exe/irb
eregon: Agreed, a different constant name would be much less confusing.

PR #223, opened 2021-04-07
Moved development dependencies to Gemfile

nobu:
As expressions in a gemspec file are evaluated at the build time, but not the run time, the conditional in the gemspec will not work as intended.


olleolleolle (approved):
👍

PR #224, opened 2021-04-07
Ripper::Lexer::Elem#state is defined since Ruby 2.5

nobu:
And the required ruby version is 2.5 or later.


olleolleolle (approved):
I checked the gemspec, and it has a requirement of 2.5+.

👍 Super!

Issue #229, opened 2021-04-21
Unexpected multibyte error message is displayed


koic:
## Description

The following is an unexpected RDBMS error message.

```console
% bin/rails c
Running via Spring preloader in process 72443
Loading development environment (Rails 6.1.3.1)
irb(main):001:0> ActiveRecord::Base.connection.truncate(:table_name)
   (2.0ms)  TRUNCATE TABLE "TABLE_NAME"
stmt.c:267:in oci8lib_300.bundle: OCIError: ORA-00942: \xE8\xA1\xA8\xE3\x81\xBE\xE3\x81\x9F\xE3\x81\xAF\xE3\x83\x93\xE3\x83\xA5\xE3\x83\xBC\xE3\x81\x8C\xE5\xAD\x98\xE5\x9C\xA8\xE3\x81\x97\xE3\x81\xBE\xE3\x81\x9B\xE3\x82\x93\xE3\x80\x82 (ActiveRecord::StatementInvalid)
stmt.c:267:in oci8lib_300.bundle: ORA-00942: \xE8\xA1\xA8\xE3\x81\xBE\xE3\x81\x9F\xE3\x81\xAF\xE3\x83\x93\xE3\x83\xA5\xE3\x83\xBC\xE3\x81\x8C\xE5\xAD\x98\xE5\x9C\xA8\xE3\x81\x97\xE3\x81\xBE\xE3\x81\x9B\xE3\x82\x93\xE3\x80\x82 (OCIError)
```

And the following is encoded to the expected message.

```console
irb(main):002:0> "ORA-00942: \xE8\xA1\xA8\xE3\x81\xBE\xE3\x81\x9F\xE3\x81\xAF\xE3\x83\x93\xE3\x83\xA5\xE3\x83\xBC\xE3\x81\x8C\xE5\xAD\x98\xE5\x9C\xA8\xE3\x81\x97\xE3\x81\xBE\xE3\x81\x9B\xE3\x82\x93\xE3\x80\x82".encode('UTF-8')
=> "ORA-00942: 表またはビューが存在しません。"
```

Pry is displaying the expected message, so it seems that REPL has a cause.

## Result of irb_info

Please past the result of `irb_info` command in IRB.

```console
irb(main):003:0> irb_info
=>
Ruby version: 3.0.1
IRB version: irb 1.3.5 (2021-04-03)
InputMethod: ReidlineInputMethod with Reline 0.2.5 and /Users/koic/.inputrc
[...]

PR #237, opened 2021-05-10
Treat encodings in exception correctly

aycabta:
This fixes https://github.com/ruby/irb/issues/229.


lib/irb.rb
```diff
@@ -640,7 +646,12 @@ def handle_exception(exc)
           message = exc.full_message(order: :top)
           order = :top
         end
-        message = convert_invalid_byte_sequence(message)
+        message = convert_invalid_byte_sequence(message, exc.message.encoding)
+        message = encode_with_invalid_byte_sequence(message, IRB.conf[:LC_MESSAGES].encoding) if message.encoding != IRB.conf[:LC_MESSAGES].encoding
+        #message = message.encode(IRB.conf[:LC_MESSAGES].encoding).scrub { |c|
+        #  puts "!!!!!!!!!!! #{c.inspect} is a scrub in #{IRB.conf[:LC_MESSAGES].encoding}"
+        #  c.bytes.map{ |b| "\\x#{b.to_s(16).upcase}" }.join
+        #}
```
olleolleolle, line 654: Perhaps these lines were sketching not meant for production use?

aycabta: Good catch, thank you so much!

PR #242, opened 2021-05-20
Pass local variables from workspace binding to lexer

jeremyevans:
This fixes at least an issue where irb will incorrectly assume
code opens a heredoc when it does not, such as this code:

```ruby
s1 = 'testing'
s2 = 'this'
s2 <<s1
p s1
s1
```

Ruby parses the `s2 <<s1` as `s2.<<(s1)`, not as a heredoc, because
`s2` is a local variable in scope.  irb was using ripper without
letting ripper know that `s2` was a local variable, so ripper would
lex it as a heredoc instead of a method call.

Fix the situation by prepending a line at line 0 with all local
variable definitions in scope whenever lexing.  This fixes the
heredoc issue, and potentially other issues that depend on whether
an identifier is a local variable or not.

Fixes [Bug #17530]

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -30,26 +30,25 @@ class RubyLex
     @prompt = nil
   end
 
-  def self.compile_with_errors_suppressed(code)
-    line_no = 1
+  def self.compile_with_errors_suppressed(code, line_no: 1)
     begin
       result = yield code, line_no
     rescue ArgumentError
       code = ";\n#{code}"
-      line_no = 0
+      line_no -= 1
       result = yield code, line_no
     end
     result
   end
 
   # io functions
-  def set_input(io, p = nil, &block)
+  def set_input(io, p = nil, context: nil, &block)
     @io = io
     if @io.respond_to?(:check_termination)
       @io.check_termination do |code|
         if Reline::IOGate.in_pasting?
           lex = RubyLex.new
-          rest = lex.check_termination_in_prev_line(code)
+          rest = lex.check_termination_in_prev_line(code, context: context)
           if rest
             Reline.delete_text
             rest.bytes.reverse_each do |c|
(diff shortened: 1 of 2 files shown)


lib/irb/ruby-lex.rb
```diff
@@ -30,26 +30,25 @@ def initialize
...
 
-  def self.compile_with_errors_suppressed(code)
-    line_no = 1
+  def self.compile_with_errors_suppressed(code, line_no: 1)
     begin
       result = yield code, line_no
     rescue ArgumentError
       code = ";\n#{code}"
-      line_no = 0
+      line_no -= 1
```
olleolleolle, line 38: Minor: Perhaps a code comment here, to explain the meaning of the negative number?

lib/irb/ruby-lex.rb
```diff
@@ -129,15 +128,25 @@ def set_prompt(p = nil, &block)
     :on_param_error
   ]
 
-  def self.ripper_lex_without_warning(code)
+  def self.ripper_lex_without_warning(code, context: nil)
     verbose, $VERBOSE = $VERBOSE, nil
+    if context
+      lvars = context&.workspace&.binding&.local_variables
+      if lvars && !lvars.empty?
+        code = "#{lvars.join('=')}=nil\n#{code}"
```
olleolleolle, line 136: 👏  This is wonderful. Thanks!

olleolleolle (approved):
Great description of the change, and a clear implementation.

lib/irb/ruby-lex.rb
jeremyevans: I didn't understand why this rescue existed, but with some help from git blame, I found it is due to incomplete magic comment handling.  I pushed another commit explaining not just the line number issue, but the point of the rescue block in general.

aycabta (approved):

PR #245, opened 2021-06-20
Sort ls result ordered by anscestry

neko314:
This PR suggests to improve `ls` command.  

I changed these two points.
* I expect it to be the same as the result of the method search, but it is not.
* I want the output to be in the same order as the results of `ancestors`, but but in fact it is displayed in the order of modules->classes.

## Example 
Definitions are below; a little bit different from sample codes in #234, I added some modules and classes have the same methods.

```ruby
class P
  def m() end
  def m2() end
end

class C < P
  def m1() end
  def m2() end
end

module M
  def m1() end
  def m3() end
end

module M2
  include M
  def m4() end
end

obj = C.new; obj.extend M2; def obj.m5() end;
```

## Before

```
> ls obj
M#methods: m1  m3
M2#methods: m4
C.methods: m5
P#methods: m
C#methods: m1  m2
```
* `m1` appears twice.
* `obj.singleton_class.ancestors` is `[M2, M, C, P, Object, PP::ObjectMixin, Kernel, BasicObject]`. I was a little confused the difference between my expectation and actual.

## After

```
> ls obj
P#methods: m
C#methods: m2
M#methods: m1  m3
M2#methods: m4
C.methods: m5
```

### DIFF
--- a/lib/irb/cmd/ls.rb
+++ b/lib/irb/cmd/ls.rb
@@ -17,25 +17,17 @@ module IRB
         klass  = (obj.class == Class || obj.class == Module ? obj : obj.class)
 
         o.dump("constants", obj.constants) if obj.respond_to?(:constants)
-        dump_singleton_methods(o, klass, obj)
-        dump_instance_methods(o, klass)
+        dump_methods(o, klass, obj)
         o.dump("instance variables", obj.instance_variables)
         o.dump("class variables", klass.class_variables)
         o.dump("locals", locals)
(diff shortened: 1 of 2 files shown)


neko314:
CI failed https://github.com/ruby/irb/pull/245/checks?check_run_id=2867533381
This problem loooks to be resolved by https://github.com/ruby/irb/pull/247

aycabta:
😎

PR #247, opened 2021-06-20
Disable locale warning from TruffleRuby

nobu:
`assert_in_out_err` fails due to the warnings since oracle/truffleruby@6b03a11d685c4ce58855628ebfef5bafbfa211af.


aycabta (approved):

eregon:
Thanks for the fix!
I think ideally assert_in_out_err/assert_separately should not change the locale, but that seems much harder to change now.

PR #267, opened 2021-08-09
Pass RI options to help

f6p:
Hello,

I noticed `help` command ignores `RI` env variable. Sometimes it has little significance but sometimes not - especially if `RI` env includes list of directories to search. This happens because calling `RDoc::RI::Driver.new` won't (contrary to what source comments say) parse `RI` env variable. To parse `RI` env var `process_args` has to be called (it returns options that could later be passed to driver).

Cheers
🍷


aycabta:
This is an important correction. Thanks!

PR #268, opened 2021-08-28
Capitalize IRBRC environment variable name in docs

purp:
The docs currently say that setting `$irbrc` to a file will cause it to be loaded. This is not true; [line 360 of `lib/irb/init.rb`](https://github.com/ruby/irb/blob/master/lib/irb/init.rb#L360) checks for `ENV["IRBRC"]`. Changed to match the code.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -100,7 +100,7 @@ require_relative "irb/easter-egg"
 # * +.irbrc+
 # * +irb.rc+
 # * +_irbrc+
-# * <code>$irbrc</code>
+# * <code>$IRBRC</code>
 #
 # The following are alternatives to the command line options. To use them type
 # as follows in an +irb+ session:


adam12:
I don't believe that is accurate. I think it _looks_ for a file named `$irbrc`. See this line: https://github.com/ruby/irb/blob/98610236ecf9499c55343e778fa0da23a0dfa1c4/lib/irb/init.rb#L379

nobu:
> I don't believe that is accurate. I think it _looks_ for a file named `$irbrc`. See this line:
> 
> https://github.com/ruby/irb/blob/98610236ecf9499c55343e778fa0da23a0dfa1c4/lib/irb/init.rb#L379

That is literally a `$irbrc` file, for very old platforms where could’t create dot files.

aycabta:
Thank you for your reporting and a patch.

> That is literally a `$irbrc` file, for very old platforms where could’t create dot files.

Yes, what you see on the doc is a description of a kind of backward compatibility. Therefore, this description is correct.

PR #298, opened 2021-10-06
Make IRB::Context#exit call super if UncaughtThrowError

jeremyevans:
Fixes calling exit after binding.irb.

Fixes [Bug #18234]


aycabta:
Thank you!

Issue #299, opened 2021-10-06
unmatched close parenthesis: (RegexpError)     


eregon:
## Description

When pasting `defined?(::Ractor)`
(and copying e.g. from https://github.com/ruby/ostruct/pull/29/files)
on ruby-head's irb I get:
```
$ irb
irb(main):001:0> defined?(::Ractor)/home/eregon/prefix/ruby-master/lib/ruby/3.1.0/irb/completion.rb:207:in `retrieve_completion_data': unmatched close parenthesis: /^Ractor)/ (RegexpError)
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/irb/completion.rb:130:in `block in <module:InputCompletor>'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:1653:in `call_completion_proc_with_checking_args'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:524:in `call_completion_proc_with_checking_args'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline.rb:221:in `block in <class:Core>'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:569:in `instance_exec'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:569:in `call'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:604:in `call'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:643:in `render_each_dialog'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:628:in `block in render_dialog'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:627:in `each'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:627:in `render_dialog'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline/line_editor.rb:491:in `rerender'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline.rb:325:in `block (3 levels) in inner_readline'
	from /home/eregon/prefix/ruby-master/lib/ruby/3.1.0/reline.rb:323:in `each'
[...]

PR #302, opened 2021-10-10
Ignore parenthesis during completion

kaiquekandykoga:
Hello guys,

This is an attempt to fix the issue https://github.com/ruby/irb/issues/299.

When absolute constant is used with ) an exception is raised with regex during completion.

It includes ) character verification to prevent raising an exception.

After

```
3.0.2 :001 > defined?(::A)
 => nil
```

Before

```
3.0.2 :001 > defined?(::A)
/irb/lib/irb/completion.rb:207:in `retrieve_completion_data': unmatched close parenthesis: /^A)/ (RegexpError)
```

Thank you very much.

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -197,7 +197,7 @@ module IRB
         end
         candidates.grep(/^#{Regexp.quote(sym)}/)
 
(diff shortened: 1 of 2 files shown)


aycabta:
Good catch, thank you!

PR #311, opened 2021-11-29
Add Alt+d keycode when convert-meta isn't used

ima1zumi:
fix https://github.com/ruby/reline/issues/357

I pushed [reline#389](https://github.com/ruby/reline/pull/389) for when convert-meta is not turned on in .inputrc.
Alt+D in irb also needs to be set to the keycode for not using convert-meta.


aycabta:
Good!

PR #312, opened 2021-12-05
Changed to call Kernel.print

osyo-manga:
If you call `binding.irb` on a class defined `#print`, it will crash, so call `Kernel.print`.

Fix [[Bug #18389]](https://bugs.ruby-lang.org/issues/18389) `binding.irb` can fail in some classes that implement `context` and `print` methods.

### DIFF
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -235,7 +235,7 @@ module IRB # :nodoc:
           alias_method to, from
         }
       else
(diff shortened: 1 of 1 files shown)


jeremyevans:
I came up with exactly the same patch in #325.  I can't think of a reason not to make this change, so I'm going to merge this.

osyo-manga:
Thanks :)

PR #313, opened 2021-12-06
Fix warning in `Lexer::Elem#[]`

osyo-manga:
Changed to use `#pos` `#event` `#tok` `#state` since using Lexer::Elem#[0~4] now gives a warning.
see: https://github.com/ruby/ruby/commit/8944009be7418614ce7d4077807ac2b60d4d5d85


yahonda:
Just FYI, There is a pull request not to show warnings https://github.com/ruby/ruby/pull/5219

aycabta:
I think this is still great for readability. Thank you!

PR #316, opened 2021-12-15
Update remark

kaiquekandykoga:
Hello guys,

If the help is printed an information about --extra-doc-dir is shown.

Add the same details in the comments.

Thank you very much.


aycabta:
Thank you!

PR #317, opened 2021-12-16
Fix completion dialog when target include parentheses

### LINKED ISSUE #314: The completion dialog following the parentheses is wrong
## Description

The completion dialog following the parentheses is wrong.

example:
```ruby
().
"{}".
"()".
```

### Actual
image

### Expect

String methods should appear in completion dialog.

## Result of irb_info

```
irb(main):001:0> irb_info
=>
Ruby version: 3.1.0
.irbrc path: /Users/mi/.irbrc
RUBY_PLATFORM: x86_64-darwin19
```

## Terminal Emulator

Alacritty

## Setting Files

none

ima1zumi:
fix #314

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -127,7 +127,7 @@ module IRB
       if preposing && postposing
         result = CompletionRequireProc.(target, preposing, postposing)
         unless result
-          result = retrieve_completion_data(target).compact.map{ |i| i.encode(Encoding.default_external) }
+          result = retrieve_completion_data(preposing + target).compact.map{ |i| i.encode(Encoding.default_external) }
         end
         result
       else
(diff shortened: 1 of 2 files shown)


aycabta:
incorrect

Hmm...this solution has some problems:

1. Breaks the completion algorithm easily because it concatenates all strings before the target
2. The dialog also shifts due to the shift in the completion start position

I think #314 is probably due to an implementation bug or misconfiguration of `completer_quote_character` or `completer_word_break_characters`.

aycabta:
Since I'm focusing on compatibility with GNU Readline, it would be wrong if the behavior is different, so if you modify it, please be careful about that.

aycabta:
If it's not a bug of `completer_quote_character` or `completer_word_break_characters`, #400 that fixes a misconfiguration may be more appropriate, but I haven't checked yet.

aycabta:
Organize the story;

1. `completer_quote_characters` handles only `'` and `"` correctly, so only `"()"` is relevant for this setting
2. `()` is probably impossible to handle correctly in principle because of the GNU Readline specification

PR #325, opened 2022-01-07
Call Kernel.print to print warning for alias

jeremyevans:
Fixing binding.irb use in class that overrides #print.

Fixes [Bug #18389]


kaiquekandykoga:
Hello @jeremyevans ,

It is duplicated https://github.com/ruby/irb/pull/312.

jeremyevans:
@kaiquekandykoga Thanks for that information, I'll close this.

PR #332, opened 2022-01-14
Add 3.1 to CI matrix

st0012:
Since 3.1 has been released, it probably needs to be added to the matrix as well?


aycabta:
Thank you!

PR #335, opened 2022-01-16
Use `require_relative` to require lib files

st0012:
1. `require` can mislead Ruby to load system irb's files and cause constant redefined warnings as other code loads the same module/class from lib folder. Like:

```
exe/irb:11:in `<main>'
/Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/irb-1.3.7/lib/irb/ruby-lex.rb:129: warning: already initialized constant RubyLex::ERROR_TOKENS
/Users/st0012/projects/irb/lib/irb/ruby-lex.rb:130: warning: previous definition of ERROR_TOKENS was here
```

2. Most files already use `require_relative`.
3. I believe `exe/irb` should load the irb from `lib` folder instead of the system irb?

### DIFF
--- a/exe/irb
+++ b/exe/irb
@@ -6,6 +6,6 @@
 #   	by Keiju ISHITSUKA(keiju@ruby-lang.org)
 #
 
-require "irb"
+require_relative '../lib/irb'
 
 IRB.start(__FILE__)
--- a/lib/irb/color.rb
+++ b/lib/irb/color.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require 'reline'
 require 'ripper'
-require 'irb/ruby-lex'
+require_relative 'ruby-lex'
 
 module IRB # :nodoc:
   module Color
--- a/lib/irb/color_printer.rb
+++ b/lib/irb/color_printer.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 require 'pp'
-require 'irb/color'
+require_relative 'color'
 
 module IRB
   class ColorPrinter < ::PP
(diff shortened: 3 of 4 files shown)


hsbt (approved):

aycabta:
I understand the issues that need to be resolved. Thank you!

MSP-Greg:
I'm not sure if Ruby master CI checks CLI commands.  Regardless, on all three ruby-loco Windows master builds, `irb --version` fails with an error similar to:
```
C:/ruby-ucrt/lib/ruby/gems/3.2.0/gems/irb-1.4.1/exe/irb:9:in `require_relative': cannot load such file -- C:/ruby-ucrt/lib/ruby/gems/3.2.0/gems/irb-1.4.1/lib/irb (LoadError)
        from C:/ruby-ucrt/lib/ruby/gems/3.2.0/gems/irb-1.4.1/exe/irb:9:in `<top (required)>'
        from C:/ruby-ucrt/bin/irb:25:in `load'
        from C:/ruby-ucrt/bin/irb:25:in `<main>'
```

st0012:
I'm checking this now. It's working on my machine as I use `exe/irb` for development and it's also an approach used by [the debugger gem](https://github.com/ruby/debug/blob/master/exe/rdbg). I'll try local gem install to see how that works.

MSP-Greg:
Sorry, I guess I wasn't clear.  In Ruby master, the irb.rb file exists in `lib/ruby/3.2.0`, but the stub is looking for it in `lib/ruby/gems/3.2.0/gems/irb-1.4.1/lib`?

st0012:
> Or, the current code would work if irb was a bundled gem, but it's a default gem?

Yes I think you're correct. Only `exe/irb` will be placed under the gem folder (`gems/irb-1.4.1`) but not `lib`. So this change will break it. I'll create another PR to revert that part.

Issue #327, opened 2022-01-07
Allow disabling dialog colors 


teoljungberg:
## Description

I'm unsure if this falls under Reline, or IRB, if I have filed this in the wrong repository do let me know and I'll move it.

IRB allows us to disable colors using `IRB.conf[:USE_COLORIZE] = false`, however - after the dialog introduction in Ruby 3.1 (or Reline 0.3.0) - the dialog is still colored even though `--nocolorize` or the config above is present.

<img width="644" alt="image" src="https://user-images.githubusercontent.com/810650/148549593-6f02cc6d-c562-4dd9-ade3-4f82204e74d2.png">

I would expect that the colors are disabled if the irb configuration is present, or by any other means if possible.

## Result of irb_info

```
irb(main):002:0> irb_info
=>
Ruby version: 3.1.0
IRB version: irb 1.4.1 (2021-12-25)
InputMethod: ReidlineInputMethod with Reline 0.3.0
.irbrc path: /Users/teo/.irbrc
RUBY_PLATFORM: x86_64-darwin21
LANG env: en_US.UTF-8
LC_ALL env: en_US.UTF-8
East Asian Ambiguous Width: 1
```

## Terminal Emulator

kitty

## Setting Files

`~/.irbrc`:

```
require "irb"
require "irb/completion"

ENV["TERM"] = "xterm"

IRB.conf[:SAVE_HISTORY] = 1000
IRB.conf[:HISTORY_FILE] =
  if defined?(Rails) && Rails.root
    Rails.root.join("tmp", "history.rb")
  else
    File.expand_path("~/.history.rb")
  end
IRB.conf[:USE_COLORIZE] = false

if defined?(Rails)
  ActiveSupport::LogSubscriber.colorize_logging = false
end

class Object
  def interesting_methods
    case self.class
    when Class
      self.public_methods.sort - Object.public_methods
    when Module
      self.public_methods.sort - Module.public_methods
    else
      self.public_methods.sort - Object.new.public_methods
    end
  end
end

def capture_exception(&block)
  block.call
rescue => exception
  exception
end

def time(&block)
  t0 = Time.now
  block.call
  puts Time.now - t0
end

def ri(*args)
  help(*args)
end
```

st0012:
Can you tell me your expectation on "no colour"? I'm asking because:

1. It could mean not adding any ANSI escape sequence to the output.
2. Or it means visually having no additional colours.

[...]

PR #336, opened 2022-01-17
Use require_relative to load extensions/commands

st0012:
Sorry that didn't notice these when doing https://github.com/ruby/irb/pull/335.

### DIFF
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -47,7 +47,7 @@ module IRB # :nodoc:
 
     @EXTEND_COMMANDS = [
       [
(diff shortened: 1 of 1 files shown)


st0012:
BTW, after this is merged, I'll start working on making colorize control more consistent to avoid issues like https://github.com/ruby/irb/issues/327

aycabta:
Color control including dialogs needs to be handled in a proper design for the entire IRB, so please wait a bit. I'm already working on it.

st0012:
@aycabta Thanks for the information. I assume that also includes coloring on commands like `show_source`?

PR #337, opened 2022-01-17
Changed to add parameters in irbrc to pass to Reline.

pocari:
I wanted to change the color of the completion dialog in Reline, so I made the following modifications to Reline
https://github.com/ruby/reline/pull/413

Initially, I was thinking of setting the color in inputrc.

However, after receiving a [report](https://github.com/ruby/reline/pull/413#issuecomment-1008236990) from Readline that warned that it was an unknown setting, I received a suggestion to set it in irbrc.

So, I made a modification to pass the configuration items to Reline using irbrc.

This is an example of using irbrc.
スクリーンショット 2022-01-17 23 00 08

This is an example without irbrc settings.
スクリーンショット 2022-01-17 23 00 41

You can try with the following settings
```
% cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem 'reline',  github: 'pocari/reline', branch: 'support-for-changing-the-background-color-of-dialogs'
gem 'irb',  github: 'pocari/irb', branch: 'feature/add-config-params-for-reline'

% bundle install
% cat ~/.irbrc

IRB.conf[:DIALOG_DEFAULT_BG_COLOR] = 40
IRB.conf[:DIALOG_POINTER_BG_COLOR] = 47

% bundle exec irb
```

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -299,11 +299,18 @@ module IRB
       if IRB.conf[:USE_AUTOCOMPLETE]
         Reline.add_dialog_proc(:show_doc, SHOW_DOC_DIALOG, Reline::DEFAULT_DIALOG_CONTEXT)
       end
-      Reline.dialog_default_bg_color = IRB.conf[:DIALOG_DEFAULT_BG_COLOR]
-      Reline.dialog_pointer_bg_color = IRB.conf[:DIALOG_POINTER_BG_COLOR]
-
-      Reline.dialog_default_fg_color = IRB.conf[:DIALOG_DEFAULT_FG_COLOR]
-      Reline.dialog_pointer_fg_color = IRB.conf[:DIALOG_POINTER_FG_COLOR]
+      if IRB.conf[:DIALOG_DEFAULT_BG_COLOR] && Reline.respond_to?('dialog_default_bg_color=')
+        Reline.dialog_default_bg_color = IRB.conf[:DIALOG_DEFAULT_BG_COLOR]
+      end
+      if IRB.conf[:DIALOG_POINTER_BG_COLOR] && Reline.respond_to?('dialog_pointer_bg_color=')
+        Reline.dialog_pointer_bg_color = IRB.conf[:DIALOG_POINTER_BG_COLOR]
+      end
+      if IRB.conf[:DIALOG_DEFAULT_FG_COLOR] && Reline.respond_to?('dialog_default_fg_color=')
+        Reline.dialog_default_fg_color = IRB.conf[:DIALOG_DEFAULT_FG_COLOR]
+      end
+      if IRB.conf[:DIALOG_POINTER_FG_COLOR] && Reline.respond_to?('dialog_pointer_fg_color=')
+        Reline.dialog_pointer_fg_color = IRB.conf[:DIALOG_POINTER_FG_COLOR]
+      end
     end
 
     def check_termination(&block)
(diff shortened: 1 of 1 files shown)


joeyparis:
@pocari I've been using this for a little while now and haven't run into any problems.

james-caresnap:
Would it be desirable to support at least 4 color settings?
- default foreground
- default background
- highlighted foreground (pointer foreground?)
- highlighted background (pointer background?)

For example, this is a screenshot on my machine with your config. The highlighted row of the autocomplete is completely illegible.

image

pocari:
@james-caresnap 

Like you said, I needed to be able to set up a foreground view.
I've implemented that. You can check it out by doing the following

```
% cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem 'reline',  github: 'pocari/reline', branch: 'support-for-changing-the-background-color-of-dialogs'
gem 'irb',  github: 'pocari/irb', branch: 'feature/add-config-params-for-reline'

% bundle install
% cat ~/.irbrc

IRB.conf[:DIALOG_DEFAULT_BG_COLOR] = 40
IRB.conf[:DIALOG_POINTER_BG_COLOR] = 47
IRB.conf[:DIALOG_DEFAULT_FG_COLOR] = 31
IRB.conf[:DIALOG_POINTER_FG_COLOR] = 30

% bundle exec irb
```

You should be able to see something like the following

スクリーンショット 2022-02-13 15 39 38

sshock:
Would love to see this feature!  Is there any reason it hasn't been approved and merged yet?

pocari:
@BrianHawley  Thank you. I have fixed it. ( [12e5a7a](https://github.com/ruby/irb/pull/337/commits/12e5a7a52bbee013078ee39c34f80f1d96bd6034) )

pocari:
By making ~/.irbrc look like the following, this PR may be unnecessary for us.
(It was possible to include code in irbrc that directly executes the Reline api)
I think this is reasonable since what we want to set up is a feature that exists only in Reline.

```
% cat ~/.irbrc
if Reline
  Reline.dialog_default_bg_color = 40
  Reline.dialog_pointer_bg_color = 47
  Reline.dialog_pointer_fg_color = 30
  Reline.dialog_pointer_bg_color = 45
end
```

https://github.com/ruby/reline/pull/413#issuecomment-1140620678

lib/irb/input-method.rb
```diff
@@ -299,6 +299,18 @@ def initialize
       if IRB.conf[:USE_AUTOCOMPLETE]
         Reline.add_dialog_proc(:show_doc, SHOW_DOC_DIALOG, Reline::DEFAULT_DIALOG_CONTEXT)
       end
+      if IRB.conf[:DIALOG_DEFAULT_BG_COLOR] && Reline.respond_to?('dialog_default_bg_color=')
```
peterzhu2118, line 302: I think it's more idiomatic to use symbols rather than strings for `respond_to?`. E.g. `Reline.respond_to?(:dialog_default_bg_color=)`.

st0012:
@pocari I've updated the `reline` APIs in https://github.com/ruby/reline/pull/454. It doesn't require significant changes on this PR though. But I hope we can:

1. Simplify the check to just `Reline.respond_to?(: dialog_default_bg_color)` (or to be safe, check all 4 of them in the single if statement) because those APIs go hand-in-hand
1. We should provide defaults in `IRB.init_config`
    - I'm also thinking about providing 2 themes `dark` and `light` with jus black/white combinations so users don't need to pick colors one by one
1. We need some tests and documents around it

Thanks again for initiating these changes. Let me know if you want to do the extra work or you want someone else to take over it.

tompng:
Thank you for the pull request.
I'm closing this because Reline now has a dialog customization API `Reline::Face` released in Reline 0.4.0.
https://github.com/ruby/reline/pull/552 [document](https://github.com/ruby/reline/blob/master/doc/reline/face.md)

PR #338, opened 2022-01-17
require_relative can't be used for default gems' exe files

st0012:
The `exe` folder and `lib` folder of default gems don't locate under the same place. While `exe/irb` will be under the gem folder, `irb.rb` will be under `lib/ruby/VERSION/`.

So `require_relative` will make `irb` unuseable when shipped with Ruby.

Related discussion in the comments: https://github.com/ruby/irb/pull/335

### DIFF
--- a/exe/irb
+++ b/exe/irb
@@ -6,6 +6,6 @@
 #   	by Keiju ISHITSUKA(keiju@ruby-lang.org)
 #
 
-require_relative '../lib/irb'
+require "irb"
 
 IRB.start(__FILE__)


st0012:
@aycabta with `require "lib"`, how do you usually start `irb` that loads the code from project instead of system lib?

MSP-Greg:
I don't know the effects of this (other than the CLI issue), but it could be something like:
```ruby
begin
  require_relative '../lib/irb
rescue LoadError
  require 'irb'
end
```

st0012:
I'd avoid rescuing `LoadError` because if it's the the underlying code that caused it, the error will be ignored. And it'll be confusing and hard to debug the issue.
I'm thinking something like this but not sure if appending load path is acceptable here.

```rb
lib = File.expand_path("../../lib", __FILE__)
$:.unshift(lib) if Dir.exist?(lib)
require "irb"

IRB.start(__FILE__)
```

MSP-Greg:
> I'd avoid rescuing LoadError

What, me worry?  Actually, you're correct.  Maybe output something like "require_relative failed, using require", or never mind.

> appending load path is acceptable here

Bundler/Rubygems would normally do that, so probably not?

Again, I don't know the original reason, whether it's testing/ci or otherwise...

st0012:
> Bundler/Rubygems would normally do that, so probably not?

I think nowadays gem's `lib` is pushed to load path via rubygem and the `spec.require_paths` config. But in `irb`'s case, its gemspec won't be loaded when `exe/irb` is ran. So what you said probably doesn't happen.

PR #350, opened 2022-02-25
Refactor RubyLex#process_literal_type

Maumagnaguagno:
Simplify part of regex ``[_a-zA-Z0-9]`` with equivalent shorthand ``\w``.
Replace case-when with match ``$1`` or default value ``?"``, making intention more clear.


st0012 (approved):
Great simplification, thank you 👍

Issue #351, opened 2022-03-03
Default behavior change ( AUTOCOMPLETE, COLORIZE )


log69:
Dear Devs, I'd like to suggest to turn off autocomplete and colorization by default for the following reasons:

1) The autocomplete list is shown in color and cannot be read at all if the environment and desktop colors are set in a different way
2) The colorization has the same problem, because the terminal colors are preset and if the highlight scheme is not good then the text cannot be red (too bright yellow on white etc)

It would be the safest and sanest way to provide the most robust settings by default that has the smallest problem set. Thank You.

olleolleolle:
The NO_COLOR environment varaible, could that help this use-case?
https://github.com/ruby/irb/search?q=NO_COLOR

sshock:
This is an issue for me as well.  The colorization often slows everything to a crawl, which can be especially painful when it happens in the middle of a paste.  And the autocomplete is a bright cyan bg color that is completely unreadable.

These are both awesome features to be sure, but I have to agree they should be off by default!  I manage a lot of servers and now I have to go configure .irbrc in the way @log69 described on each one of them to disable these.

kaiquekandykoga:
Perhaps already known, but I would like to complement saying that it is also possible to initialize IRB disabling autocomplete and colorize with options.

irb --noautocomplete --nocolorize

wlipa:
Definitely agree.  The autocomplete UI is invasive and distracting.  It should be something you opt into rather than something turned on for you.

csutter:
Yes please – really struggling with this when using IRB through the Rails console on deployed applications.

It's also incredibly slow and inaccurate in the context of a large codebase...

DanielLemky:
Agreed. The autocomplete suggestions often cause the terminal window to scroll up so that the current line jumps to the top of of the terminal window due to the number of suggestions that appear, particularly when using the terminal in the bottom of an IDE where the window height is limited. This becomes a nuisance when reading from a previous line and all of a sudden the window jumps up due to numerous suggestions being shown.

yvasilkov:
Agree. The autocomplete considerably slows down typing, while options list is updating the 'Delete' (on Mac, same as Backspace) keystrokes are not captured.

st0012:
As something that'll be run on multiple platforms and any environment (dev, test, and prod), I think supporting 

> The NO_COLOR environment varaible, could that help this use-case? https://github.com/ruby/irb/search?q=NO_COLOR
> 
[...]

PR #362, opened 2022-05-09
Commands should respect `USE_COLORIZE` config

st0012:
Currently, colored commands `ls` and `show_source` doesn't respect the `USE_COLORIZE` config:

截圖 2022-05-09 12 04 28

This PR addresses the issue and here's the result:

截圖 2022-05-09 12 07 48

截圖 2022-05-09 12 11 44

BTW as suggested in https://github.com/ruby/irb/issues/351#issuecomment-1058205478, I think `irb` should adopt the `NO_COLOR` env var for easier no-color config.

### DIFF
--- a/lib/irb/cmd/ls.rb
+++ b/lib/irb/cmd/ls.rb
@@ -10,7 +10,7 @@ module IRB
   module ExtendCommand
     class Ls < Nop
       def execute(*arg, grep: nil)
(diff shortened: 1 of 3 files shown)


st0012:
@peterzhu2118 Would you mind giving it a look? Thx

Issue #368, opened 2022-05-29
Crash with none array `methods` variable or method.


ksss:
## Description

When I run the following code, irb crashes and exits when I enter any key.

```rb
$ cat t.rb
#! /usr/bin/env ruby

class Foo
  def a(methods)
    binding.irb
  end
end

Foo.new.a(1)
```

```
$ ruby t.rb

From: t.rb @ line 5 :

    1: #! /usr/bin/env ruby
    2:
    3: class Foo
    4:   def a(methods)
 => 5:     binding.irb
    6:   end
    7: end
    8:
    9: Foo.new.a(1)
# Pressing any key...
irb(#<Foo:0x0000000100b0ec20>):001:0> a(eval):1:in `|': Array can't be coerced into Integer (TypeError)
	from (eval):1:in `a'
	from /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/irb-1.4.1/lib/irb/completion.rb:362:in `eval'
	from /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/irb-1.4.1/lib/irb/completion.rb:362:in `retrieve_completion_data'
	from /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/irb-1.4.1/lib/irb/completion.rb:130:in `block in <module:InputCompletor>'
	from /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline/line_editor.rb:1681:in `call_completion_proc_with_checking_args'
	from /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline/line_editor.rb:543:in `call_completion_proc_with_checking_args'
	from /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline.rb:226:in `block in <class:Core>'
	from /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline/line_editor.rb:588:in `instance_exec'
	from /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline/line_editor.rb:588:in `call'
[...]

PR #369, opened 2022-05-30
Change to explicit method call in completion

osyo-manga:
Ensure that methods are called even when local variables are defined.
see: https://github.com/ruby/irb/issues/368

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -11,7 +11,29 @@ require_relative 'ruby-lex'
 
 module IRB
   module InputCompletor # :nodoc:
+    using Module.new {
+      refine ::Binding do
+        def eval_methods
+          eval("::Kernel.instance_method(:methods).bind(self).call")
+        end
 
+        def eval_private_methods
+          eval("::Kernel.instance_method(:private_methods).bind(self).call")
+        end
+
+        def eval_instance_variables
+          eval("::Kernel.instance_method(:instance_variables).bind(self).call")
+        end
+
+        def eval_global_variables
+          eval("::Kernel.instance_method(:global_variables).bind(self).call")
+        end
+
+        def eval_class_constants
+          eval("::Module.instance_method(:constants).bind(self.class).call")
+        end
+      end
+    }
 
     # Set of reserved words used by Ruby, you should not use these for
     # constants or variables
@@ -296,10 +318,10 @@ module IRB
         sep = $2
         message = $3
 
-        gv = eval("send(:global_variables)", bind).collect{|m| m.to_s}.push("true", "false", "nil")
-        lv = eval("send(:local_variables)", bind).collect{|m| m.to_s}
-        iv = eval("self.instance_variables", bind).collect{|m| m.to_s}
-        cv = eval("self.class.constants", bind).collect{|m| m.to_s}
+        gv = bind.eval_global_variables.collect{|m| m.to_s}.push("true", "false", "nil")
+        lv = bind.local_variables.collect{|m| m.to_s}
+        iv = bind.eval_instance_variables.collect{|m| m.to_s}
+        cv = bind.eval_class_constants.collect{|m| m.to_s}
 
         if (gv | lv | iv | cv).include?(receiver) or /^[A-Z]/ =~ receiver && /\./ !~ receiver
           # foo.func and foo is var. OR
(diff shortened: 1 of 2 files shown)


lib/irb/completion.rb
```diff
@@ -296,9 +296,9 @@ def self.retrieve_completion_data(input, bind: IRB.conf[:MAIN_CONTEXT].workspace
         sep = $2
         message = $3
 
-        gv = eval("global_variables", bind).collect{|m| m.to_s}.push("true", "false", "nil")
-        lv = eval("local_variables", bind).collect{|m| m.to_s}
-        iv = eval("instance_variables", bind).collect{|m| m.to_s}
+        gv = eval("send(:global_variables)", bind).collect{|m| m.to_s}.push("true", "false", "nil")
+        lv = eval("send(:local_variables)", bind).collect{|m| m.to_s}
```
osyo-manga, line 300: Under Ruby 2.6, private methods cannot be called with `self.`, so `#send` is used.

ksss:
Thanks for the support.
Would it be possible to support if the `methods` method is overridden?

```rb
class Foo
  def methods
    binding.irb
    1
  end
end

Foo.new.methods
```

If we can use `UnboundMethod#bind` to call the method before the overwrite, it may be possible to handle this, but I don't know if it can be done...

osyo-manga:
Thanks for comment.
Supported https://github.com/ruby/irb/pull/369/commits/b01b38d525e0515f060640f017121b15f59a71e1.

test/irb/test_completion.rb
```diff
@@ -107,11 +107,56 @@ def test_complete_require_relative
...
+      private_methods = ''
+      methods = ''
+      global_variables = ''
+      local_variables = ''
+      instance_variables = ''
       str_example = ''
-      str_example.clear # suppress "assigned but unused variable" warning
+      @str_example = ''
+
+       # suppress "assigned but unused variable" warning
```
st0012, line 119:
```suggestion
      # suppress "assigned but unused variable" warning
```

Can we fix this comment's indentation?

test/irb/test_completion.rb
```diff
@@ -107,11 +107,56 @@ def test_complete_require_relative
...
     def test_complete_variable
+      # Bug fix issues https://github.com/ruby/irb/issues/368
+      private_methods = ''
+      methods = ''
+      global_variables = ''
+      local_variables = ''
+      instance_variables = ''
       str_example = ''
-      str_example.clear # suppress "assigned but unused variable" warning
+      @str_example = ''
```
st0012, line 117: It looks like `str_example` and `@str_example` are test subjects, and the rest are for making sure `irb` won't confuse them when completion?

If that's the case, can we add comments to mention that?

lib/irb/completion.rb
```diff
@@ -11,7 +11,29 @@
 
 module IRB
   module InputCompletor # :nodoc:
+    using Module.new {
+      refine ::Binding do
+        def eval_methods
+          eval("::Kernel.instance_method(:methods).bind(self).call")
```
st0012, line 17: Instead of wrapping the entire chain inside `eval`, can we only wrap the necessary part?

```rb
        def eval_methods
          ::Kernel.instance_method(:methods).bind(eval("self")).call
        end

        def eval_private_methods
          ::Kernel.instance_method(:private_methods).bind(eval("self")).call
        end

        def eval_instance_variables
          ::Kernel.instance_method(:instance_variables).bind(eval("self")).call
        end

        def eval_global_variables
          ::Kernel.instance_method(:global_variables).bind(eval("self")).call
        end

        def eval_class_constants
          ::Module.instance_method(:constants).bind(eval("self.class")).call
        end
```

st0012:
@osyo-manga I think the PR is almost ready. Can you do the small adjustments I mentioned above? Thx

st0012:
@osyo-manga Given that this issue crashes irb, I want to make sure it's fixed and tested before Ruby 3.2. So I opened https://github.com/ruby/irb/pull/404 based on your commits. Let me know if you prefer making those changes yourself, I'll close mine 🙂

lib/irb/completion.rb
osyo-manga: Thanks comments :)
I fixed it.
https://github.com/ruby/irb/pull/369/commits/78ef990c1ce5a4b82b7dbce8dc72277122579011

test/irb/test_completion.rb
osyo-manga: Thanks comment.
I added comment https://github.com/ruby/irb/pull/369/commits/697708da7ce8ac803dd54dcea5ac7fe23f1a306a.

osyo-manga:
@st0012 Sorry for the delay in replying.
PR has been updated.

st0012:
@osyo-manga Thank you 👍 
@peterzhu2118 Can you help me merge this? Thx

peterzhu2118 (approved):
Thank you for your contribution!

osyo-manga:
Thank for merge :)

PR #370, opened 2022-06-12
Additions/revisions to help-message

BurdetteLamar:
Changed:
- Added text to options that said just 'same as ruby -whatever'.
- Added defaults.
- Removed an errant tab.

I'm not adding the documentation label, because I'm not sure about how else this might be used.

If/when approved, section "Command line options" in irb.rb will need to be updated.

### DIFF
--- a/lib/irb/lc/help-message
+++ b/lib/irb/lc/help-message
@@ -10,52 +10,53 @@
 #
 #
 Usage:  irb.rb [options] [programfile] [arguments]
-  -f                Suppress read of ~/.irbrc
-  -d                Set $DEBUG to true (same as `ruby -d')
-  -r load-module    Same as `ruby -r'
-  -I path           Specify $LOAD_PATH directory
-  -U                Same as `ruby -U`
-  -E enc            Same as `ruby -E`
-  -w                Same as `ruby -w`
-  -W[level=2]       Same as `ruby -W`
+  -f                Don't initialize from configuration file.
+  -d                Set $DEBUG and $VERBOSE to true (same as 'ruby -d').
+  -r load-module    Require load-module (same as 'ruby -r').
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
@hsbt, I'm not sure whether @aycabta is fully back in action;  may I request a review from you, if you think that would be best?

lib/irb/lc/help-message
```diff
@@ -10,52 +10,53 @@
...
-  -U                Same as `ruby -U`
-  -E enc            Same as `ruby -E`
-  -w                Same as `ruby -w`
-  -W[level=2]       Same as `ruby -W`
+  -f                Don't initialize from configuration file.
+  -d                Set $DEBUG and $VERBOSE to true (same as 'ruby -d').
+  -r load-module    Require load-module (same as 'ruby -r').
+  -I path           Specify $LOAD_PATH directory (same as 'ruby -I').
+  -U                Set external and internal encodings to UTF-8.
+  -E ex[:in]        Set default external and internal encodings
```
olleolleolle, line 18: `-E enc` became `-E ex[:in]`, which is terse. Perhaps we could repeat these short forms in parentheses after the written-out words in the help text?

```suggestion
  -E ex[:in]        Set default external (en) and internal (in) encodings
```

BurdetteLamar:
@hsbt, I'm not sure that @aycabta is back in action.  Can you suggest someone to review this?  Thanks.

lib/irb/lc/help-message
peterzhu2118: I think there's a typo in this suggestion

```suggestion
  -E ex[:in]        Set default external (ex) and internal (in) encodings
```

BurdetteLamar:
This is now approved, but not by the maintainer.  Can it be merged?

peterzhu2118:
This is probably fine to merge.

PR #374, opened 2022-06-26
Centralize coloring control

st0012:
When deciding whether to color the output, we should always consider 2 things:

1. Coloring is supported on the current mode (tty) and platform
2. The user has coloring enabled

If one of them is `false`, we shouldn't color the output.

However, the current implementation only checks 1 by default and 2 needs an additional condition.
It makes code unnecessarily complicated and also error-prone. For example #362 was caused by missing 2.

So in this PR, I moved 2 into `Color.colorable?` (where 1 locates) and removes all the manual check on 2, as well as `Context#use_colorize`.

### Test Result

**Colorized**

截圖 2022-06-26 17 21 04

**Non-colorized**

(The weird floating "s"s were caused by my Neovim + Floaterm setup.)

截圖 2022-06-26 17 21 49

### DIFF
--- a/test/irb/test_context.rb
+++ b/test/irb/test_context.rb
@@ -179,7 +179,6 @@ module TestIRB
     end
 
     def test_default_config
-      assert_equal(true, @context.use_colorize?)
       assert_equal(true, @context.use_autocomplete?)
     end
 
--- a/test/irb/test_context.rb
+++ b/test/irb/test_context.rb
@@ -145,26 +145,37 @@ module TestIRB
       assert_equal "=> 1\n", out
     end
 
-    def test_eval_object_without_inspect_method
-      verbose, $VERBOSE = $VERBOSE, nil
-      all_assertions do |all|
-        IRB::Inspector::INSPECTORS.invert.each_value do |mode|
-          all.for(mode) do
-            input = TestInputMethod.new([
-                "[BasicObject.new, Class.new]\n",
-              ])
-            irb = IRB::Irb.new(IRB::WorkSpace.new(Object.new), input)
-            irb.context.inspect_mode = mode
-            out, err = capture_output do
-              irb.eval_input
-            end
-            assert_empty err
-            assert_match(/\(Object doesn't support #inspect\)\n(=> )?\n/, out)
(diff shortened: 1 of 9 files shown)


st0012:
cc @peterzhu2118

test/irb/test_context.rb
```diff
@@ -145,6 +145,48 @@ def test_output_to_pipe
       assert_equal "=> 1\n", out
     end
 
+    def test_eval_object_without_inspect_method
+      verbose, $VERBOSE = $VERBOSE, nil
+      all_assertions do |all|
+        IRB::Inspector::INSPECTORS.invert.each_value do |mode|
```
peterzhu2118, line 151: I think this is equivalent

```suggestion
        IRB::Inspector::INSPECTORS.each_key do |mode|
```

test/irb/test_context.rb
```diff
@@ -167,8 +209,8 @@ def test_eval_object_without_inspect_method
       $VERBOSE = verbose
     end
 
+
```
peterzhu2118, line 212: There's an extra newline here.

test/irb/test_context.rb
```diff
@@ -145,30 +145,40 @@ def test_output_to_pipe
...
+        [:marshal, "123", Marshal.dump(123)],
+      ],
+      failed: [
+        [false, "BasicObject.new", /\(Object doesn't support #inspect\)\n(=> )?\n/],
+        [:p, "class Foo; undef inspect ;end; Foo.new", /\(Object doesn't support #inspect\)\n(=> )?\n/],
+        [true, "BasicObject.new", /\(Object doesn't support #inspect\)\n(=> )?\n/],
+        [:yaml, "BasicObject.new", /\(Object doesn't support #inspect\)\n(=> )?\n/],
+        [:marshal, "[Object.new, Class.new]", /\(Object doesn't support #inspect\)\n(=> )?\n/]
+      ]
+    }.each do |scenario, cases|
```
st0012, line 163: @peterzhu2118 I ended up merging all cases together as the test body is the same. Wdyt?

PR #377, opened 2022-07-09
[DOC] Include updated help message

BurdetteLamar:


peterzhu2118 (commented):
We can also change `lib/irb/lc/ja/help-message` which will let us simplify the `IRB.print_usage` method.

BurdetteLamar:
> We can also change `lib/irb/lc/ja/help-message` which will let us simplify the `IRB.print_usage` method.

File changed.  But the work on the method needs to be a separate (non-doc) PR.

lib/irb/lc/ja/help-message
```diff
@@ -54,4 +44,3 @@ Usage:  irb.rb [options] [programfile] [arguments]
   -h, --help	    irb のヘルプを表示する.
   --		    以降のコマンドライン引数をオプションとして扱わない.
```
peterzhu2118, line 46: There's an extra newline here.

BurdetteLamar: Fixed.

peterzhu2118 (approved):
Looks good!

PR #380, opened 2022-07-23
[PoC] Support dialog theme/color configuration

st0012:
See https://bugs.ruby-lang.org/issues/19010

### Todos

If we want to proceed on this idea, we have the following todos:

- [ ] Add test cases
- [ ] Release `reline 0.4.0` for the new APIs
- [ ] Bump `reline` requirement and drop Ruby 2.5 support because `reline` now requires 2.6+
    - [ ] This includes CI changes
- [ ] Add documentation
- [ ] Automatic background detection to toggle bright theme

### DIFF
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -144,6 +144,36 @@ module IRB
       if @newline_before_multiline_output.nil?
         @newline_before_multiline_output = true
       end
+
+      if @io.is_a?(ReidlineInputMethod) && Reline.respond_to?(:dialog_default_bg_color)
+        configure_dialog_colors
+      end
+    end
+
+    def configure_dialog_colors
+      @use_bright_theme = IRB.conf[:USE_BRIGHT_THEME]
+
+      if @use_bright_theme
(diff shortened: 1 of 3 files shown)


st0012:
cc @peterzhu2118 @nobu @hsbt

lib/irb/context.rb
```diff
@@ -140,6 +140,37 @@ def initialize(irb, workspace = nil, input_method = nil)
...
+        @dialog_default_bg_color   = :black
+        @dialog_default_fg_color   = :white
+        @dialog_highlight_bg_color = :white
+        @dialog_highlight_fg_color = :black
+      when :light
+        @dialog_default_bg_color   = :white
+        @dialog_default_fg_color   = :black
+        @dialog_highlight_bg_color = :black
+        @dialog_highlight_fg_color = :white
+      end
```
peterzhu2118, line 163: We should have an `else` case that raises an error.

peterzhu2118 (commented):
:+1: I like having themes. Makes the UX better.

PR #381, opened 2022-07-24
Add latest version of reline into test matrix

st0012:
When adopting newest reline features (e.g. https://github.com/ruby/irb/pull/380), I find it hard to know what failures are caused by the `reline` upgrade and what's caused by the code change. [Example build](https://github.com/ruby/irb/actions/runs/2723567680).

And if we make the CI to run against the latest `reline`, we will lose the coverage on currently released `reline` version.

So I think we can make the CI run with both "latest release" and "master" to make such changes easier and safer.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   irb:
-    name: irb (${{ matrix.ruby }} / ${{ matrix.os }})
+    name: rake test ${{ matrix.ruby }} ${{ matrix.with_latest_reline && '(latest reline)' || '' }}
     strategy:
       matrix:
         ruby:
@@ -20,8 +20,11 @@ jobs:
           - head
           - truffleruby-head
         os: [ubuntu-latest]
+        with_latest_reline: [true, false]
       fail-fast: false
     runs-on: ${{ matrix.os }}
+    env:
+      WITH_LATEST_RELINE: ${{matrix.with_latest_reline}}
     steps:
     - uses: actions/checkout@master
     - name: Set up Ruby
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   irb:
-    name: irb (${{ matrix.ruby }} / ${{ matrix.os }})
+    name: rake test ${{ matrix.ruby }} ${{ matrix.with_latest_reline && '(latest reline)' || '' }}
```
st0012, line 11: We only run against `ubuntu-latest` so I think it's not necessary to indicate that in job names.

.github/workflows/test.yml
```diff
@@ -20,6 +20,11 @@ jobs:
           - head
           - truffleruby-head
         os: [ubuntu-latest]
+        with_latest_reline: [true, false]
+        exclude:
+          - ruby: '2.5'
```
st0012, line 25: I plan to drop the support of Ruby 2.5 because it won't be supported by the next `reline` release.

st0012:
cc @peterzhu2118 @hsbt

.github/workflows/test.yml
```diff
@@ -29,15 +34,19 @@ jobs:
       with:
         ruby-version: ${{ matrix.ruby }}
     - name: Install dependencies
+      env: 
+        WITH_LATEST_RELINE: ${{matrix.with_latest_reline}}
```
peterzhu2118, line 38: We can use a top level `env` so that every step has access to this environment variable (see [documentation](https://docs.github.com/en/actions/learn-github-actions/environment-variables)).

.github/workflows/test.yml
peterzhu2118: We can probably drop support for 2.6 and earlier completely since they aren't supported Rubies anymore.

st0012: I don't mind keeping 2.6 available until the day we need to make workaround for it

peterzhu2118: We can probably completely drop support for 2.5, it EOL'd in March 2021, almost 1.5 years ago.

st0012: It requires other changes like gemspec so I’ll drop it in another PR 👍

peterzhu2118 (approved):

Issue #328, opened 2022-01-08
Change default completion background color


ianks:
## Description

When starting IRB in Ruby 3.1, autocomplete suggestions are shown but hard to read due to the default background color. Ideally, the default background color settings would ensure enough contrast to make the text legible.

<img width="246" alt="Screen Shot 2022-01-08 at 12 22 23 PM" src="https://user-images.githubusercontent.com/3303032/148653612-e3dff786-1a10-4923-a0eb-3975cae10a7f.png">

## Result of irb_info

Please paste the result of `irb_info` command in IRB.

ex.)

```
irb(main):001:0> irb_info
=>
Ruby version: 3.1.0                            
IRB version: irb 1.4.1 (2021-12-25)            
InputMethod: ReidlineInputMethod with Reline 0.3.0
RUBY_PLATFORM: arm64-darwin21                  
LANG env: en_US.UTF-8                          
East Asian Ambiguous Width: 1                  
```

## Terminal Emulator

VSCode terminal

## Setting Files

> Are you using `~/.irbrc` and `~/.inputrc`?

No I am not

ianks:
I believe https://github.com/ruby/irb/issues/328 would allow for this

kaiquekandykoga:
> I believe #328 would allow for this

The link provided is redirecting to the same page.

osyo-manga:
This?
https://github.com/ruby/reline/pull/413

sshock:
Making it configurable would be great!  Looking forward to https://github.com/ruby/reline/pull/413 and https://github.com/ruby/irb/pull/337 being approved and merged!

But the default should be fixed too.  White text on a bright cyan background is just plain awful.  Can we at least make the background dark cyan?

pocari:
Having made the [changes](https://github.com/ruby/reline/pull/413/commits/9bf4979ecd53f3a22909491e150f351d727b0e2a) as [advised](https://github.com/ruby/reline/pull/413#discussion_r874176336) by @St0012, it is now simple to change the default background and foreground color. Now it is just a matter of specifying the appropriate value in the block that sets the default value.

ima1zumi:
Fixed https://github.com/ruby/reline/pull/552

PR #383, opened 2022-07-25
Drop Ruby 2.5 support

st0012:
Because Ruby 2.5 has reached EOL for more than 1.5 years and it won't be supported by the next `reline` version either.

### DIFF
--- a/test/lib/core_assertions.rb
+++ b/test/lib/core_assertions.rb
@@ -694,8 +694,7 @@ eom
         if !errs.empty?
           msg = "exceptions on #{errs.length} threads:\n" +
             errs.map {|t, err|
-            "#{t.inspect}:\n" +
-              RUBY_VERSION >= "2.5.0" ? err.full_message(highlight: false, order: :top) : err.message
+            "#{t.inspect}:\n" + err.full_message(highlight: false, order: :top)
           }.join("\n---\n")
           if message
             msg = "#{message}\n#{msg}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,15 +16,10 @@ jobs:
           - '3.0'
           - '2.7'
           - '2.6'
-          - '2.5'
           - head
           - truffleruby-head
         os: [ubuntu-latest]
         with_latest_reline: [true, false]
-        exclude:
-          - ruby: '2.5'
-            os: ubuntu-latest
-            with_latest_reline: true
       fail-fast: false
     runs-on: ${{ matrix.os }}
     env:
--- a/irb.gemspec
+++ b/irb.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.required_ruby_version = Gem::Requirement.new(">= 2.5")
+  spec.required_ruby_version = Gem::Requirement.new(">= 2.6")
 
   spec.add_dependency "reline", ">= 0.3.0"
 end
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -596,11 +596,7 @@ module IRB
 
       if exc.backtrace
         order = nil
-        if '2.5.0' == RUBY_VERSION
-          # Exception#full_message doesn't have keyword arguments.
-          message = exc.full_message # the same of (highlight: true, order: bottom)
-          order = :bottom
-        elsif '2.5.1' <= RUBY_VERSION && RUBY_VERSION < '3.0.0'
+        if RUBY_VERSION < '3.0.0'
           if STDOUT.tty?
             message = exc.full_message(order: :bottom)
             order = :bottom
--- a/test/irb/test_color.rb
+++ b/test/irb/test_color.rb
@@ -125,23 +125,19 @@ module TestIRB
       end
 
       tests.each do |code, result|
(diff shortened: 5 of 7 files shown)


st0012:
cc @peterzhu2118

test/lib/core_assertions.rb
```diff
@@ -694,8 +694,7 @@ def assert_join_threads(threads, message = nil)
         if !errs.empty?
           msg = "exceptions on #{errs.length} threads:\n" +
             errs.map {|t, err|
-            "#{t.inspect}:\n" +
-              RUBY_VERSION >= "2.5.0" ? err.full_message(highlight: false, order: :top) : err.message
+            "#{t.inspect}:\n" + err.full_message(highlight: false, order: :top)
           }.join("\n---\n")
```
peterzhu2118, line 698: I realize that this isn't the fault of this PR, but the indentation of these lines look odd. Could you fix it?

peterzhu2118 (approved):
Looks good!

hsbt:
Is there any reason why remove Ruby 2.5 without EOL status? I prefer to keep the old Ruby until we use new feature of the recent Ruby versions or hard to maintain CI status.

peterzhu2118:
I'm not sure what you mean. Ruby 2.5 was EOL as of March 2021. CI currently doesn't run with IRB master + reline master because tests fail, so the next release of reline will make tests fail.

hsbt:
@peterzhu2118 I'm not against this change strongly. But I have a concern that you and @st0012 change reline and irb with the incompatible changes like the support version without maintainer who are @aycabta approval.

I know @aycabta is inactive today. But we should wait for him first. Can you wait until RubyKaigi 2022?

st0012:
@hsbt To be clear, I didn't drop the 2.5 support on `reline`. It's done in [this commit](https://github.com/ruby/reline/commit/dae9eca3233d19f53fe61b773c50a8be7775c1d9), which is required by `relines`'s own dependency `io-console`.
So I'm just making sure `irb` will work with its own dependencies.

st0012:
@hsbt I can understand your concern on the incompatible changes and I want to make it clear that

- I have no intention to disrespect @aycabta's great work nor the fact that he's these projects' maintainer
- I don't want to design and push these changes on my own either. That's why in my `reline` PRs and #380 I also asked you and @nobu's opinions

I also want to explain why I'm pushing the recent changes now:

1. The background color issue on autocompletion affects many developers. There are at least 2 issues opened (#328 and #351) with a dozen of comments talking about it. I also have seen multiple "tips" about disabling the feature altogether because it's distracting to developers ([example](https://twitter.com/sdogruyol/status/1538512030449254400)). And I'm also bothered by this issue.
1. The `reline` APIs we recently introduced to address the issue were [originally proposed last December](https://github.com/ruby/reline/pull/413). And it took us more than 2 months (from May to July) to review it and polish the APIs in follow up PRs.
1. So if we continue the work with the current pace, we'll probably be able to solve the issue (release `reline`, adopt the APIs in `irb`, finish documents...etc.) by October. And that only leaves us 2 months to let the community test it before we cut a new `irb` release and make it ship with Ruby 3.2.

So I hope you can understand that:

1. @peterzhu2118 and I only focus on this issue and try to make as few as breaking changes as possible. As I said, this PR is just to make `irb` support the same Ruby version as `reline` and `io-console`.
2. We are not rushing it. We just want to make sure we have time to make the solution great and tested for Ruby 3.2.

So back to your question:

> I know [@aycabta](https://github.com/aycabta) is inactive today. But we should wait for him first. Can you wait until RubyKaigi 2022?

Yes we can wait, including this PR, #380 and  `reline`'s release. 

But IMO, a better option is to have a solution in `irb` master by RubyKaigi and ask Rubyists to test it and help us improve the solution before the Ruby release.

hsbt:
@st0012 Thanks for clarification. I meant https://github.com/ruby/reline/pull/456#issuecomment-1192650048 as incompatible change.




> Yes we can wait, including this PR, #380 and `reline`'s release.

Also, Thanks. You and @peterzhu2118 's work are great too.

st0012:
> I meant https://github.com/ruby/reline/pull/456#issuecomment-1192650048 as incompatible change.

@hsbt I don't think that's an incompatible change because the original APIs haven't been released yet?

hsbt:
Ah, I see.

st0012:
@hsbt Since we have now clarified that:

1. We didn't drop Ruby 2.5 unnecessarily
2. We didn't make any incompatible changes

Do we still have to wait for @aycabta? As I said, we can wait, but I also want to see if there's any alternatives than just waiting. 

For example, if you worry about us not knowing what not to change, I can reassure you that we are fully aware of the importance of compatibility as shown in the above comments.

Or if you worry about @aycabta won't be able to keep up with the changes, I'll be happy to document changes happened after this Feb when he's back (including changes made by other contributors). I'm happy to have a call to walk through them as well.

I think the best way to make great tools is to keep the development and feedback loop going. Because many user experience issues like this take time to surface and solutions also take time to experiment/polish.

hsbt:
> 1. We didn't drop Ruby 2.5 unnecessarily

Yes. But It's IMO.

> 2\. We didn't make any incompatible changes

I say it's maintainer's convenience. And my mainly concerns are you and @peterzhu2118 changes irb and reline without @aycabta 's opinion in this year.

hsbt:
Unfotunately, We couldn't get @aycabta 's opinion. I decide to coordinate the next irb release for Ruby 3.2 final release.

PR #385, opened 2022-08-04
Fix indent after multiline string

tompng:
This pullreq fixes two indent problem with multiline string.

### Fixed indent after `"stri\ng"`, `%[stri\ng]` `%w[string\nstring]`
```
# Case1
irb(main):001:0> if 0
irb(main):002:1"   "1
irb(main):003:1"   2"
irb(main):004:1* 3
expected:          3

# Case2
irb(main):001:1* if 0
irb(main):002:2"   p("1
irb(main):003:2"     2",
irb(main):004:2*   3,
expected:            3,
```

Multiline string can contain `"\n"`, but the next token does not matches `/(\A +)/`.
Skipped updating prev_spaces if token is :on_tstring_content or :on_words_sep

###  Fixed calculation of corresponding_token_depth after heredoc
```
irb(main):001:0> if 1
irb(main):002:1"   <<~A
irb(main):003:1"   A
irb(main):004:1*   p(
irb(main):005:3*     p(
irb(main):006:4*     )
irb(main):007:1* )
expected:          )
```
and after embdoc
```
irb(main):001:1* if 1
irb(main):002:1* =begin
irb(main):003:1* =end
irb(main):004:2*   p(
irb(main):005:3*     p(
irb(main):006:2*     )
irb(main):007:1* )
expected:          )
```

Added `:on_heredoc_end` and `:on_embdoc_end`. It also ends with `"\n"`.
```ruby
# all tokens that include "\n" in ruby/ruby source code
Dir.glob('**/*.rb').flat_map do |file|
  RubyLex.ripper_lex_without_warning(File.read(file), context: context).select { _1.tok.include? $/ }
end.map(&:event).uniq
# => [:on_comment, :on_nl, :on_ignored_nl, :on_tstring_content, :on_heredoc_end, :on_words_sep, :on_embdoc_beg, :on_embdoc, :on_embdoc_end, :on_sp, :on___end__]
```

### DIFF
--- a/test/irb/test_ruby_lex.rb
+++ b/test/irb/test_ruby_lex.rb
@@ -697,5 +697,68 @@ module TestIRB
         assert_equal('<<A', string_literal&.tok)
       end
     end
+
+    def test_corresponding_token_depth_with_heredoc_and_embdoc
+      reference_code = <<~EOC.chomp
+        if true
+          hello
+          p(
+          )
+      EOC
+      code_with_heredoc = <<~EOC.chomp
+        if true
+          <<~A
+          A
+          p(
+          )
+      EOC
+      code_with_embdoc = <<~EOC.chomp
+        if true
+        =begin
+        =end
+          p(
+          )
+      EOC
+      [reference_code, code_with_heredoc, code_with_embdoc].each do |code|
+        lex = RubyLex.new
+        lines = code.lines
+        lex.instance_variable_set('@tokens', RubyLex.ripper_lex_without_warning(code))
+        assert_equal 2, lex.check_corresponding_token_depth(lines, lines.size)
+      end
+    end
+
+    def test_find_prev_spaces_with_multiline_literal
+      lex = RubyLex.new
+      reference_code = <<~EOC.chomp
+        if true
+          1
+          hello
+          1
+          world
+        end
+      EOC
+      code_with_percent_string = <<~EOC.chomp
+        if true
+          %w[
+            hello
+          ]
+          world
+        end
+      EOC
+      code_with_quoted_string = <<~EOC.chomp
+        if true
+          '
+            hello
+          '
+          world
+        end
+      EOC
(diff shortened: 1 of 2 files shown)


test/irb/test_ruby_lex.rb
```diff
@@ -632,5 +632,34 @@ def test_unterminated_code
...
+        lines = code.lines
+        lex.instance_variable_set('@tokens', RubyLex.ripper_lex_without_warning(code))
+        assert_equal 2, lex.check_corresponding_token_depth(lines, lines.size)
+      end
+    end
+
+    def test_find_prev_spaces_with_multiline_literal
+      lex = RubyLex.new
+      codes = [
+        "if true\n  1\n  hello\n  1\n  world\nend",
```
st0012, line 653: I think to most people it's not easy to newline and spaces in their head. Do you think we can rewrite test cases into something like

```rb
            %[
if true
  1
  hello
  1
  world
end
            ].strip,
```

to improve its readability?

(Note: we can just put `strip` in the `each` block if you think the new format is better)

test/irb/test_ruby_lex.rb
```diff
@@ -632,5 +632,34 @@ def test_unterminated_code
         assert_empty(error_tokens, 'Error tokens must be ignored if there is corresponding non-error token')
       end
     end
+
+    def test_corresponding_token_depth_with_heredoc_and_embdoc
+      codes = [
+        "if true\n  hello\n  p(\n    p(\n    )\n  )",
```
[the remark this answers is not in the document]
tompng: I simplified it

test/irb/test_ruby_lex.rb
tompng: I agree that `"\n  \n  ..."` is not so readable. I rewrite it using `<<~EOC`.

st0012 (approved):
Sorry for the delay and thank you for the fix 👍

PR #387, opened 2022-08-05
Update regarding NO_COLOR value

nobu:
https://no-color.org has been updated (jcs/no_color#83):

> Command-line software which adds ANSI color to its output by default
should check for a `NO_COLOR` environment variable that, when present
and **not an empty string** (regardless of its value), prevents the
addition of ANSI color.

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -44,7 +44,7 @@ module IRB # :nodoc:
     @CONF[:IRB_RC] = nil
 
     @CONF[:USE_SINGLELINE] = false unless defined?(ReadlineInputMethod)
-    @CONF[:USE_COLORIZE] = !ENV['NO_COLOR']&.slice(0)
+    @CONF[:USE_COLORIZE] = (nc = ENV['NO_COLOR']).nil? || nc.empty?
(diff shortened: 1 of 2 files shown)


lib/irb/init.rb
```diff
@@ -44,7 +44,7 @@ def IRB.init_config(ap_path)
     @CONF[:IRB_RC] = nil
 
     @CONF[:USE_SINGLELINE] = false unless defined?(ReadlineInputMethod)
-    @CONF[:USE_COLORIZE] = !ENV['NO_COLOR']
+    @CONF[:USE_COLORIZE] = !ENV['NO_COLOR']&.slice(0)
```
st0012, line 47: This is a brilliant way to check emptiness! But it took me a while to comprehend because of usage of `!` + `slice`. 

So personally I still prefer the common idioms like

```suggestion
    @CONF[:USE_COLORIZE] = (nc = ENV['NO_COLOR']).nil? || nc.empty?
```

st0012 (commented):
Just a small suggestion, but I think it's already good to merge 👍

lib/irb/init.rb
nobu: That condition is inverted.

PR #389, opened 2022-08-05
Fix history file saving with concurrent irb sessions when history file doesn't exist

### LINKED ISSUE #388: History file being overidden if initially empty
## Description

I'm running multiple containers via Docker Compose with all of them mapped to a single .irb-history file which is placed on Docker volume. When they are launched, the (last?) one overrides the history.

Consider following example:

- Dockerfile:
  ```dockerfile
  FROM ruby:alpine
  RUN echo "IRB.conf[:HISTORY_FILE] = ENV['IRB_HISTORY_FILE'] if ENV['IRB_HISTORY_FILE']" >> ~/.irbrc
  ```

- docker-compose.yaml:
  ```yaml
  services:
    first:
      build:
        context: .
      volumes:
        - history:/root/.history
      environment:
        HISTFILE: /root/.history/.bash_history
        IRB_HISTORY_FILE: /root/.history/.irb-history
      command: tail -f /dev/null
    second:
      extends:
        service: first
  volumes:
    history:
  ```

- run:
  ```console
  docker compose up -d
  ```

  run in another terminal:
  ```console
  docker compose exec first irb
  puts 'first'
  ```

  run in another terminal:
  ```console
  docker compose exec second irb
  puts 'second'
  ```
  run in first terminal:
  ```console
  docker compose down
  ```

  run in first terminal:
  ```console
  docker compose run first cat /root/.history/.irb-history
  ```

  result:
  ```console
  puts 'second'
  ```

## Result of irb_info

```
irb(main):001:0> irb_info
=> 
Ruby version: 3.1.2                            
IRB version: irb 1.4.1 (2021-12-25)            
InputMethod: ReidlineInputMethod with Reline 0.3.0
.irbrc path: /root/.irbrc                      
RUBY_PLATFORM: aarch64-linux-musl              
LANG env: C.UTF-8                              
East Asian Ambiguous Width: 1 
```

## Setting Files

`~/.irbrc`:

```ruby
IRB.conf[:HISTORY_FILE] = ENV['IRB_HISTORY_FILE'] if ENV['IRB_HISTORY_FILE']
```

jeremyevans:
If history file didn't exist when irb was started, @loaded_history_mtime
would be nil.  However, if the history file didn't exist before, but it
exists when saving history, that means the history file was modified,
and we should handle it the same way as we handle the other case where
the history file was modified.

Fixes #388

### DIFF
--- a/lib/irb/ext/save-history.rb
+++ b/lib/irb/ext/save-history.rb
@@ -107,7 +107,7 @@ module IRB
           raise
         end
 
-        if File.exist?(history_file) && @loaded_history_mtime &&
+        if File.exist?(history_file) &&
            File.mtime(history_file) != @loaded_history_mtime
           history = history[@loaded_history_lines..-1]
           append_history = true
--- a/lib/irb/ext/save-history.rb
+++ b/lib/irb/ext/save-history.rb
@@ -109,7 +109,7 @@ module IRB
 
         if File.exist?(history_file) &&
            File.mtime(history_file) != @loaded_history_mtime
-          history = history[(@loaded_history_lines||0)..-1]
+          history = history[@loaded_history_lines..-1] if @loaded_history_lines
           append_history = true
         end
 
--- a/lib/irb/ext/save-history.rb
+++ b/lib/irb/ext/save-history.rb
@@ -109,7 +109,7 @@ module IRB
 
         if File.exist?(history_file) &&
            File.mtime(history_file) != @loaded_history_mtime
-          history = history[@loaded_history_lines..-1]
+          history = history[(@loaded_history_lines||0)..-1]
           append_history = true
         end
 
--- a/test/irb/test_history.rb
+++ b/test/irb/test_history.rb
@@ -158,6 +158,27 @@ module TestIRB
       end
     end
 
+    def test_history_concurrent_use_not_present
+      omit "Skip Editline" if /EditLine/n.match(Readline::VERSION)
+      IRB.conf[:SAVE_HISTORY] = 1
+      assert_history(<<~EXPECTED_HISTORY, nil, <<~INPUT) do |history_file|
+        exit
+        5
+        exit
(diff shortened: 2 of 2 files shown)


jeremyevans:
@st0012 Are you comfortable reviewing/merging this?

test/irb/test_history.rb
```diff
@@ -158,6 +158,27 @@ def test_history_concurrent_use
       end
     end
 
+    def test_history_concurrent_use_not_present
```
st0012, line 161: I'm not sure if this case covers the fix. I ran it against the old implementation and it still passed.

lib/irb/ext/save-history.rb
```diff
@@ -107,9 +107,9 @@ def save_history
           raise
         end
 
-        if File.exist?(history_file) && @loaded_history_mtime &&
+        if File.exist?(history_file) &&
            File.mtime(history_file) != @loaded_history_mtime
-          history = history[@loaded_history_lines..-1]
+          history = history[(@loaded_history_lines||0)..-1]
```
st0012, line 112: How about 

```rb
history = history[@loaded_history_lines..-1] if @loaded_history_lines
```

because `history[0..-1]` basically means the same as `history`?

And I think it also express the intention clearer. wdyt?

test/irb/test_history.rb
jeremyevans: I couldn't think of a way to get it to pass with the current `assert_history` helper method, so I changed the test to be more direct, loading the history without the history file existing, writing the history file, then saving the history.  I checked that this failed previously and now passes.

lib/irb/ext/save-history.rb
jeremyevans: Makes sense to me.  I included that change in the commit that fixes the test.

st0012 (commented):
@jeremyevans I think it looks good and can be merged 👍

PR #390, opened 2022-08-07
Fix code terminated check with heredoc and backtick

tompng:
Fixes #361 and also fixes similar problem with ``def `();end`` ``self.`('sleep 1')`` (#334)

### Heredoc
This code shows a sample of finding heredoc_beg corresponding to heredoc_end.
```ruby
<<A+<<B+%[#{<<C+<<D
#{<<E+<<F
#{<<G+<<H
G
H
}
E
F
}
A
B
C
D
}]
```
In line 1, there are heredocs(A,B,C,D). These heredocs will start from the next line. (pushed into `pending_heredoc`)
Heredoc will be closed in an order `A → B → C → D`, so it should be reversed and pushed to start_token when reached a new line.

### Backtick
```
irb(main):001:0` def `();end
irb(main):002:0` 
irb(main):003:0` self.`('sleep 1')
irb(main):004:0` 
# expect
irb(main):001:0> def `(); end
=> :`
irb(main):002:0> self.`('sleep 1')
=> ""
```

Checks if backtick is `` `command` `` or not.
```ruby
`sleep 1`         #<Ripper::Lexer::Elem: on_backtick@1:0 BEG token: "`">
self.`('sleep 1') #<Ripper::Lexer::Elem: on_backtick@1:5 ARG token: "`">
def `(arg); end   #<Ripper::Lexer::Elem: on_backtick@1:4 ENDFN token: "`">
```

### lexer.parse.sort_by(&:pos)
Test that I added failed in ruby 2.5.0 and 2.6.0 without the third commit.
Result of `Ripper::Lexer.new("<<A\nA\n").parse` (used in < 2.7.0) is not ordered by position.
```
[#<struct Ripper::Lexer::Elem pos=[1, 0], event=:on_heredoc_beg, tok="<<A", state=EXPR_BEG>,
 #<struct Ripper::Lexer::Elem pos=[2, 0], event=:on_heredoc_end, tok="A\n", state=EXPR_BEG>,
 #<struct Ripper::Lexer::Elem pos=[1, 3], event=:on_nl, tok="\n", state=EXPR_BEG>]
```

PR #391, opened 2022-08-10
Fix colorize_code, fix irb crash.

tompng:
Fixed `IRB::Color.colorize_code` to fix irb crash.

irb crashes when I pasted the code below.
```ruby
__END__
foobar
```

```
/Users/tomoya/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline/line_editor.rb:1177:in `+': no implicit conversion of nil into String (TypeError)
	from /Users/tomoya/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline/line_editor.rb:1177:in `render_partial'
	from /Users/tomoya/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline/line_editor.rb:508:in `rerender'
```

`IRB::Color.colorize_code` did not colorized the DATA part, and `Reline.output_modifier_proc` returned too few lines, caused `string + nil` error.
IRB::Color.colorize_code must colorize all lines, and IMO, all characters.

Ripper.lex sometimes skips some characters.
```ruby
# "world" is skipped.
Ripper.lex("hello\n__END__\nworld").map{_3}.join #=> "hello\n__END__\n"
# "\0\0foo\n" and "bar" are skipped.
Ripper.lex("<<A\0\0foo\nA\nbar").map{_3}.join #=> "<<AA\n"
```
Since colorize_code is used in `ruby -run -e colorize`, ability to colorize every characters in a code that includes `"\0"` `"\4"` `"\32"` might be useful.
In this pull request, I changed `IRB::Color.scan` to also scan those skipped chars and include it in the colorized string.

### DIFF
--- a/lib/irb/color.rb
+++ b/lib/irb/color.rb
@@ -128,10 +128,14 @@ module IRB # :nodoc:
 
         symbol_state = SymbolState.new
         colored = +''
-        length = 0
-        end_seen = false
 
         scan(code, allow_last_error: !complete) do |token, str, expr|
+          # handle uncolorable code
+          if token.nil?
+            colored << Reline::Unicode.escape_for_print(str)
+            next
+          end
+
           # IRB::ColorPrinter skips colorizing fragments with any invalid token
           if ignore_error && ERROR_TOKENS.include?(token)
             return Reline::Unicode.escape_for_print(code)
@@ -147,15 +151,7 @@ module IRB # :nodoc:
               colored << line
             end
           end
-          length += str.bytesize
-          end_seen = true if token == :on___end__
         end
-
-        # give up colorizing incomplete Ripper tokens
-        unless end_seen or length == code.bytesize
-          return Reline::Unicode.escape_for_print(code)
-        end
-
         colored
       end
 
@@ -170,33 +166,38 @@ module IRB # :nodoc:
       end
 
       def scan(code, allow_last_error:)
-        pos = [1, 0]
-
         verbose, $VERBOSE = $VERBOSE, nil
         RubyLex.compile_with_errors_suppressed(code) do |inner_code, line_no|
           lexer = Ripper::Lexer.new(inner_code, '(ripper)', line_no)
+          byte_pos = 0
+          line_positions = [0]
+          inner_code.lines.each do |line|
+            line_positions << line_positions.last + line.bytesize
+          end
+
+          on_scan = proc do |elem|
+            str = elem.tok
+            start_pos = line_positions[elem.pos[0] - 1] + elem.pos[1]
+            end_pos = start_pos + str.bytesize
+            next if start_pos < byte_pos
+
+            yield(nil, inner_code.byteslice(byte_pos...start_pos), nil) if byte_pos < start_pos
+            yield(elem.event, str, elem.state)
+            byte_pos = end_pos
+          end
+
           if lexer.respond_to?(:scan) # Ruby 2.7+
             lexer.scan.each do |elem|
-              str = elem.tok
               next if allow_last_error and /meets end of file|unexpected end-of-input/ =~ elem.message
-              next if ([elem.pos[0], elem.pos[1] + str.bytesize] <=> pos) <= 0
-
-              str.each_line do |line|
-                if line.end_with?("\n")
-                  pos[0] += 1
-                  pos[1] = 0
-                else
-                  pos[1] += line.bytesize
-                end
-              end
-
-              yield(elem.event, str, elem.state)
+              on_scan.call(elem)
             end
           else
-            lexer.parse.each do |elem|
-              yield(elem.event, elem.tok, elem.state)
+            lexer.parse.sort_by(&:pos).each do |elem|
+              on_scan.call(elem)
             end
           end
+          # yield uncolorable DATA section
+          yield(nil, inner_code.byteslice(byte_pos...inner_code.bytesize), nil) if byte_pos < inner_code.bytesize
         end
       ensure
         $VERBOSE = verbose
--- a/lib/irb/color.rb
+++ b/lib/irb/color.rb
@@ -176,14 +176,18 @@ module IRB # :nodoc:
           end
 
           on_scan = proc do |elem|
-            str = elem.tok
             start_pos = line_positions[elem.pos[0] - 1] + elem.pos[1]
-            end_pos = start_pos + str.bytesize
-            next if start_pos < byte_pos
 
-            yield(nil, inner_code.byteslice(byte_pos...start_pos), nil) if byte_pos < start_pos
-            yield(elem.event, str, elem.state)
-            byte_pos = end_pos
+            # yield uncolorable code
+            if byte_pos < start_pos
+              yield(nil, inner_code.byteslice(byte_pos...start_pos), nil)
+            end
+
+            if byte_pos <= start_pos
+              str = elem.tok
+              yield(elem.event, str, elem.state)
+              byte_pos = start_pos + str.bytesize
+            end
           end
 
           if lexer.respond_to?(:scan) # Ruby 2.7+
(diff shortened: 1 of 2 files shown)


lib/irb/color.rb
```diff
@@ -128,10 +128,13 @@ def colorize_code(code, complete: true, ignore_error: false, colorable: colorabl
 
         symbol_state = SymbolState.new
         colored = +''
-        length = 0
-        end_seen = false
 
         scan(code, allow_last_error: !complete) do |token, str, expr|
+          if token.nil?
```
st0012, line 133: When I commented out this entire condition, the tests still passed. 
I think it's because the `str` will still be passed to `Reline::Unicode.escape_for_print` at line `145` and then be pushed to `colored` at line `150`. So the end result is still the same.

Therefore, this condition doesn't seem to be needed?

lib/irb/color.rb
```diff
@@ -170,33 +165,37 @@ def without_circular_ref(obj, seen:, &block)
...
         verbose, $VERBOSE = $VERBOSE, nil
         RubyLex.compile_with_errors_suppressed(code) do |inner_code, line_no|
           lexer = Ripper::Lexer.new(inner_code, '(ripper)', line_no)
+          byte_pos = 0
+          line_positions = [0]
+          inner_code.lines.each do |line|
+            line_positions << line_positions.last + line.bytesize
+          end
+
+          on_scan = proc do |elem|
```
st0012, line 177: Do you think it's more understandable if we rewrite the block like this?

```rb
  on_scan = proc do |elem|
    start_pos = line_positions[elem.pos[0] - 1] + elem.pos[1]

    # yield uncolorable code
    if byte_pos < start_pos
      yield(nil, inner_code.byteslice(byte_pos...start_pos), nil)
    end

    if byte_pos <= start_pos
      str = elem.tok
      yield(elem.event, str, elem.state)
      byte_pos = start_pos + str.bytesize
    end
  end
```

Reasons for the changes:

1. It's easier to understand if we always put `byte_pos` and `start_pos` at the same position in comparisons.
2. We can move local variable assignment closer to where it's going to be used.
3. In complicated logic like this, I feel traditional if block is easier to understand.

lib/irb/color.rb
```diff
@@ -170,33 +165,37 @@ def without_circular_ref(obj, seen:, &block)
...
-                end
-              end
-
-              yield(elem.event, str, elem.state)
+              on_scan.call(elem)
             end
           else
-            lexer.parse.each do |elem|
-              yield(elem.event, elem.tok, elem.state)
+            lexer.parse.sort_by(&:pos).each do |elem|
```
st0012, line 194: Tests still pass without this change. Why is it needed?

lib/irb/color.rb
```diff
@@ -170,33 +165,37 @@ def without_circular_ref(obj, seen:, &block)
...
+              on_scan.call(elem)
             end
           end
+          yield(nil, inner_code.byteslice(byte_pos...inner_code.bytesize), nil) if byte_pos < inner_code.bytesize
```
st0012, line 198: Can we add a comment to explain what cases this handles?

tompng: I added comment `yield uncolorable DATA section`

lib/irb/color.rb
tompng: The code you suggested looks more readable. thanks

lib/irb/color.rb
tompng: I added test case `<<A+1\nA` for this.

Another code `lexer.parse.reject` in ruby-lex.rb should also be fixed to `lexer.parse.sort_by(&:pos).reject`. This change is in another pull request #390

lib/irb/color.rb
tompng: It actually works without this change, (I did not notice that) but I think this code is necessary.

I think `symbol_state.scan_token(token)` and `dispatch_seq` is originally designed to receive `Ripper::Lexer::Elem#event` that is not nil.
The current implementation of scan_token and dispatch_seq works just by chance with nil, but I don't think we should pass nil to it.

st0012: I see, thanks for the clear explanation 👍

lib/irb/color.rb
st0012: I removed this `sort_by` again and the test still passed 🤔

tompng: This `else` branch is only executed when ruby < 2.7 and test failed in my environment `ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-darwin19]` without `sort_by`.

```
[tomoya:irb]% ruby -v
ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-darwin19]
[tomoya:irb]% git log -n1
commit da54e7f0813d88546091af5bf220d83a6bd10198 (HEAD -> colorize_every_characters, tompng/colorize_every_characters)
Author: tompng <tomoyapenguin@gmail.com>
Date:   Mon Sep 19 14:14:10 2022 +0900

    Rewrite on_scan proc to be more readable.
[tomoya:irb]% git diff
[tomoya:irb]% rake test
...skip...
..........................
Finished in 6.179684 seconds.
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
114 tests, 1158 assertions, 0 failures, 0 errors, 6 pendings, 0 omissions, 0 notifications
94.7368% passed
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
18.45 tests/s, 187.39 assertions/s
[tomoya:irb]% # delete sort_by(&:pos)
[tomoya:irb]% git diff
diff --git a/lib/irb/color.rb b/lib/irb/color.rb
index 7071696..abe0590 100644
--- a/lib/irb/color.rb
+++ b/lib/irb/color.rb
@@ -196,7 +196,7 @@ module IRB # :nodoc:
               on_scan.call(elem)
             end
           else
-            lexer.parse.sort_by(&:pos).each do |elem|
+            lexer.parse.each do |elem|
               on_scan.call(elem)
             end
           end
[tomoya:irb]% rake test
...skip...
..................F
=========================================================================================================================================================================
Failure: test_colorize_code(TestIRB::TestColor)
/Users/tomoya/Documents/github/ruby/irb/test/irb/test_color.rb:269:in `assert_equal_with_term'
/Users/tomoya/Documents/github/ruby/irb/test/irb/test_color.rb:139:in `block in test_colorize_code'
     136: 
     137:       tests.each do |code, result|
     138:         if colorize_code_supported?
  => 139:           assert_equal_with_term(result, code, complete: true)
     140:           assert_equal_with_term(result, code, complete: false)
     141: 
     142:           assert_equal_with_term(code, code, complete: true, tty: false)
/Users/tomoya/Documents/github/ruby/irb/test/irb/test_color.rb:137:in `each'
/Users/tomoya/Documents/github/ruby/irb/test/irb/test_color.rb:137:in `test_colorize_code'
Case: colorize_code("<<A+1\nA", complete: true)
Result: "#{RED}<<A#{CLEAR}+1\n#{RED}A#{CLEAR}"
<"\e[31m<<A\e[0m+\e[34m\e[1m1\e[0m\n" + "\e[31mA\e[0m"> expected but was
<"\e[31m<<A\e[0m+1\n" + "\e[31mA\e[0m">

diff:
? <<A+1m1
  A
=========================================================================================================================================================================
...skip...
..........................
Finished in 4.4035 seconds.
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
114 tests, 1103 assertions, 1 failures, 0 errors, 6 pendings, 0 omissions, 0 notifications
93.8596% passed
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
25.89 tests/s, 250.48 assertions/s
rake aborted!
Command failed with status (1)
/Users/tomoya/.rbenv/versions/2.6.0/bin/bundle:23:in `load'
/Users/tomoya/.rbenv/versions/2.6.0/bin/bundle:23:in `<main>'
Tasks: TOP => test
(See full trace by running task with --trace)
```

st0012: Ah sorry I missed that 🤦‍♂️

st0012:
@tompng thanks for your fix, I think this PR looks good now 🙌 
@peterzhu2118 can you help me merge it when you have time? Thx

peterzhu2118 (approved):
Looks good. Thank you for your contribution!

PR #393, opened 2022-08-23
Drop hard dependency on RDoc.

voxik:
This has been introduced in 026700499dfd640b2072d7bf0370247a98d5ac40, but it seems that this is just be mistake, otherwise the later handling of `LoadError` would not be needed.

Of course I might misunderstand the code. But if that is the case, I still think that RDoc should not be loaded by default during startup, but later when really needed.

PR #395, opened 2022-08-25
Require RDoc in `input-method.rb` again in a limited scope.

junaruga:
This PR is a follow up by https://github.com/ruby/irb/pull/393#issuecomment-1227362145.

RDoc is implemented as soft dependency in IRB. See how the rdoc is required in the files. I reverted the commit below.

```
$ grep -ril rdoc lib/
lib/irb/cmd/help.rb
lib/irb/completion.rb
lib/irb/easter-egg.rb
lib/irb/input-method.rb
```

Ideally I wanted to run test the case of 'no rdoc'  with the logic in `test/lib/helper.rb` below. But I couldn't do it. Perhaps the `Kernel#require` was overridden by Rubygems or Bundler in the process of running unit-test.

```
if ENV["NO_RDOC"] == "true"
  module Kernel
    alias_method :old_require, :require
    def require(name)
      raise LoadError, "cannot load such file -- rdoc (test)" if name == "rdoc"
      old_require(name)
    end
  end
end

require 'irb'
```

---

Revert "Remove `require` in signal handler to avoid ThreadError"

This reverts commit 5f749c613c895cf1b11b5e4cbd1205363bc58028.

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -320,6 +320,11 @@ module IRB
         [195, 164], # The "ä" that appears when Alt+d is pressed on xterm.
         [226, 136, 130] # The "∂" that appears when Alt+d in pressed on iTerm2.
       ]
+      begin
+        require 'rdoc'
+      rescue LoadError
+        return nil
+      end
 
       if just_cursor_moving and completion_journey_data.nil?
         return nil


hsbt:
👍 Nice catch, Thanks.

PR #397, opened 2022-08-31
Always use local variables in current context to parse code

tompng:
Ruby will parse differently depending on defined local_variables.
```ruby
a = 1
a +1 # a + 1
b +1 # b(+1)
a *a # a * a
b *b # b(*b)
a **a # a ** a
b **b # b(**b)
a %(1) # a % 1
b %(1) # b("1")
a /1/i # a / 1 / i
b /1/i # b(regexp)
```
Therefore coloring, indent, `code_block_open` and `assignment_expression?` needs local_variables in current context.

### assignment expression check

The code below is an assignment expression and the evaluated result (large array) should be truncated.
```ruby
a /"/i if false; a=1; x=1000.times.to_a#".size
```
But once it is evaluated, it turns into a division expression(`a / "string".size`).
Assignment expression check should be done before evaluation.

### semicolon and newline character
There are `;\n` in the code to be parsed. `"a=b=nil;\n#{original_code}"`
Without `\n`, syntax valid code `=begin\n=end` will be syntax error.
Without `;`, `.to_s` will wrongly be syntax ok.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -513,6 +513,7 @@ module IRB
             if IRB.conf[:MEASURE] && IRB.conf[:MEASURE_CALLBACKS].empty?
               IRB.set_measure_callback
             end
+            is_assignment = assignment_expression?(line)
             if IRB.conf[:MEASURE] && !IRB.conf[:MEASURE_CALLBACKS].empty?
               result = nil
               last_proc = proc{ result = @context.evaluate(line, line_no, exception: exc) }
@@ -529,7 +530,7 @@ module IRB
               @context.evaluate(line, line_no, exception: exc)
             end
             if @context.echo?
-              if assignment_expression?(line, context: @context)
+              if is_assignment
                 if @context.echo_on_assignment?
                   output_value(@context.echo_on_assignment? == :truncate)
                 end
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -828,7 +828,7 @@ module IRB
       # array of parsed expressions. The first element of each expression is the
       # expression's type.
       verbose, $VERBOSE = $VERBOSE, nil
-      code = "#{RubyLex.local_variables_assign_code(context: @context) || 'nil;'}\n#{line}"
+      code = "#{RubyLex.generate_local_variables_assign_code(context: @context) || 'nil;'}\n#{line}"
       node_type = Ripper.sexp(code)&.dig(1)&.drop(1)&.dig(-1, 0)
       result = ASSIGNMENT_NODE_TYPES.include?(node_type)
       $VERBOSE = verbose
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -529,7 +529,7 @@ module IRB
               @context.evaluate(line, line_no, exception: exc)
             end
             if @context.echo?
-              if assignment_expression?(line)
+              if assignment_expression?(line, context: @context)
                 if @context.echo_on_assignment?
                   output_value(@context.echo_on_assignment? == :truncate)
                 end
@@ -827,7 +827,9 @@ module IRB
       # array of parsed expressions. The first element of each expression is the
       # expression's type.
       verbose, $VERBOSE = $VERBOSE, nil
-      result = ASSIGNMENT_NODE_TYPES.include?(Ripper.sexp(line)&.dig(1,-1,0))
+      code = "#{RubyLex.local_variables_assign_code(context: @context) || 'nil;'}\n#{line}"
+      node_type = Ripper.sexp(code)&.dig(1)&.drop(1)&.dig(-1, 0)
+      result = ASSIGNMENT_NODE_TYPES.include?(node_type)
       $VERBOSE = verbose
       result
     end
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -513,6 +513,7 @@ module IRB
             if IRB.conf[:MEASURE] && IRB.conf[:MEASURE_CALLBACKS].empty?
               IRB.set_measure_callback
             end
+            # Assignment expression check should be done before @context.evaluate to handle code like `a /2#/ if false; a = 1`
             is_assignment = assignment_expression?(line)
             if IRB.conf[:MEASURE] && !IRB.conf[:MEASURE_CALLBACKS].empty?
               result = nil
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -506,7 +506,7 @@ module IRB
 
       @scanner.set_auto_indent(@context) if @context.auto_indent_mode
 
-      @scanner.each_top_level_statement do |line, line_no|
+      @scanner.each_top_level_statement(@context) do |line, line_no|
         signal_status(:IN_EVAL) do
           begin
             line.untaint if RUBY_VERSION < '2.7'
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -828,11 +828,12 @@ module IRB
       # array of parsed expressions. The first element of each expression is the
       # expression's type.
       verbose, $VERBOSE = $VERBOSE, nil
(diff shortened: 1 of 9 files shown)


lib/irb/ruby-lex.rb
```diff
@@ -269,18 +275,15 @@ def each_top_level_statement
     end
   end
 
-  def lex
+  def lex(context: nil)
```
st0012, line 278: Is this and the below context passing necessary? It looks like `lex` is never called with `context` argument?

lib/irb/ruby-lex.rb
```diff
@@ -269,18 +275,15 @@ def each_top_level_statement
...
     end
     code = @line + (line.nil? ? '' : line)
     code.gsub!(/\s*\z/, '').concat("\n")
-    @tokens = self.class.ripper_lex_without_warning(code)
-    @continue = process_continue
-    @code_block_open = check_code_block(code)
-    @indent = process_nesting_level
-    @ltype = process_literal_type
+    @tokens = self.class.ripper_lex_without_warning(code, context: context)
+    @ltype, @indent, @continue, @code_block_open = check_state(code, @tokens, context: context)
```
st0012, line 286: Good catch on using `check_state` instead 👍

lib/irb/ruby-lex.rb
```diff
@@ -136,16 +136,20 @@ def set_prompt(p = nil, &block)
     :on_param_error
   ]
 
+  def self.local_variables_assign_code(context: nil, local_variables: [])
```
st0012, line 139: Can we call it `generate_local_variables_assign_code`?

test/irb/test_color.rb
```diff
@@ -155,6 +155,18 @@ def test_colorize_code
       end
     end
 
+    def test_colorize_code_with_local_variables
+      code = "a /(b +1)/i"
+      result_without_lvars = "a #{RED}#{BOLD}/#{CLEAR}#{RED}(b +1)#{CLEAR}#{RED}#{BOLD}/i#{CLEAR}"
+      result_with_lvar = "a /(b #{BLUE}#{BOLD}+1#{CLEAR})/i"
+      result_with_lvars = "a /(b +#{BLUE}#{BOLD}1#{CLEAR})/i"
+      if colorize_code_supported?
```
st0012, line 163: This is not necessary because irb requires Ruby 2.5+, which always satisfy this condition. I added https://github.com/ruby/irb/pull/413 to remove this obsolete check.

lib/irb/ruby-lex.rb
tompng: This context is mainly used to test RubyLex's nesting_level and code_block_open calculation in test_ruby_lex.rb

I fixed to call lex with context in L248.

I also changed the optional keyword parameter `context:` to normal required parameter.
- Not to forget calling method without context
- Method in this library that receives only context is using normal parameter, not keyword parameter. (ex: `set_auto_indent(context)`, `suspend_context(context)`)

tompng: > Is it necessary?

Yes.
Since RubyLex#lex just returns @input.call in mulitline mode, this code is needed only when USE_MULTILINE is false.
Example when nil is passed to context here:
```
% irb --nomultiline
irb(main):001:0> a = 8
=> 8
irb(main):002:0> a /2 #=> should be 4
irb(main):003:0/ 
irb(main):004:0/ 
```

lib/irb.rb
```diff
@@ -506,13 +506,14 @@ def eval_input
...
 
-      @scanner.each_top_level_statement do |line, line_no|
+      @scanner.each_top_level_statement(@context) do |line, line_no|
         signal_status(:IN_EVAL) do
           begin
             line.untaint if RUBY_VERSION < '2.7'
             if IRB.conf[:MEASURE] && IRB.conf[:MEASURE_CALLBACKS].empty?
               IRB.set_measure_callback
             end
+            is_assignment = assignment_expression?(line)
```
st0012, line 516: Why moving this line here though? It doesn't look like to change the behavior?

lib/irb.rb
```diff
@@ -827,7 +828,9 @@ def assignment_expression?(line)
       # array of parsed expressions. The first element of each expression is the
       # expression's type.
       verbose, $VERBOSE = $VERBOSE, nil
-      result = ASSIGNMENT_NODE_TYPES.include?(Ripper.sexp(line)&.dig(1,-1,0))
+      code = "#{RubyLex.generate_local_variables_assign_code(context: @context) || 'nil;'}\n#{line}"
+      node_type = Ripper.sexp(code)&.dig(1)&.drop(1)&.dig(-1, 0)
```
st0012, line 832: This is not super easy to understand. Would you mind adding a comment for it?

lib/irb.rb
```diff
@@ -827,7 +828,9 @@ def assignment_expression?(line)
...
+      result = ASSIGNMENT_NODE_TYPES.include?(node_type)
       $VERBOSE = verbose
```
st0012, line 834: I feel this should be in an `ensure` block.

lib/irb/ruby-lex.rb
```diff
@@ -136,16 +136,20 @@ def set_prompt(p = nil, &block)
     :on_param_error
   ]
 
+  def self.generate_local_variables_assign_code(context: nil, local_variables: [])
```
st0012, line 139: I feel it weird to let `generate_local_variables_assign_code` take both `context` and `local_variables` while:

1. They're never used at the same time.
2. Local variables are all retrieved from the same context, just at different timings.

Can we make it `self.generate_local_variables_assign_code(local_variables)` instead?

And we can add something like `Context#local_variables` to encapsulate `&.workspace&.binding&.local_variables`

So that:

- We can have less methods taking `context` while what they actually need are just the local variables
- But it's also easier to generate local variables from a context
- `generate_local_variables_assign_code` can have a simpler implementation

lib/irb.rb
tompng: It actually changes the behavior in a rare case.
```ruby
irb = IRB::Irb.new
context = irb.instance_variable_get('@context')

code = "a /'/i if false; a=1; x=1000.times.to_a#'.size"
irb.assignment_expression?(code) #=> true
context.evaluate(code, 1) #=> [0,1,2,3,...] assignment to local variable x
irb.assignment_expression?(code) #=> false
context.evaluate(code, 1) #=> 0 not an assignment expression
```

`assignment_expression?` check should be done before `@context.evaluate` (in L519 and L530)

I added test case for this

lib/irb/ruby-lex.rb
tompng: > `self.generate_local_variables_assign_code(local_variables)` `Context#local_variables`

It looks good, I fixed it

lib/irb/input-method.rb
```diff
@@ -286,7 +286,8 @@ def initialize
         if IRB.conf[:USE_COLORIZE]
           proc do |output, complete: |
             next unless IRB::Color.colorable?
-            IRB::Color.colorize_code(output, complete: complete)
+            lvars = IRB.CurrentContext&.workspace&.binding&.local_variables || []
```
st0012, line 289:
```suggestion
            lvars = IRB.CurrentContext&.local_variables || []
```

lib/irb.rb
st0012: Wow thanks for spotting this. I think it also worth a comment then?

st0012: Thank you for the example 👍

st0012:
@tompng I think the `head` tests will pass once you rebased `master`.

st0012:
@k0kubun would you mind giving it a look? thx

k0kubun (approved):
Makes sense. Thank you!

tompng:
🎉 
Thanks for reviewing and merging 🙏

PR #402, opened 2022-09-27
Fix: cannot paste code correctly when it contains heredoc 

tompng:
### Problem
When I paste the code below,
```
123
<<EOS
EOS
```
I got
```
irb(main):001:0> 123
=> 123
irb(main):002:0" <<EOSEOS
irb(main):003:0" EOS
irb(main):004:0" 
```

RubyLex.ripper_lex_without_warning tries to ignore error tokens if possible, but the current implementation fails with `<<EOS`
```ruby
RubyLex.ripper_lex_without_warning('<<EOS').map(&:tok)
#=> ["<<EOS", "EOS"] # should be ["<<EOS"]
```

### Changes
Result of `lexer.scan` are sorted by pos, so we only need to look at prev token.
If there is overlap, error token should be ignored.
In the original code, it does not check overlap, but checks if start positions are the same

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -152,17 +152,13 @@ class RubyLex
       lexer = Ripper::Lexer.new(inner_code, '-', line_no)
       if lexer.respond_to?(:scan) # Ruby 2.7+
         tokens = []
-        pos_to_index = {}
         lexer.scan.each do |t|
           next if t.pos.first == 0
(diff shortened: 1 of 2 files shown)


st0012 (commented):
@tompng Thank you for the fix!

@peterzhu2118 I think this looks good and have tested it locally. Can you help me merge it?

I have some refactor ideas for the method but they're not related to this fix, so I'll open another PR once this is merged.

PR #409, opened 2022-10-04
Rename IDB::ReidlineInputMethod to IRB::RelineInputMethod

gmcgibbon:
Deprecates IDB::ReidlineInputMethod and USE_REIDLINE in favor of IRB::RelineInputMethod and USE_RELINE. The Input method uses Reline to read input from the console, so it can be named directly after the Reline library like other inputs methods are (Readline, Stdio, etc.).

### DIFF
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -22,7 +22,7 @@ module IRB
     #
     # The optional +input_method+ argument:
     #
-    # +nil+::     uses stdin or Reidline or Readline
+    # +nil+::     uses stdin or Reline or Readline
     # +String+::  uses a File
     # +other+::   uses this as InputMethod
     def initialize(irb, workspace = nil, input_method = nil)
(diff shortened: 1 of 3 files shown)


hsbt:
Nice catch, Thanks!

lib/irb/context.rb
```diff
@@ -48,8 +48,13 @@ def initialize(irb, workspace = nil, input_method = nil)
...
         @use_multiline = IRB.conf[:USE_MULTILINE]
-      elsif IRB.conf.has_key?(:USE_REIDLINE) # backward compatibility
-        @use_multiline = IRB.conf[:USE_REIDLINE]
+      elsif IRB.conf.has_key?(:USE_RELINE) # backward compatibility
+        @use_multiline = IRB.conf[:USE_RELINE]
+      elsif IRB.conf.has_key?(:USE_REIDLINE)
+        warn <<~MSG.strip
+          USE_REIDLINE is deprecated, please use USE_RELINE instead.
+        MSG
+        @use_multiline = IRB.conf[:USE_RELINE]
```
st0012, line 57: If this is for backward compatibility, shouldn't this be `USE_REIDLINE`?

st0012:
@gmcgibbon Thanks for the PR. But even though most users don't use it directly, gems may reference the `ReidlineInputMethod` directly ([example](https://github.com/ankane/archer/blob/master/lib/archer.rb#L64)). This change also means apps that use `Sorbet` will fail type checking after upgrading irb because the names will mismatch in their rbi files. 

Given the above reasons, I actually planned to put it on hold and split the changes in different levels of releases.

@hsbt Can we:
1. Revert this change 
2. Add the deprecation message for `USE_REIDLINE` config option
3. Cut `v1.4.3` for the above changes

I can make PRs for them if you think the approach is reasonable.

hsbt:
I understood your suggestion. But I won't care about sorbet use-case because I'm not sorbet user.

lib/irb/context.rb
hsbt: maked sense.

lib/irb/input-method.rb
```diff
@@ -470,4 +470,13 @@ def inspect
       str
     end
   end
+
+  class ReidlineInputMethod < RelineInputMethod
```
st0012, line 474: Sorry that I missed this definition. I think it's a safe change and added #411  to pick up a few leftovers.

PR #411, opened 2022-10-05
Rename leftover reidline references

st0012:
I cleaned up a few remaining `reidline` references.

Also, based on https://github.com/ruby/irb/commit/93f87ec65344b44123b0150a5fc07de1fd4c80c4, it appears that the maintainer @aycabta wanted to deprecate any `USE_*LINE` configs in favor of `USE_MULTILINE`. So we should deprecate `USE_RELINE` as well.


st0012:
cc @hsbt

hsbt (approved):

PR #379, opened 2022-07-18
Fixed sorting of variables and matched suggestions in completion

makketagg:
Fixed sorting of suggestions. Fixed #378

PR #412, opened 2022-10-06
Increase completion's test coverage and add some minor improvements

st0012:
## Changes

1. Add tests to cover primitive types' completion
2. Group some completion tests together
3. Evaluate completion candidates only when we really need them (usually when `doc_namespace == false`)
4. Correct `assert_equal`'s argument order in completion tests (I'll go through other tests in follow up PRs)
5. Sort constant completion's candidates like #379 does

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -255,7 +255,7 @@ module IRB
             candidates = []
           end
 
-          select_message(receiver, message, candidates, "::")
+          select_message(receiver, message, candidates.sort, "::")
         end
 
       when /^(:[^:.]+)(\.|::)([^.]*)$/
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -171,10 +171,10 @@ module IRB
         receiver = $1
         message = $3
 
-        candidates = String.instance_methods.collect{|m| m.to_s}
         if doc_namespace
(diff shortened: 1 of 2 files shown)


st0012:
cc @peterzhu2118

lib/irb/completion.rb
```diff
@@ -207,29 +207,33 @@ def self.retrieve_completion_data(input, bind: IRB.conf[:MAIN_CONTEXT].workspace
...
           ["Proc.#{message}", "Hash.#{message}"]
         else
+          proc_candidates = Proc.instance_methods.collect{|m| m.to_s}
+          hash_candidates = Hash.instance_methods.collect{|m| m.to_s}
           select_message(receiver, message, proc_candidates | hash_candidates)
         end
 
       when /^(:[^:.]*)$/
         # Symbol
-        return nil if doc_namespace
```
peterzhu2118, line 220: I prefer this guard style when there's a lot of code that follows it rather than using an `else` clause. I find it to be easier to understand.

st0012: I prefer letting it match the rest of the `when` condition. I think having a similar shape of code will help skimming through this lengthy statement.

st0012:
@peterzhu2118 thank you ❤️

PR #413, opened 2022-10-07
Remove unnecessary coloring support check

st0012:
Since IRB [requires Ruby 2.5+](https://github.com/ruby/irb/blob/master/irb.gemspec#L37), we don't need to check if coloring is supported in tests.


st0012:
cc @peterzhu2118

peterzhu2118 (approved):
Looks good, thank you!

Issue #396, opened 2022-08-31
Adding test cases on no rdoc gem case.


junaruga:
## Description

> What are your expected behavior and actual behavior of your environment?

On the latest master b24852058fc87c940252c8a711c60ae2eb298082, the rdoc is required as soft dependency (= load the `rdoc` when it is needed).

```
$ grep -ril rdoc lib/
lib/irb/cmd/help.rb
lib/irb/completion.rb
lib/irb/easter-egg.rb
lib/irb/input-method.rb
```

And this is useful. Because on Fedora project, there is `rubygem-irb` RPM package, and users can install the RPM without the rdoc gem. So, I want to see the test cases of no rdoc to prevent this kind of issue, https://github.com/ruby/irb/issues/342.

I think we can add the test cases by mocking `Kernel#require` with returning `LoadError` when the argument is `rdoc`. (ref: https://github.com/ruby/irb/pull/395#issue-1351284638)

When chatting on the Ruby Slack, I was told that the [test-unit-rr](https://rubygems.org/gems/test-unit-rr) gem could be used to implement the mock. 

And while the ruby/irb's test is also used in ruby/ruby, and ruby/ruby doesn't want to depend on the test-unit-rr, we can implement the new tests using a flog to check if the current running environment is ruby/irb or ruby/ruby by a flag like `ruby_core?` used in Bundler's test,  https://github.com/ruby/ruby/blob/d6f21b308bcff03e82f8b3dbf11a852ce111b3b3/spec/bundler/support/path.rb#L250-L259 .

## Result of irb_info

> Please paste the result of `irb_info` command in IRB.

As this ticket is about enhancing the tests, there is no information about this item.

## Terminal Emulator

> What's your terminal emulator?

As this ticket is about enhancing the tests, there is no information about this item.

## Setting Files

> Are you using `~/.irbrc` and `~/.inputrc`?

As this ticket is about enhancing the tests, there is no information about this item.

st0012:
FYI, I have added tests for all `rdoc` requires except the one in `easter-egg.rb` with:

- https://github.com/ruby/irb/pull/421
- https://github.com/ruby/irb/pull/418
- https://github.com/ruby/irb/pull/416

junaruga:
Thanks! Great work!

hsbt:
@junaruga Can we close this?

junaruga:
Thanks, that's excellent work!
[...]

PR #416, opened 2022-10-18
Move rdoc require out of repeated execution path

st0012:
`SHOW_DOC_DIALOG` will be called repeatedly whenever the corresponding key is pressed, but we only need to require `rdoc` once. So ideally the require can be put outside of the proc.

And because when `rdoc` is not available the entire proc will be nonfunctional, we can stop registering the SHOW_DOC_DIALOG if we failed to require `rdoc`.

### DIFF
--- a/test/irb/test_input_method.rb
+++ b/test/irb/test_input_method.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: false
+
+require "test/unit"
+require "irb"
+
+module TestIRB
+  class TestRelineInputMethod < Test::Unit::TestCase
+    def setup
+      @conf_backup = IRB.conf.dup
+      IRB.conf[:LC_MESSAGES] = IRB::Locale.new
+    end
+
+    def teardown
+      IRB.conf.replace(@conf_backup)
+    end
+
+    def test_initialization
+      IRB::RelineInputMethod.new
+
+      assert_nil Reline.completion_append_character
+      assert_equal '', Reline.completer_quote_characters
+      assert_equal IRB::InputCompletor::BASIC_WORD_BREAK_CHARACTERS, Reline.basic_word_break_characters
+      assert_equal IRB::InputCompletor::CompletionProc, Reline.completion_proc
+      assert_equal IRB::InputCompletor::PerfectMatchedProc, Reline.dig_perfect_match_proc
+    end
+
+    def test_initialization_without_use_autocomplete
+      original_show_doc_proc = Reline.dialog_proc(:show_doc)&.dialog_proc
+      empty_proc = Proc.new {}
+      Reline.add_dialog_proc(:show_doc, empty_proc)
+
+      IRB.conf[:USE_AUTOCOMPLETE] = false
+
+      IRB::RelineInputMethod.new
+
+      refute Reline.autocompletion
+      assert_equal empty_proc, Reline.dialog_proc(:show_doc).dialog_proc
+    ensure
+      Reline.add_dialog_proc(:show_doc, original_show_doc_proc, Reline::DEFAULT_DIALOG_CONTEXT)
+    end
+
+    def test_initialization_with_use_autocomplete
+      original_show_doc_proc = Reline.dialog_proc(:show_doc)&.dialog_proc
+      empty_proc = Proc.new {}
+      Reline.add_dialog_proc(:show_doc, empty_proc)
+
+      IRB.conf[:USE_AUTOCOMPLETE] = true
+
+      IRB::RelineInputMethod.new
+
+      assert Reline.autocompletion
+      assert_equal IRB::RelineInputMethod::SHOW_DOC_DIALOG, Reline.dialog_proc(:show_doc).dialog_proc
+    ensure
+      Reline.add_dialog_proc(:show_doc, original_show_doc_proc, Reline::DEFAULT_DIALOG_CONTEXT)
+    end
+  end
+end
+
--- a/test/irb/test_input_method.rb
+++ b/test/irb/test_input_method.rb
@@ -53,6 +53,37 @@ module TestIRB
     ensure
       Reline.add_dialog_proc(:show_doc, original_show_doc_proc, Reline::DEFAULT_DIALOG_CONTEXT)
     end
+
+    def test_initialization_with_use_autocomplete_but_without_rdoc
+      original_show_doc_proc = Reline.dialog_proc(:show_doc)&.dialog_proc
+      empty_proc = Proc.new {}
+      Reline.add_dialog_proc(:show_doc, empty_proc)
+
+      IRB.conf[:USE_AUTOCOMPLETE] = true
+
+      without_rdoc do
+        IRB::RelineInputMethod.new
+      end
+
+      assert Reline.autocompletion
+      # doesn't register show_doc dialog
+      assert_equal empty_proc, Reline.dialog_proc(:show_doc).dialog_proc
+    ensure
+      Reline.add_dialog_proc(:show_doc, original_show_doc_proc, Reline::DEFAULT_DIALOG_CONTEXT)
+    end
+
+    def without_rdoc(&block)
+      ::Kernel.send(:alias_method, :old_require, :require)
+
+      ::Kernel.define_method(:require) do |name|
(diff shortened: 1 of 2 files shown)


junaruga:
Now I checked the PR's code.

This PR's intent is to improve the performance, avoid calling the `require 'rdoc'` repeatedly, because accessing the rdoc files to check if it exists repeatedly is costy. Right?

I suppose if we keep the compatibility from the previous code,  when the `require 'rdoc'` raises the `LoadError`, we still need to set the `nil` to the `Reline.add_dialog_proc`'s 2nd argument. So, the code can be below.

```
      if IRB.conf[:USE_AUTOCOMPLETE]
        dialog = nil 
        begin
          require 'rdoc'
          dialog = SHOW_DOC_DIALOG
        rescue LoadError
        end
        Reline.add_dialog_proc(:show_doc, dialog, Reline::DEFAULT_DIALOG_CONTEXT)
      end 
```

But if we don't need to call the `Reline.add_dialog_proc` in the case of the `require 'rdoc'` raising the `LoadError`, the current PR looks ok.

st0012:
> This PR's intent is to improve the performance, avoid calling the require 'rdoc' repeatedly, because accessing the rdoc files to check if it exists repeatedly is costy. Right?

Yes. Especially in the case when `rdoc` is not installed, it'll repeatedly raise and rescue `LoadError`.

> I suppose if we keep the compatibility from the previous code, when the require 'rdoc' raises the LoadError, we still need to set the nil to the Reline.add_dialog_proc's 2nd argument.

Actually, the `dialog` can't be `nil` because when being rendered it'll cause error [at this line](https://github.com/ruby/reline/blob/d00896a571ff953a1ac1cc044a57c66f35c23959/lib/reline/line_editor.rb#L590):

```
/Users/st0012/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/reline-0.3.1/lib/reline/line_editor.rb:588:in `instance_exec': no block given (LocalJumpError)
```

So my test cases were actually misleading and I've updated them.

If we still want to have the `:show_proc` dialog registered for compatibility reason, we can use `Proc.new {}` instead.

> But if we don't need to call the Reline.add_dialog_proc in the case of the require 'rdoc' raising the LoadError

I think we don't need to call `Reline.add_dialog_proc` in that case tbh. Because `reline` [iterates through registered dialogs](https://github.com/ruby/reline/blob/d00896a571ff953a1ac1cc044a57c66f35c23959/lib/reline/line_editor.rb#L651) in every rendering and [performs certain computation](https://github.com/ruby/reline/blob/d00896a571ff953a1ac1cc044a57c66f35c23959/lib/reline/line_editor.rb#L671-L685) even when if the dialog proc is empty, it should be avoided when possible.

junaruga:
> Yes. Especially in the case when rdoc is not installed, it'll repeatedly raise and rescue LoadError.

Ah, ok. That process looks costy.

> Actually, the dialog can't be nil because when being rendered it'll cause error [at this line](https://github.com/ruby/reline/blob/d00896a571ff953a1ac1cc044a57c66f35c23959/lib/reline/line_editor.rb#L590):

Oh I see. Thanks for finding the issue.

> If we still want to have the :show_proc dialog registered for compatibility reason, we can use Proc.new {} instead.

The `:show_proc` dialog? You meant `:show_doc` dialog?

Right now it's not easy to test the non-rdoc case. When we create the Ruby RPM package in the Fedora project, we make the separated `rdoc` RPM package as a soft (weak or optional) dependency. In the situation, I can test it the case. So, right now I am ok to merge this PR.

> I think we don't need to call Reline.add_dialog_proc in that case tbh. Because reline [iterates through registered dialogs](https://github.com/ruby/reline/blob/d00896a571ff953a1ac1cc044a57c66f35c23959/lib/reline/line_editor.rb#L651) in every rendering and [performs certain computation](https://github.com/ruby/reline/blob/d00896a571ff953a1ac1cc044a57c66f35c23959/lib/reline/line_editor.rb#L671-L685) even when if the dialog proc is empty, it should be avoided when possible.

OK. Thanks for checking it.

st0012:
> The :show_proc dialog? You meant :show_doc dialog?

Yes 🤦‍♂️ 

> Right now it's not easy to test the non-rdoc case. 

Yeah I'm aware of that. I'll see if having unit-tests like `TestRelineInputMethod`  makes implementing the conditional mocking you mentioned in https://github.com/ruby/irb/issues/396 easier.

junaruga:
> Yeah I'm aware of that. I'll see if having unit-tests like TestRelineInputMethod makes implementing the conditional mocking you mentioned in https://github.com/ruby/irb/issues/396 easier.

OK. That's great for you to work on covering the no-rdoc case.
Mentors for the IRB, could you review this PR?

st0012:
@junaruga I've added a case for the `rdoc` missing scenario based on your snippet in https://github.com/ruby/irb/pull/395#issue-1351284638. Let me know if that's an acceptable approach to you.

test/irb/test_input_method.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    def test_initialization_with_use_autocomplete_but_without_rdoc
+      original_show_doc_proc = Reline.dialog_proc(:show_doc)&.dialog_proc
+      empty_proc = Proc.new {}
+      Reline.add_dialog_proc(:show_doc, empty_proc)
+
+      IRB.conf[:USE_AUTOCOMPLETE] = true
+
+      without_rdoc do
+        IRB::RelineInputMethod.new
```
junaruga, line 65: The assert commands should be in the `wethout_rdoc` block shouldn't be?

```
      without_rdoc do
        IRB::RelineInputMethod.new
        assert Reline.autocompletion
        # doesn't register show_doc dialog
        assert_equal empty_proc, Reline.dialog_proc(:show_doc).dialog_proc
      end
```

test/irb/test_input_method.rb
```diff
@@ -0,0 +1,89 @@
...
+      end
+
+      assert Reline.autocompletion
+      # doesn't register show_doc dialog
+      assert_equal empty_proc, Reline.dialog_proc(:show_doc).dialog_proc
+    ensure
+      Reline.add_dialog_proc(:show_doc, original_show_doc_proc, Reline::DEFAULT_DIALOG_CONTEXT)
+    end
+
+    def without_rdoc(&block)
```
junaruga, line 75: How about moving the `without_rdoc` method to the common file `test/lib/helper.rb`, as the logic: `require 'rdoc'` and `resure LoadError` is in other parts too?

```
$ grep -r rdoc lib/
lib/irb/cmd/help.rb:        require 'rdoc/ri/driver'
lib/irb/completion.rb:        require 'rdoc'
lib/irb/easter-egg.rb:          require "rdoc"
lib/irb/input-method.rb:          require 'rdoc'
```

junaruga:
> I've added a case for the rdoc missing scenario based on your snippet in https://github.com/ruby/irb/pull/395#issue-1351284638. Let me know if that's an acceptable approach to you.

Thanks for that! That looks great. I like the approach of the `without_rdoc` as a direction. I commented. I am not sure if the test with the `without_rdoc` works on Bundler on RubyGems to pass the CI after moving the assertions to the inside of the `without_rdoc`.

test/irb/test_input_method.rb
st0012: I want to leave it to the next PR (which I'll open after this is merged) because the goal of this one is to move out of the `require "rdoc"`. Increasing the coverage is just a good addition. WDYT?

test/irb/test_input_method.rb
st0012: I think it doesn't need to be? Because `Reline`'s state isn't affected by the `without_rdoc` block. We just need to make sure the logic that changes `Reline`'s state is inside the block.

junaruga: Ah, you are right. Yes, you don't need to move the assertions.

test/irb/test_input_method.rb
junaruga: Yes, I agree with you. It can be the next PR. Without the `test_initialization_with_use_autocomplete_but_without_rdoc` test in this PR, it's still better than the current situation.

st0012:
cc @peterzhu2118

PR #418, opened 2022-10-24
Add test for the help command

st0012:
This PR adds tests for 2 of 3 possible `help` command outcomes:

- [x] When executed without `rdoc`: it returns the command constant directly
    - It probably should return some useful message, but that's for another day.
- [x] When executed with argument, like `help String#gsub`: it renders the argument's document via `rdoc`
- [ ] When executed without argument: it starts rdoc's interactive mode, which intercepts stdin and is not testable under the current test framework

Because the `help` command [redefines itself based on if rdoc is required succesfully](https://github.com/ruby/irb/blob/master/lib/irb/cmd/help.rb), there's no easy way to recover it to the pre-test state. I think the best way is to reload the command file, but any alternative is welcome.

### DIFF
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -406,6 +406,57 @@ module TestIRB
         ], out)
     end
 
+    def test_help
+      IRB.init_config(nil)
+      input = TestInputMethod.new([
+          "help 'String#gsub'\n",
+          "\n",
+        ])
+      IRB.conf[:VERBOSE] = false
+      IRB.conf[:PROMPT_MODE] = :SIMPLE
+      irb = IRB::Irb.new(IRB::WorkSpace.new(self), input)
+      IRB.conf[:MAIN_CONTEXT] = irb.context
+      out, err = capture_output do
+        irb.eval_input
+      end
+
+      # the help command lazily loads rdoc by redefining the execute method
+      assert_match(/discarding old execute/, err) unless RUBY_ENGINE == 'truffleruby'
+
+      # the former is what we'd get without document content installed, like on CI
+      # the latter is what we may get locally
+      possible_rdoc_output = [/Nothing known about String#gsub/, /Returns a copy of self with all occurrences of the given pattern/]
+      assert(possible_rdoc_output.any? { |output| output.match?(out) }, "Expect the help command to match one of the possible outputs")
+    ensure
+      # this is the only way to reset the redefined method without coupling the test with its implementation
+      load "irb/cmd/help.rb"
+    end
+
+    def test_help_without_rdoc
+      IRB.init_config(nil)
+      input = TestInputMethod.new([
+          "help 'String#gsub'\n",
+          "\n",
+        ])
+      IRB.conf[:VERBOSE] = false
+      IRB.conf[:PROMPT_MODE] = :SIMPLE
+      irb = IRB::Irb.new(IRB::WorkSpace.new(self), input)
+      IRB.conf[:MAIN_CONTEXT] = irb.context
+      out, err = capture_output do
+        without_rdoc do
+          irb.eval_input
+        end
+      end
+
+      # since LoadError will be raised, the execute won't be redefined
+      assert_no_match(/discarding old execute/, err)
+      # if it fails to require rdoc, it only returns the command object
+      assert_match(/=> IRB::ExtendCommand::Help\n/, out)
+    ensure
+      # this is the only way to reset the redefined method without coupling the test with its implementation
+      load "irb/cmd/help.rb"
+    end
+
     def test_irb_load
       IRB.init_config(nil)
       File.write("#{@tmpdir}/a.rb", "a = 'hi'\n")
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -429,7 +429,7 @@ module TestIRB
       assert(possible_rdoc_output.any? { |output| output.match?(out) }, "Expect the help command to match one of the possible outputs")
     ensure
       # this is the only way to reset the redefined method without coupling the test with its implementation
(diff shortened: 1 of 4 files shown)


test/irb/test_input_method.rb
```diff
@@ -83,19 +83,6 @@ def test_initialization_with_use_autocomplete_but_without_rdoc
     ensure
       Reline.add_dialog_proc(:show_doc, original_show_doc_proc, Reline::DEFAULT_DIALOG_CONTEXT)
     end
-
-    def without_rdoc(&block)
-      ::Kernel.send(:alias_method, :old_require, :require)
-
-      ::Kernel.define_method(:require) do |name|
-        raise LoadError, "cannot load such file -- rdoc (test)" if name == "rdoc"
-        original_require(name)
```
st0012, line 92: Here's a typo, but in these tests it's never reached 🙈

test/lib/helper.rb
```diff
@@ -9,6 +9,19 @@ class TestCase
       def windows? platform = RUBY_PLATFORM
         /mswin|mingw/ =~ platform
       end
+
+      def without_rdoc(&block)
+        ::Kernel.send(:alias_method, :old_require, :require)
+
+        ::Kernel.define_method(:require) do |name|
+          raise LoadError, "cannot load such file -- rdoc (test)" if name.match?("rdoc") || name.match?(/^rdoc\/.*/)
```
st0012, line 17: I expanded it to match `rdoc` or `rdoc/*`.

st0012:
@junaruga I think different `rdoc` invocation is likely have its own quirk when being tested, I decided to deal with them in different PRs.

st0012:
cc @peterzhu2118

junaruga:
> I think different rdoc invocation is likely have its own quirk when being tested, I decided to deal with them in different PRs.

OK. Thanks for the work!

test/lib/helper.rb
```diff
@@ -9,6 +9,19 @@ class TestCase
...
+          ::Kernel.send(:old_require, name)
+        end
+
+        yield
+      ensure
+        ::Kernel.send(:alias_method, :require, :old_require)
```
peterzhu2118, line 23: To suppress

```
/Users/peter/src/irb/test/lib/helper.rb:23: warning: method redefined; discarding old require
/Users/peter/src/irb/test/lib/helper.rb:16: warning: previous definition of require was here
```

```suggestion
        EnvUtil.suppress_warning { ::Kernel.send(:alias_method, :require, :old_require) }
```

test/irb/test_cmd.rb
```diff
@@ -406,6 +406,57 @@ def test_irb_source
...
+        end
+      end
+
+      # since LoadError will be raised, the execute won't be redefined
+      assert_no_match(/discarding old execute/, err)
+      # if it fails to require rdoc, it only returns the command object
+      assert_match(/=> IRB::ExtendCommand::Help\n/, out)
+    ensure
+      # this is the only way to reset the redefined method without coupling the test with its implementation
+      load "irb/cmd/help.rb"
```
peterzhu2118, line 457: Ditto.

```suggestion
      EnvUtil.suppress_warning { load "irb/cmd/help.rb" }
```

test/irb/test_cmd.rb
```diff
@@ -406,6 +406,57 @@ def test_irb_source
...
+      # the help command lazily loads rdoc by redefining the execute method
+      assert_match(/discarding old execute/, err) unless RUBY_ENGINE == 'truffleruby'
+
+      # the former is what we'd get without document content installed, like on CI
+      # the latter is what we may get locally
+      possible_rdoc_output = [/Nothing known about String#gsub/, /Returns a copy of self with all occurrences of the given pattern/]
+      assert(possible_rdoc_output.any? { |output| output.match?(out) }, "Expect the help command to match one of the possible outputs")
+    ensure
+      # this is the only way to reset the redefined method without coupling the test with its implementation
+      load "irb/cmd/help.rb"
```
peterzhu2118, line 432: And here.

```suggestion
      EnvUtil.suppress_warning { load "irb/cmd/help.rb" }
```

test/irb/test_cmd.rb
```diff
@@ -406,6 +406,57 @@ def test_irb_source
...
+    def test_help
+      IRB.init_config(nil)
+      input = TestInputMethod.new([
+          "help 'String#gsub'\n",
+          "\n",
+        ])
+      IRB.conf[:VERBOSE] = false
+      IRB.conf[:PROMPT_MODE] = :SIMPLE
+      irb = IRB::Irb.new(IRB::WorkSpace.new(self), input)
+      IRB.conf[:MAIN_CONTEXT] = irb.context
```
peterzhu2118, line 418: Are setting these `IRB.conf` needed? The tests seems to pass even when they're commented out.

test/irb/test_cmd.rb
```diff
@@ -406,6 +406,53 @@ def test_irb_source
         ], out)
     end
 
+    def test_help
+      IRB.init_config(nil)
+      input = TestInputMethod.new([
+          "help 'String#gsub'\n",
+          "\n",
+        ])
+      IRB.conf[:PROMPT_MODE] = :SIMPLE
```
st0012, line 415: This one is needed to reduce noise from the output.

test/irb/test_helper.rb
```diff
@@ -0,0 +1,14 @@
+module IRBTestHelper
```
st0012, line 1: I name it like this because:

- `TestHelper` is too general and could be used in other lib's tests.
- `IRB::TestHelper` adds a new constant under the `IRB` constant, which ideally shouldn't affect anything, but I want to avoid unless necessary.

peterzhu2118 (approved):
Looks good, thank you!

PR #423, opened 2022-10-27
Remove unecesary and harmful pend for TruffleRuby in TestRaiseNoBacktraceException

eregon:
* Specifically the second one causes `$HOME` to be unset, which breaks `File.expand_path('~')`.

The issue was introduced in https://github.com/ruby/irb/commit/4452adbe04cdb678f78cf94fc8673071cc4015b3, cc @aycabta

### DIFF
--- a/test/irb/test_raise_no_backtrace_exception.rb
+++ b/test/irb/test_raise_no_backtrace_exception.rb
@@ -4,7 +4,6 @@ require 'test/unit'
 module TestIRB
   class TestRaiseNoBacktraceException < Test::Unit::TestCase
     def test_raise_exception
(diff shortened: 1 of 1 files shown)


eregon:
The test [fails](https://github.com/ruby/irb/actions/runs/3340094570/jobs/5529812610) because
https://github.com/ruby/irb/blob/415e9ce69127ee3aede14fa07e47449f2a6da0d4/test/irb/test_raise_no_backtrace_exception.rb#L42
and such a locale is most likely not installed in GitHub Actions, and so I guess `nl_langinfo(CODESET)` returns ASCII or C or similar.

So maybe the test does not run as expected (on CRuby on GitHub Actions)?

PR #426, opened 2022-11-02
Allow non-identifier aliases like Pry's @ and $

k0kubun:
## Background
At DevMeeting-2022-08-25, We decided to start accepting inputs that are not valid as a Ruby expression at https://bugs.ruby-lang.org/issues/18975#note-5:

> We also discussed the idea of handling IRB command inputs that are not valid as a Ruby expression, and the meeting attendees agreed to make this change.

## Changes
* Introduce API for `~/.irbrc` to define an alias: `IRB::ExtendCommandBundle.def_alias_command`
* When a line starts with a non-identifier (not matching `/A\w+\z`) alias command, skip the Ruby-based termination check and replace the alias with the target command.

### Example
If you define an alias like:

```rb
IRB.conf[:COMMAND_ALIASES] = { :"@" => :whereami }
```

You can use it as a shorthand of `whereami` like Pry:

```rb
irb(main)[01:0]> @

From: /tmp/a.rb @ line 2 :

    1: def test
 => 2:   binding.irb
    3: end
    4: test

=> nil
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -426,6 +426,9 @@ module IRB
     def initialize(workspace = nil, input_method = nil)
       @context = Context.new(self, workspace, input_method)
       @context.main.extend ExtendCommandBundle
+      @context.command_aliases.each do |alias_name, cmd_name|
+        @context.main.install_alias_method(alias_name, cmd_name)
+      end
       @signal_status = :IN_IRB
       @scanner = RubyLex.new
     end
@@ -433,9 +436,6 @@ module IRB
     def run(conf = IRB.conf)
       conf[:IRB_RC].call(context) if conf[:IRB_RC]
       conf[:MAIN_CONTEXT] = context
-      conf[:COMMAND_ALIASES].each do |alias_name, cmd_name|
-        context.main.install_alias_method(alias_name, cmd_name)
-      end
 
       prev_trap = trap("SIGINT") do
         signal_handle
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -433,6 +433,9 @@ module IRB
     def run(conf = IRB.conf)
       conf[:IRB_RC].call(context) if conf[:IRB_RC]
       conf[:MAIN_CONTEXT] = context
+      conf[:COMMAND_ALIASES].each do |alias_name, cmd_name|
+        context.main.install_alias_method(alias_name, cmd_name)
+      end
 
       prev_trap = trap("SIGINT") do
         signal_handle
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -67,7 +67,7 @@ class RubyLex
         else
           # Accept any single-line input starting with a non-identifier alias (ex: @, $)
           command = code.split(/\s/, 2).first
-          if context&.symbol_alias(command)
+          if context.symbol_alias(command)
             next true
           end
 
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -65,6 +65,12 @@ class RubyLex
             false
           end
         else
+          # Accept any single-line input starting with a non-identifier alias (ex: @, $)
+          command = code.split(/\s/, 2).first
+          if !command.match?(/\A\w+\z/) && context&.main&.respond_to?(command)
+            next true
+          end
+
           code.gsub!(/\s*\z/, '').concat("\n")
           ltype, indent, continue, code_block_open = check_state(code, context: context)
           if ltype or indent > 0 or continue or code_block_open
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -67,7 +67,7 @@ class RubyLex
         else
           # Accept any single-line input starting with a non-identifier alias (ex: @, $)
           command = code.split(/\s/, 2).first
-          if !command.match?(/\A\w+\z/) && context&.main&.respond_to?(command)
+          if context&.symbol_alias(command)
             next true
           end
 
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -576,7 +576,7 @@ module TestIRB
       ])
       IRB.init_config(nil)
       IRB.conf[:COMMAND_ALIASES] = { :'$' => :show_source }
-      workspace = IRB::WorkSpace.new(build_main)
+      workspace = IRB::WorkSpace.new(Object.new)
       IRB.conf[:VERBOSE] = false
       irb = IRB::Irb.new(workspace, input)
       IRB.conf[:MAIN_CONTEXT] = irb.context
@@ -635,7 +635,7 @@ module TestIRB
       ])
       IRB.init_config(nil)
       IRB.conf[:COMMAND_ALIASES] = { :'@' => :whereami }
-      workspace = IRB::WorkSpace.new(build_main)
+      workspace = IRB::WorkSpace.new(Object.new)
       irb = IRB::Irb.new(workspace, input)
       IRB.conf[:MAIN_CONTEXT] = irb.context
       out, err = capture_output do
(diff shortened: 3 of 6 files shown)


st0012:
### Using `IRB.conf` instead of `IRB::ExtendCommandBundle`?

I think `IRB::ExtendCommandBundle` is an internal module and probably better not be used directly by users? In contrary, `IRB.conf` has been used to configure from prompt to at_exit callback. So having something like

```rb
IRB.conf[:command_aliases] = { :"@" => :whereami }
```

Seems to make sense too and could also help feature discoverability.

---

### Using Hash lookup instead of method lookup?

Do you think we can simplify evaluation check by keeping a hash of symbol aliases and just do a hash lookup?

Something like:

```diff
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -480,8 +480,9 @@ module IRB

       # Transform a non-identifier alias (ex: @, $)
       command = line.split(/\s/, 2).first
-      if !command.match?(/\A\w+\z/) && main.respond_to?(command)
-        line = line.gsub(/\A#{Regexp.escape(command)}/, main.method(command).original_name.to_s)
+
+      if aliased_cmd = ExtendCommandBundle.symbol_aliases[command.to_sym]
+        line = line.gsub(command, aliased_cmd.to_s)
       end

       set_last_value(@workspace.evaluate(self, line, irb_path, line_no))
diff --git a/lib/irb/extend-command.rb b/lib/irb/extend-command.rb
index 3a94d45..8edf208 100644
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -45,6 +45,8 @@ module IRB # :nodoc:
       [:quit, :irb_exit, OVERRIDE_PRIVATE_ONLY],
     ]

+    @SYMBOL_ALIASES = {}
+
     @EXTEND_COMMANDS = [
       [
         :irb_current_working_workspace, :CurrentWorkingWorkspace, "cmd/chws",
@@ -224,6 +226,14 @@ module IRB # :nodoc:

     def self.def_alias_command(alias_name, cmd_name, flag = NO_OVERRIDE)
       @ALIASES.push [alias_name, cmd_name, flag]
+
+      unless alias_name.match?(/\A\w+\z/)
+        @SYMBOL_ALIASES[alias_name] = cmd_name
+      end
+    end
+
+    def self.symbol_aliases
+      @SYMBOL_ALIASES
     end

     # Installs alias methods for the default irb commands, see
@@ -245,6 +255,7 @@ module IRB # :nodoc:
         }
       else
         Kernel.print "irb: warn: can't alias #{to} from #{from}.\n"
+        @SYMBOL_ALIASES.delete(to) if @SYMBOL_ALIASES.key?(to)
       end
     end
```

I think it's simpler and we'll be able to support truffleruby too.

---

And for testing, can we add these scenarios too?

1. Aliasing commands that take arguments, like `show_source`.
2. Non-symbol aliases, like `wi` for `whereami`.
3. When aliasing with `@` or `$`, calling `@foo` or `$foo` won't be affected.

k0kubun:
> Using IRB.conf instead of IRB::ExtendCommandBundle?

Sounds good. I've had `IRB::ExtendCommandBundle` in my ~/.irbrc, but it's for monkey-patching purposes, so it'd be just me. I used `IRB.conf` instead https://github.com/ruby/irb/pull/426/commits/b9feed47234b69634887b182055c9b2242b26077. For consistency with other configs, I used a SHOUT_CASE though.

> Using Hash lookup instead of method lookup?

Addressed it in https://github.com/ruby/irb/pull/426/commits/711f4f833092e603168368b291531103ef86029e

> And for testing, can we add these scenarios too?

:+1: https://github.com/ruby/irb/pull/426/commits/86d6ea02af1f06d5146e05dfaca5d8a6f45b4bfe

lib/irb.rb
```diff
@@ -433,6 +433,9 @@ def initialize(workspace = nil, input_method = nil)
     def run(conf = IRB.conf)
       conf[:IRB_RC].call(context) if conf[:IRB_RC]
       conf[:MAIN_CONTEXT] = context
+      conf[:COMMAND_ALIASES].each do |alias_name, cmd_name|
```
st0012, line 436: Can we do this in `Irb#initialize` instead? That way we won't need to duplicate the same logic in tests. 
Also, since we've stored the aliases in `Context`'s `@command_aliases`, can we expose it as reader and use it as the only source of truth?

test/irb/test_cmd.rb
```diff
@@ -610,5 +628,57 @@ def test_whereami
...
+      assert_empty err
+      assert_match(/"foo"/, out)
+      assert_match(/"bar"/, out)
+    end
+
+    private
+
+    def build_main
+      main = Object.new
+      main.extend(IRB::ExtendCommandBundle)
```
st0012, line 676: This probably isnt' necessary because [`Irb#initialize` already does it](https://github.com/ruby/irb/blob/master/lib/irb.rb#L428)?

test/irb/test_cmd.rb
```diff
@@ -610,5 +628,57 @@ def test_whereami
...
+      # Simulate part of the IRB::Irb#run behavior
+      IRB.conf[:COMMAND_ALIASES].each do |alias_name, cmd_name|
```
st0012, line 678: We can get rid of this if we move the command installation to `Irb#initialize`.

lib/irb/ruby-lex.rb
```diff
@@ -65,6 +65,12 @@ def set_input(io, p = nil, context: nil, &block)
             false
           end
         else
+          # Accept any single-line input starting with a non-identifier alias (ex: @, $)
+          command = code.split(/\s/, 2).first
+          if context&.symbol_alias(command)
```
st0012, line 70: Not directly related to this PR, but outside of tests the `context` should never be `nil`. [This](https://github.com/ruby/irb/blob/master/lib/irb.rb#L489) is its only caller of `set_input`, and it must not passing nil context, otherwise `@context.io` would already raise error.

test/irb/test_cmd.rb
st0012: And if the above 2 comments are correct, we probably won't need this `build_main` anymore.

lib/irb.rb
k0kubun: > Can we do this in Irb#initialize instead? That way we won't need to duplicate the same logic in tests.

What about the `conf = IRB.conf` in the argument? It seems weird to handle `IRB.conf` in `#initialize` when you could override it in the argument of `#run`.

> Also, since we've stored the aliases in Context's @command_aliases

What do you mean? We only have `@symbol_aliases` to achieve your "Hash look up" idea. If you only use `@symbol_aliases` here, it will not alias non-symbol aliases.

I can change it to `@command_aliases` and fix other stuff though.

test/irb/test_cmd.rb
k0kubun: It was for defining `install_alias_method`.

test/irb/test_cmd.rb
k0kubun: Your patch will stop supporting non-symbol aliases in `IRB.conf[:COMMAND_ALIASES]`.

lib/irb/ruby-lex.rb
k0kubun: Thanks for https://github.com/ruby/irb/pull/427 :+1: Just pushed https://github.com/ruby/irb/pull/426/commits/ed0406fa479bd4da7847f98ed7ab891ffa1c4e70

lib/irb.rb
k0kubun: addressed it in https://github.com/ruby/irb/pull/426/commits/02e9369c059b4c45e829d712bb26610f1a3453b8

test/irb/test_cmd.rb
k0kubun: fixed it appropriately in https://github.com/ruby/irb/pull/426/commits/02e9369c059b4c45e829d712bb26610f1a3453b8

st0012 (approved):
Thank you, I love it 👍 
Will we have another discussion about default aliases?

k0kubun:
Thank you for the reviews!

> Will we have another discussion about default aliases?

Yeah, I prefer that doing it separately. I have two somewhat related changes and I'd like to prioritize them.

I thought about making `@` not a default, but now I'm actually leaning towards it since regular users simply wouldn't have `~/.irbrc`. Anyway I'll file it later.

PR #429, opened 2022-11-03
Provide a base test class and let tests restore encodings conveniently

st0012:
Because `RelineInputMethod#initializes` calls `set_encoding`, which changes stdio/out/err and Encoding's default encoding values, we need to make sure any test that directly or indirectly (e.g. through Context) initializes `RelineInputMethod` restores encodings.

`ReadlineInputMethod` also changes encodings but currently no tests cover it.

### DIFF
--- a/test/irb/helper.rb
+++ b/test/irb/helper.rb
@@ -2,6 +2,19 @@ require "test/unit"
 
 module TestIRB
   class TestCase < Test::Unit::TestCase
+    def save_encodings
+      @default_encoding = [Encoding.default_external, Encoding.default_internal]
+      @stdio_encodings = [STDIN, STDOUT, STDERR].map {|io| [io.external_encoding, io.internal_encoding] }
+    end
+
+    def restore_encodings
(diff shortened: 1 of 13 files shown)


st0012:
FWIW, I also patched this locally to make sure nothing is missed:

```diff
--- a/test/irb/helper.rb
+++ b/test/irb/helper.rb
@@ -2,6 +2,15 @@ require "test/unit"

 module TestIRB
   class TestCase < Test::Unit::TestCase
+    def teardown
+      assert_equal(Encoding.default_external, STDIN.external_encoding)
+      assert_equal(nil, STDIN.internal_encoding)
+      assert_equal(nil, STDOUT.external_encoding)
+      assert_equal(nil, STDOUT.internal_encoding)
+      assert_equal(nil, STDERR.external_encoding)
+      assert_equal(nil, STDERR.internal_encoding)
+    end
+
```

k0kubun (approved):
I hope this will reduce the number of future problems in encoding.

PR #430, opened 2022-11-04
Support non-string input in show_source

k0kubun:
This PR allows a non-String literal input in `show_source`, just like Pry.

## Example

```rb
$ irb -rerb
irb(main)[01:0]> show_source ERB#result

From: /home/k0kubun/.rbenv/versions/3.1.2/lib/ruby/3.1.0/erb.rb:901

  def result(b=new_toplevel)
    unless @_init.equal?(self.class.singleton_class)
      raise ArgumentError, "not initialized"
    end
    eval(@src, b, (@filename || '(erb)'), @lineno)
  end

=> nil
```

### DIFF
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -147,6 +147,20 @@ module IRB # :nodoc:
 
     ]
 
+    # Convert a command name to its implementation class if such command exists
+    def self.load_command(command)
+      command = command.to_sym
+      @EXTEND_COMMANDS.each do |cmd_name, cmd_class, load_file, *aliases|
+        next if cmd_name != command && aliases.all? { |alias_name, _| alias_name != command }
+
+        if !defined?(ExtendCommand) || !ExtendCommand.const_defined?(cmd_class, false)
+          require_relative load_file
+        end
+        return ExtendCommand.const_get(cmd_class, false)
+      end
+      nil
+    end
+
     # Installs the default irb commands:
     #
     # +irb_current_working_workspace+::   Context#main
--- a/lib/irb/cmd/show_source.rb
+++ b/lib/irb/cmd/show_source.rb
@@ -9,6 +9,24 @@ module IRB
 
   module ExtendCommand
     class ShowSource < Nop
+      class << self
+        def transform_args(args)
+          # Return a string literal as is for backward compatibility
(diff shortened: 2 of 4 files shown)


lib/irb/extend-command.rb
```diff
@@ -147,6 +147,20 @@ def irb_context
...
     ]
 
+    # Convert a command name to its implementation class if such command exists
+    def self.load_command(command)
+      command = command.to_sym
+      @EXTEND_COMMANDS.each do |cmd_name, cmd_class, load_file, *aliases|
+        next if cmd_name != command && aliases.all? { |alias_name, _| alias_name != command }
+
+        if !defined?(ExtendCommand) || !ExtendCommand.const_defined?(cmd_class, false)
+          require_relative load_file
```
st0012, line 157: I think [this line](https://github.com/ruby/irb/blob/master/lib/irb/extend-command.rb#L195) can be removed because of this change?

k0kubun: No, this require_relative is executed only when it's used at the beginning of a file, but because it's a method, it could still be called in the middle of a line. So I don't think it's still safe to remove it.

st0012: Sorry I don't get it. Here's my understanding:

- This piece of code is always run **before** the command methods (e.g. `irb_show_source`) being called.
- It checks if the command definition (`ExtendedCommand::ShowSource`) has been loaded yet. If it hasn't, we load it here with `require_relative "cmd/show_source"`.
- So by the time the `irb_show_source` method is reached, the definition should always be there. And therefore the `require_relative` I linked is obsolete now.

st0012: Ah I didn't know that's an expected usage, thanks for sharing. I don't see a point of doing that though because we generally don't care about what the command returns but its side-effects (printing to stdin/out). 
But I think supporting it or not is outside of the scope of this PR.

st0012: With that being said, you think you can add another test case for `p irb_show_source("ERB#result")`?

k0kubun: :+1: done https://github.com/ruby/irb/pull/430/commits/e8b1c9d099ea4c40985cac0ce1c8b97079122376

PR #436, opened 2022-11-08
Add execute_lines to reduce command tests' boilerplate code

st0012:
Right now we have a lot of repeated setup for command tests. I think we can hide them from contributors with a helper.

(Tested against `ruby/ruby` with `make test-all -j8 TESTS=../test/irb/ TESTOPTS="--no-retry"`)


k0kubun (approved):
😍

PR #438, opened 2022-11-10
Make $ and @ default aliases

k0kubun:
These are non-disturbing syntaxes and are also well-known in Pry. Given their long names, they are very convenient. It seems useful to make them available without ~/.irbrc.


st0012 (approved):
People are going to love this 😄

PR #446, opened 2022-11-18
Add `debug` command

### LINKED ISSUE #425: Add a command to activate and switch to `ruby/debug` console
## Background

When debugging programs with Ruby 2.7+ , it's common to start with adding `binding.irb` breakpoints in the code. This is because `irb` is the only console-based tool (including REPLs and debuggers) that doesn't require installation nor require.

However, because `irb` is a REPL and lacks debugger features like step-debugging, frame navigation...etc., users may need to switch to debuggers like [ruby/debug](https://github.com/ruby/debug) for advanced debugging. 

And currently, that switch looks like this:

1. Exit the current `irb` session.
2. Add `require "debug"` if it's not yet required.
3. Replace `binding.irb` with `ruby/debug`'s breakpoints like `binding.b` or `debugger` (they're the same).
4. Rerun the program and start debugging.

It's a 4-step process and requires switching between terminal and editor.

During a discussion with @k0kubun, he mentioned that maybe `irb` can help reduce the pain.

## Proposal

Add a new `debug` command to `irb`, which when being executed will

1. Run `require "debug/session"` and start the debugger.
    - If it causes `LoadError`, it means the new `debug` gem is not installed. So it will print a message to remind the user instead.
2. Replace the current `irb` session with a new `debug` console.
3. The user can start debugging at the same location as `binding.irb` was placed.

With this command, a user's `irb` -> `debug` switch would be:

1. Type `debug` in the `irb` session.
2. Start debugging.

## Demo

irb debug command

## Discussion Points

- Should the command be `debug` or `debugger`?
    - `debug`:
        - As a verb it seems to convey the intention well as "Debug in the current context"
        - On the other hand, it's another debugging related name to remember.
    - `debugger`:
        - It's the same as the `debugger` statement `ruby/debug` already has.
        - It may give a false impression that we're simply calling the same statement in `irb`.
- When users exit the debugger, should they exit completely or return to `irb`?
    - IMO, it should exit completely so users don't need to exit twice.
- Because `irb` registers autocompletion logic to `reline`, `debug`'s console (also powered by `reline`) will inherit that too.
    - Perhaps it should be suspended to avoid unexpected side-effects?

st0012:
This implements and closes #425

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -393,6 +393,8 @@ module IRB
   end
 
   class Irb
+    DIR_NAME = __dir__
+
     ASSIGNMENT_NODE_TYPES = [
       # Local, instance, global, class, constant, instance, and index assignment:
       #   "foo = bar",
@@ -432,6 +434,16 @@ module IRB
       end
       @signal_status = :IN_IRB
       @scanner = RubyLex.new
+    end
+
+    def debug_break
+      # it means the debug command is executed
+      if defined?(DEBUGGER__) && DEBUGGER__.respond_to?(:original_capture_frames)
+        # after leaving this initial breakpoint, revert the capture_frames patch
+        DEBUGGER__.singleton_class.send(:alias_method, :capture_frames, :original_capture_frames)
+        # and remove the redundant method
+        DEBUGGER__.singleton_class.send(:undef_method, :original_capture_frames)
+      end
     end
 
     def run(conf = IRB.conf)
@@ -931,5 +943,6 @@ class Binding
     binding_irb = IRB::Irb.new(workspace)
     binding_irb.context.irb_path = File.expand_path(source_location[0])
     binding_irb.run(IRB.conf)
+    binding_irb.debug_break
   end
 end
--- a/lib/irb/cmd/debug.rb
+++ b/lib/irb/cmd/debug.rb
@@ -1,0 +1,33 @@
+require_relative "nop"
+
+module IRB
+  # :stopdoc:
+
+  module ExtendCommand
+    class Debug < Nop
+      def execute(*args)
+        require "debug/session"
+        DEBUGGER__.start(nonstop: true)
+        DEBUGGER__.singleton_class.send(:alias_method, :original_capture_frames, :capture_frames)
+
+        def DEBUGGER__.capture_frames(skip_path_prefix)
+          frames = original_capture_frames(skip_path_prefix)
+          frames.reject! do |frame|
+            frame.realpath&.start_with?(::IRB::Irb::DIR_NAME) || frame.path.match?(/internal:prelude/)
+          end
+          frames
+        end
+
+        file, lineno = IRB::Irb.instance_method(:debug_break).source_location
+        DEBUGGER__::SESSION.add_line_breakpoint(file, lineno + 1, oneshot: true, hook_call: false)
+        # exit current Irb#run call
+        throw :IRB_EXIT
+      rescue LoadError => e
+        puts <<~MSG
+          You need to install the debug gem before using this command.
+          If you use `bundle exec`, please add `gem "debug"` into your Gemfile.
+        MSG
+      end
+    end
+  end
+end
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -117,6 +117,10 @@ module IRB # :nodoc:
       ],
 
       [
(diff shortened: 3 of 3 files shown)


k0kubun:
I've played with this implementation myself for a short while. This seems to be written pretty well. Great job :+1:

Some minor feedback:

* It silently stops the REPL on irb(1). I think we need to support only `binding.irb`, so we should probably just print an error for the irb(1) case instead of stopping the REPL.
* It's probably better to avoid the name `original_capture_frames` since other gems could also do the same thing. Maybe something like `capture_frames_without_irb` is better.
* It might be nice if we can avoid method definition/undef on every `debug` command invocation because it flushes some method caches. However, it seems not straightforward to disable the behavior after the first break if you use `Module#prepend` to override it. I tried and failed it, so it seems okay to leave it until we come up with a solution.

A question:

* I wonder if we can go back to the original `binding.irb` after invoking `debug` accidentally or when you want to use IRB-only commands again. Sure, you can use `irb` command on the debugger, but it's a nested REPL and the "quit twice" problem re-appears here; you need to quit the debugger after using IRB. It feels like we're just moving the "quit twice" problem from IRB to debug.gem.
  * Can we implement a debug.gem command to quit the debugger and run `binding.irb` in the current frame without having the debugger as the parent?

---

Normally, I would leave it to you to address them before merging it, but I have one idea I'd like to try on this implementation. So let me merge this first, file a pull request to address the first two points myself, and file another one to add one change if that works out.

k0kubun:
> In this case, I think the question should be: why doesn't debug have that functionality? If it's helpful for debugging, we should implement it in debug anyway instead of letting people to switch back to irb.

If we get to support the full functionality of `ls` and `show_source` that we've recently added to IRB in debug.gem as well, I'll likely not need to go back to IRB. Maybe it's not a problem if debug.gem tries to be a superset of IRB.

off topic: Speaking of making it a superset, I'm still kind of uncomfortable that C-d asks "Really quit? [Y/n]" and I'd like to add a configuration to make it work exactly like `continue`, but that's a separate discussion :)

st0012:
> If we get to support the full functionality of whereami, ls, and show_source that we've recently added to IRB in debug.gem as well, I'll likely not need to go back to IRB.

It now has `whereami` and `ls` (except aliases and `-g` option though). I also opened [a proposal for `show_source`](https://github.com/ruby/debug/issues/526) but I currently don't have energy to follow through. If you can chime in perhaps we can have all 3 of them in Ruby 3.2 😄 

>  I'd like to add a configuration to make it work exactly like continue (or C-d on IRB)

Personally I'd hope it to be `q!` (quit without asking). But `continue` makes sense as well. This does feel like something we can customise. (BTW, [this](https://github.com/ruby/debug/blob/master/lib/debug/local.rb#L72) should be the line to change.)

k0kubun:
> It now has whereami and ls (except aliases and -g option though)

I'm talking exactly about `@` alias and `-G` option 🙂 

> Personally I'd hope it to be q! (quit without asking). But continue makes sense as well.

So, when you put binding.irb and press C-d, do you want to quit the process instead of moving until it hits another binding.pry? binding.irb is like a breakpoint for me, so personally I don't understand debug.gem just quits it instead of continuing it (unlike pry-byebug) or even asks it.

st0012:
In `debug`, I usually can get the information I need or I can get with one breakpoint. So when I leave it, I usually want to restart the program for a clean state or test my code. With this usage, quit (by typing `q!` instead of using `C-d`) is more convenient to me. 

But with `irb` or `pry`, multiple breakpoints are usually needed to go to different places, so continue would be more convenient than quit here.

k0kubun:
I also quit irb or pry immediately very often, but still less frequently than C-d to continue. Whether you put a breakpoint by a command or leaving a method call, it doesn't change the fact that you use contiune multiple times and quit only once in one program execution. So somehow less frequently used command is made easier to input. The "Really quit?" prompt would be very annoying if you need to use it as frequent as contiune. 

I heard it's made that way because it's how gdb works. I'm the guy who use pry-byebug or irb more often, and debug is probably designed by people who use gdb more.

st0012:
> I'm the guy who use pry-byebug or irb more often, and debug is probably designed by people who use gdb more.

I agree and feel that. And [this](https://github.com/ruby/debug/pull/632) is another behaviour that I wish it to match `byebug` instead of `gdb`

PR #447, opened 2022-11-18
Minor fixes on debug command

k0kubun:
Addressed the first two points in https://github.com/ruby/irb/pull/446#issuecomment-1319732273 myself, and added some comments and minor refactoring changes.

### DIFF
--- a/lib/irb/cmd/debug.rb
+++ b/lib/irb/cmd/debug.rb
@@ -5,28 +5,68 @@ module IRB
 
   module ExtendCommand
     class Debug < Nop
-      def execute(*args)
-        require "debug/session"
-        DEBUGGER__.start(nonstop: true)
-        DEBUGGER__.singleton_class.send(:alias_method, :original_capture_frames, :capture_frames)
+      BINDING_IRB_FRAME_REGEXPS = [
+        '<internal:prelude>',
+        binding.method(:irb).source_location.first,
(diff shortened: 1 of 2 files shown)


lib/irb/cmd/debug.rb
```diff
@@ -5,28 +5,68 @@ module IRB
...
-        DEBUGGER__.singleton_class.send(:alias_method, :original_capture_frames, :capture_frames)
-
-        def DEBUGGER__.capture_frames(skip_path_prefix)
-          frames = original_capture_frames(skip_path_prefix)
-          frames.reject! do |frame|
-            frame.realpath&.start_with?(::IRB::Irb::DIR_NAME) || frame.path.match?(/internal:prelude/)
-          end
-          frames
+        unless binding_irb?
+          puts "`debug` command is supported only on binding.irb"
```
st0012, line 16: nitpick: I think the message could be a bit more explicit:

```
`debug` command is only available when IRB is started with binding.irb
```

st0012 (approved):
Thanks for the improvement 👍

lib/irb/cmd/debug.rb
k0kubun:
```suggestion
          puts "`debug` command is only available when IRB is started with binding.irb"
```

PR #448, opened 2022-11-18
Discover and load debug.gem even if it's not in Gemfile

k0kubun:
Note: This PR contains the diff of https://github.com/ruby/irb/pull/447 as well. Please look at that first, and ignore the first commit when reviewing this PR.

So this is what I wanted to do in this command. The most useful and unique feature of `binding.irb` is that you don't need to write that in Gemfile. This `load_bundled_debug_gem` implementation applies it to the `debug` command on `binding.irb`.

This is a bit hacky for sure, but you can always avoid the risk by just adding debug.gem to Gemfile. Thus I believe this is a superset in terms of supported use cases.

### DIFF
--- a/lib/irb/cmd/debug.rb
+++ b/lib/irb/cmd/debug.rb
@@ -5,28 +5,68 @@ module IRB
 
   module ExtendCommand
     class Debug < Nop
-      def execute(*args)
-        require "debug/session"
-        DEBUGGER__.start(nonstop: true)
-        DEBUGGER__.singleton_class.send(:alias_method, :original_capture_frames, :capture_frames)
+      BINDING_IRB_FRAME_REGEXPS = [
+        '<internal:prelude>',
+        binding.method(:irb).source_location.first,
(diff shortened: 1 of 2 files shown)


lib/irb/cmd/debug.rb
```diff
@@ -5,28 +5,90 @@ module IRB
...
+        return false unless debug_gem
+
+        # Attempt to forcibly load the bundled gem
+        $LOAD_PATH << "#{debug_gem}/lib"
+        begin
+          require "debug/session"
+        rescue LoadError
+          false
+        else
+          puts "Loaded #{File.basename(debug_gem)}"
```
st0012, line 89: Can these 2 lines be put under `require "debug/session"` so we don't need the `else`?

st0012: I agree with your argument and that's why I didn't made the same argument for the similar rescue block in `setup_debugger`.
So the difference between our views is actually how much of the chance this puts can cause the error we're trying to rescue. From my personal understanding on `File.basename` and `puts`, I thought it'd be very small so it's worth moving it here.

Issue #445, opened 2022-11-17
Autocompletion Meta Issue


st0012:
## Background

Since its introduction in version `1.4.0`, the autocompletion feature has received mixed feedback. We can see many issues and comments in the repo mentioning its problems. Personally I find the feature useful and pretty like it, but I also experienced some problems with it.

However, many feedback were given through various channels (GH issues, Twitter, Slack channels...etc.). And from my experience, people mentioned different problems, not just one. 

So I decided to reach out to the community to ask how they feel about this feature, as well as the specific problems they experienced. And then combine with feedback I collected from other channels, I would make a list of problems, so we can discuss and solve them more effectively.

## My Poll Results

I did 2 polls, on one [Twitter](https://twitter.com/_st0012/status/1589391456355385345) and another one on [ruby.social](https://ruby.social/@st0012/109299337468404201):

<img width="45%" alt="Poll on Twitter" src="https://user-images.githubusercontent.com/5079556/202045148-2d6e30fc-93d8-47ec-9fd9-2372e53a5c40.png">  <img width="45%" alt="Poll on ruby.social" src="https://user-images.githubusercontent.com/5079556/202045302-c1e2a076-acb8-4be8-bf2b-105d15d998ab.png">

We can see that in both polls:

- Users generally feel this feature is helpful
- But more than half of them think it has issues
- There's a non-small percentage of users decided to disable it completely

## Problems Collected From Users

Based on the comments from these polls as well as comments in related GH issues, I think we can list its common issues (ordered by frequency):

1. Lack of debounce (mentioned multiple times in the poll comments)
1. Makes backspacing slow (mentioned multiple times in the poll comments)
1. Slow in production (mentioned in [#351](https://github.com/ruby/irb/issues/351) and poll comments)
1. Background colour could be problematic (mentioned in [#351](https://github.com/ruby/irb/issues/351) and [this ticket](https://bugs.ruby-lang.org/issues/18996))
1. Dropdown moves the input line's position (mentioned in [#351](https://github.com/ruby/irb/issues/351#issuecomment-1113903167))
1. Doesn't provide helpful suggestions when chaining method calls (https://github.com/ruby/irb/issues/339)

## What's Next?

I'll open an issue for every problem if they don't already have one, start from the top of the list. 

[...]

PR #449, opened 2022-11-18
Add commands to start and use the debugger

k0kubun:
This PR adds the following IRB commands that start the debugger and seamlessly run the command at the same time:

* `break`
* `delete`
* `next`
* `step`
* `continue`
* `finish`
* `backtrace`
* `info`
* `catch`

Given that we're trying to make debug.gem a superset of IRB functionality https://github.com/ruby/irb/pull/446#issuecomment-1319888659, there should be few issues in switching to debug.gem.

### DIFF
--- a/lib/irb/cmd/debug.rb
+++ b/lib/irb/cmd/debug.rb
@@ -11,7 +11,7 @@ module IRB
       ].map { |file| /\A#{Regexp.escape(file)}:\d+:in `irb'\z/ }
       IRB_DIR = File.expand_path('..', __dir__)
 
-      def execute(*args)
+      def execute(debug_command = nil)
         unless binding_irb?
           puts "`debug` command is only available when IRB is started with binding.irb"
           return
@@ -25,11 +25,16 @@ module IRB
           return
         end
 
+        command = nil
+        if debug_command
+          command = ['irb', nil, debug_command]
+        end
+
         # To make debugger commands like `next` or `continue` work without asking
         # the user to quit IRB after that, we need to exit IRB first and then hit
         # a TracePoint on #debug_break.
         file, lineno = IRB::Irb.instance_method(:debug_break).source_location
-        DEBUGGER__::SESSION.add_line_breakpoint(file, lineno + 1, oneshot: true, hook_call: false)
+        DEBUGGER__::SESSION.add_line_breakpoint(file, lineno + 1, oneshot: true, hook_call: false, command: command)
         # exit current Irb#run call
         throw :IRB_EXIT
       end
--- a/lib/irb/cmd/debug.rb
+++ b/lib/irb/cmd/debug.rb
@@ -11,7 +11,7 @@ module IRB
       ].map { |file| /\A#{Regexp.escape(file)}:\d+:in `irb'\z/ }
       IRB_DIR = File.expand_path('..', __dir__)
 
-      def execute(debug_command = nil)
+      def execute(pre_cmds: nil, do_cmds: nil)
         unless binding_irb?
           puts "`debug` command is only available when IRB is started with binding.irb"
           return
@@ -26,8 +26,8 @@ module IRB
         end
 
         command = nil
-        if debug_command
-          command = ['irb', nil, debug_command]
+        if pre_cmds || do_cmds
+          command = ['irb', pre_cmds, do_cmds]
         end
 
         # To make debugger commands like `next` or `continue` work without asking
--- a/lib/irb/cmd/debug.rb
+++ b/lib/irb/cmd/debug.rb
@@ -25,16 +25,19 @@ module IRB
           return
         end
 
-        command = nil
+        options = { oneshot: true, hook_call: false }
         if pre_cmds || do_cmds
-          command = ['irb', pre_cmds, do_cmds]
+          options[:command] = ['irb', pre_cmds, do_cmds]
+        end
+        if DEBUGGER__::LineBreakpoint.instance_method(:initialize).parameters.include?([:key, :skip_src])
+          options[:skip_src] = true
         end
 
         # To make debugger commands like `next` or `continue` work without asking
         # the user to quit IRB after that, we need to exit IRB first and then hit
         # a TracePoint on #debug_break.
         file, lineno = IRB::Irb.instance_method(:debug_break).source_location
-        DEBUGGER__::SESSION.add_line_breakpoint(file, lineno + 1, oneshot: true, hook_call: false, command: command)
+        DEBUGGER__::SESSION.add_line_breakpoint(file, lineno + 1, **options)
         # exit current Irb#run call
         throw :IRB_EXIT
       end
--- a/Gemfile
+++ b/Gemfile
@@ -11,4 +11,5 @@ group :development do
   gem "stackprof" if is_unix && !is_truffleruby
   gem "test-unit"
   gem "reline", github: "ruby/reline" if ENV["WITH_LATEST_RELINE"] == "true"
+  gem "debug"
 end
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ The following commands are available on IRB.
   * Show the source code around binding.irb again.
 * `debug`
   * Start the debugger of debug.gem.
+* `break`, `delete`, `next`, `step`, `continue`, `finish`
+  * Start the debugger of debug.gem and run the command on it.
 
 ## Documentation
 
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ The following commands are available on IRB.
   * Show the source code around binding.irb again.
 * `debug`
   * Start the debugger of debug.gem.
-* `break`, `delete`, `next`, `step`, `continue`, `finish`
+* `break`, `delete`, `next`, `step`, `continue`, `finish`, `backtrace`, `info`, `catch`
   * Start the debugger of debug.gem and run the command on it.
 
 ## Documentation
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -96,7 +96,7 @@ require_relative "irb/easter-egg"
 #   * Show the source code around binding.irb again.
 # * debug
 #   * Start the debugger of debug.gem.
-# * `break`, `delete`, `next`, `step`, `continue`, `finish`
+# * break, delete, next, step, continue, finish, backtrace, info, catch
 #   * Start the debugger of debug.gem and run the command on it.
 #
 # == Configuration
@@ -472,10 +472,6 @@ module IRB
     def initialize(workspace = nil, input_method = nil)
       @context = Context.new(self, workspace, input_method)
       @context.main.extend ExtendCommandBundle
-      @context.command_aliases.each do |alias_name, cmd_name|
-        next if @context.symbol_alias?(alias_name) || @context.main.respond_to?(alias_name)
-        @context.main.install_alias_method(alias_name, cmd_name)
-      end
       @signal_status = :IN_IRB
       @scanner = RubyLex.new
     end
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -427,7 +427,7 @@ module IRB
       @context = Context.new(self, workspace, input_method)
       @context.main.extend ExtendCommandBundle
       @context.command_aliases.each do |alias_name, cmd_name|
-        next if @context.symbol_alias(alias_name)
+        next if @context.symbol_alias?(alias_name) || @context.main.respond_to?(alias_name)
         @context.main.install_alias_method(alias_name, cmd_name)
       end
       @signal_status = :IN_IRB
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -96,6 +96,8 @@ require_relative "irb/easter-egg"
 #   * Show the source code around binding.irb again.
 # * debug
 #   * Start the debugger of debug.gem.
+# * `break`, `delete`, `next`, `step`, `continue`, `finish`
+#   * Start the debugger of debug.gem and run the command on it.
 #
 # == Configuration
 #
--- a/lib/irb/cmd/backtrace.rb
+++ b/lib/irb/cmd/backtrace.rb
@@ -6,13 +6,13 @@ module IRB
   # :stopdoc:
 
   module ExtendCommand
-    class Info < Debug
+    class Backtrace < Debug
       def self.transform_args(args)
         args&.dump
       end
 
       def execute(*args)
-        super(pre_cmds: ["info", *args].join(" "))
+        super(pre_cmds: ["backtrace", *args].join(" "))
       end
     end
   end
--- a/lib/irb/cmd/break.rb
+++ b/lib/irb/cmd/break.rb
@@ -7,8 +7,12 @@ module IRB
 
   module ExtendCommand
     class Break < Debug
-      def execute(*args)
-        super(['break', *args].join(' '))
+      def self.transform_args(args)
+        args&.dump
+      end
+
+      def execute(args = nil)
+        super(pre_cmds: "break #{args}")
       end
     end
   end
--- a/lib/irb/cmd/break.rb
+++ b/lib/irb/cmd/break.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require_relative "debug"
+
+module IRB
+  # :stopdoc:
+
+  module ExtendCommand
+    class Break < Debug
+      def execute(*args)
+        super(['break', *args].join(' '))
+      end
+    end
+  end
+
+  # :startdoc:
+end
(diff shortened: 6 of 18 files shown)


st0012:
I like the idea but IMO we need to have more discussion before introducing these commands. Currently I have these questions:

- What subset of functionalities make sense in the context of `irb`? Just off the top of my head:
    - Stepping commands
    - Frame control commands
    - Breakpoint commands
- When adding a command, do we want to completely replicate it here, or just part of it? For example:
    - In `debug`, I usually use command aliases like `n` (`next`) and `c` (`continue`). So typing full commands in `irb` is not convenient to me.
        - But on the other hand, if we make `n` and `c` commands in `irb`, it could annoy many users as they could easily be variable names.
    - Commands like `break` or `catch` support many options, like `break Foo#bar if: a == b do: caller`. Do we want to support them as well?
- How can we help users discover them? `debug` has `help` command to help users learn all the commands' usages. But `irb` doesn't have it and we also don't have good documentation atm. This is already a problem, but it'll be more obvious after we introducing these commands.
    - Side note: I want to rename the current `help` command to something like `doc` and let `help` command prints command usages.

Implementation-wise, I think we also need to figure how to write tests for them because the scenarios to consider will grow very fast. `debug` has a test framework that allows us to test the debugger in a separate process so we can verify its functionalities end-to-end ([example](https://github.com/ruby/debug/blob/master/test/console/break_test.rb#L142-L174)).

I also found that the `break` command isn't working:

```
From: test.rb @ line 8 :

     3:     10
     4:   end
     5: end
     6:
     7: a = 1
 =>  8: binding.irb
     9:
    10: b = 2
    11: c = 3

irb(main):001:0> break Foo#bar
Loaded debug-1.6.3
[3, 11] in test.rb
     3|     10
     4|   end
     5| end
     6|
     7| a = 1
=>   8| binding.irb
     9|
    10| b = 2
    11| c = 3
=>#0    <main> at test.rb:8
(rdbg:irb) break Foo
Unknown breakpoint format: Foo


### Breakpoint

* `b[reak]`
  * Show all breakpoints.
* `b[reak] <line>`
  * Set breakpoint on `<line>` at the current frame's file.
* `b[reak] <file>:<line>` or `<file> <line>`
  * Set breakpoint on `<file>:<line>`.
* `b[reak] <class>#<name>`
   * Set breakpoint on the method `<class>#<name>`.
* `b[reak] <expr>.<name>`
   * Set breakpoint on the method `<expr>.<name>`.
* `b[reak] ... if: <expr>`
  * break if `<expr>` is true at specified location.
* `b[reak] ... pre: <command>`
  * break and run `<command>` before stopping.
* `b[reak] ... do: <command>`
  * break and run `<command>`, and continue.
* `b[reak] ... path: <path>`
  * break if the path matches to `<path>`. `<path>` can be a regexp with `/regexp/`.
* `b[reak] if: <expr>`
  * break if: `<expr>` is true at any lines.
  * Note that this feature is super slow.


# and the program exits
```

Given that we now have the `debug` command to quickly enter debugging session and the effort for these commands are not small, I think we can prioritise other improvements before Ruby 3.2. WDYT?

---

Not directly related, but besides these commands a feature we can learn from  `debug` is its scriptable breakpoints. For example:

```
binding.irb(do: "ls") # run ls and continue the program
binding.irb(pre: "ls") # run ls and stop the program
binding.irb(pre: "ls ;; foo.bar?") # run ls, call foo.bar?, and stop the program
```

k0kubun:
> What subset of functionalities make sense in the context of irb? Just off the top of my head:

Same. It shouldn't be everything; you should use debug.gem directly if you want all. We need to hit a reasonable balance between convenience and maintainability. 20% of the features could cover 80% of the use cases, or something close enough to it.

My ambition is always about making it a Pry alternative that is available by default. In this case it's about a pry-byebug alternative. So, looking at https://github.com/deivid-rodriguez/pry-byebug, I'd say `break` (and `delete` on our side, which I've forgotten. I'll add that later)/`step`/`next`/`finish`/`continue` should be the first set of features we should support, which is what this PR does, and we could also have `backtrace`/`up`/`down`/`frame`.

> When adding a command, do we want to completely replicate it here, or just part of it?

I personally prefer being a bit conservative about a single-char shorthand at least in the first release because `n` and `c` are much more likely to conflict with local variable names. I'd like to support it only through command aliases for now.

As to supporting complicated features, as long as we can delegate that to debug.gem with a reasonable implementation cost, I think there's no reason to reject it. At least for the scope that could work by writing a simple `transform_args`, it should be fine.

> How can we help users discover them? debug has help command to help users learn all the commands' usages. But irb doesn't have it and we also don't have good documentation atm. This is already a problem, but it'll be more obvious after we introducing these commands.

This should probably be discussed in a separate issue. Personally, the way I discover debug.gem's commands and pry-byebug's commands was to look at their README. Nothing stops us from writing a table like https://github.com/ruby/debug#debug-command-on-the-debug-console at the README on https://github.com/ruby/irb, and it seems like enough discoverability, at least for me.

> Implementation-wise, I think we also need to figure how to write tests for them because the scenarios to consider will grow very fast. debug has a test framework that allows us to test the debugger in a separate process so we can verify its functionalities end-to-end ([example](https://github.com/ruby/debug/blob/master/test/console/break_test.rb#L142-L174)).

Nice :+1: Agreed.

> I also found that the break command isn't working:

This was kind of half-intended; I didn't think I supported everything yet, but at the same time I knew this would work perfectly as soon as I write a small `transform_args`. I should be able to fix it as soon as I work on it.

> Given that we now have the debug command to quickly enter debugging session and the effort for these commands are not small, I think we can prioritise other improvements before Ruby 3.2. WDYT?

So you're proposing to push it for another year (the whole point of this is to NOT write them in Gemfile; when you work on random OSS projects, you can only use what's installed by default) because we don't have isolated tests for it? If that's the only concern, I'll write a starter test code for each command. Again, with the presence of `transform_args`, delegating the implementation to debug.gem to fix any issues that could be fixed by that is a really trivial concern.

As soon as I started using `debug`, it turned out to be really frustrating that the features I added in this PR didn't exist. I kept using `next` before starting `debug`. Because debug.gem copied some IRB coloring stuff and tried to look similar to it, it's pretty hard to recognize whether it's IRB or debug.gem. On one hand, it's useful; obviously that's what this PR is taking advantage of. But this UI similarity instinctively trigger you to immediately use debug.gem's commands on `binding.irb`. IMO, I don't want to see Ruby 3.2 only with `debug` command because of this experience.

st0012:
> because we don't have isolated tests for it?

Not just that. My concern is that with other supporting work (e.g. command documentation), issues like https://github.com/ruby/irb/issues/445, and we're pretty short on time (1 month away from 3.2 release), I'm not sure if we can ship these commands in good shape and test them well.

But I guess we can give a try because it does look like most of the problems may be on argument handling.

However, I do feel strongly that we should make sure command documentation can be shipped along with this. So I just opened https://github.com/ruby/irb/issues/450

k0kubun:
> My concern is that with other supporting work (e.g. command documentation), issues like https://github.com/ruby/irb/issues/445

Just to be clear, I'm not asking you to implement those non-`debug` commands. I do understand that reviewing PRs is also a time-consuming task, but I hope it doesn't take 1 month to review this PR and follow-up bug fixes on it.

> But I guess we can give a try because it does look like most of the problems may be on argument handling.

Thanks :+1: I'll give it a try, and give it up if it turns out to be too hard to write tests or fix the argument handling.

> However, I do feel strongly that we should make sure command documentation can be shipped along with this. So I just opened https://github.com/ruby/irb/issues/450

:+1:

lib/irb/cmd/debug.rb
```diff
@@ -25,11 +25,16 @@ def execute(*args)
...
+        if debug_command
+          command = ['irb', nil, debug_command]
+        end
+
         # To make debugger commands like `next` or `continue` work without asking
         # the user to quit IRB after that, we need to exit IRB first and then hit
         # a TracePoint on #debug_break.
         file, lineno = IRB::Irb.instance_method(:debug_break).source_location
-        DEBUGGER__::SESSION.add_line_breakpoint(file, lineno + 1, oneshot: true, hook_call: false)
+        DEBUGGER__::SESSION.add_line_breakpoint(file, lineno + 1, oneshot: true, hook_call: false, command: command)
```
st0012, line 37: I'm not sure if inserting a new breakpoint with command is the best way to implement these commands. It may mess up with level computation (not sure).

I think ideally we should make `irb` an UI of debugger. That means the command can be [passed directly into debugger's session](https://github.com/ruby/debug/blob/d7f181c322beadaa456a4db7af5298526dd5b1e1/lib/debug/session.rb#L404-L409).

k0kubun: > I'm not sure if inserting a new breakpoint with command is the best way to implement these commands. It may mess up with level computation (not sure).

I'm not sure what's different in the level computation between `debug` command and non-`debug` commands. My understanding is that it's primarily handled by the monkey-patched `capture_frames` on the oneshot breakpoint issued by IRB. If it's broken for non-`debug` commands, I'm not sure if it would work for `debug` either. We move to the debugger with the same implementation, and we start executing commands from the session we entered in the same way by a human input or this `command` option. 

> I think ideally we should make irb an UI of debugger. 

I agree. And this is what I meant in https://github.com/ruby/irb/issues/425#issuecomment-1318264770 to give up for the Ruby 3.2 release. My idea is that, in Ruby 3.2, it's okay to make these commands a superset / thin wrapper of `debug`, and hopefully make IRB a UI of debugger in Ruby 3.3 or so if we want.

If there's something that doesn't work with this implementation, I'll probably just give it up instead of working on the perfect UI integration. But I assume it would mean that `debug` command is also broken on that case.

st0012: > hopefully make IRB a UI of debugger in Ruby 3.3 or so for these commands if we want.

I don't mean adding an irb UI to the debug gem. I think we can customise a UI class and then [initialise the debugger with it](https://github.com/ruby/debug/blob/d7f181c322beadaa456a4db7af5298526dd5b1e1/lib/debug/session.rb#L2116-L2123). Instead of passing a `UI_LocalConsole.new`, we pass `UI_IRBConsole.new`, for example.

k0kubun: Yes, I knew about the `UI_*` classes, and we were talking about the same thing.

k0kubun: I don't stop you from making that improvement if you want, but my point is that I believe the current design should work for the cases where typing the same command on the debugger mode started by the `debug` command you added works, and it saves us some time if you're concerned about the 1-month schedule. (And again, I'll give up if I find some unexpected blocker.)

st0012:
After using the `next` command, the `debug` seems to take over the session completely:

```
From: test.rb @ line 3 :

    1: class Foo
    2:   def bar
 => 3:     binding.irb
    4:     b = 2
    5:     c = 3
    6:     d = 4
    7:   end
    8: end

irb(#<Foo:0x0000000106fec0e8>):001:0> next
Loaded debug-1.6.3
[1, 10] in test.rb
     1| class Foo
     2|   def bar
=>   3|     binding.irb
     4|     b = 2
     5|     c = 3
     6|     d = 4
     7|   end
     8| end
     9|
    10| a = 1
=>#0    Foo#bar at test.rb:3
  #1    <main> at test.rb:12
(rdbg:irb) next
[1, 10] in test.rb
     1| class Foo
     2|   def bar
     3|     binding.irb
=>   4|     b = 2
     5|     c = 3
     6|     d = 4
     7|   end
     8| end
     9|
    10| a = 1
=>#0    Foo#bar at test.rb:4
  #1    <main> at test.rb:12
(rdbg) next    # command
[1, 10] in test.rb
     1| class Foo
     2|   def bar
     3|     binding.irb
     4|     b = 2
=>   5|     c = 3
     6|     d = 4
     7|   end
     8| end
     9|
    10| a = 1
=>#0    Foo#bar at test.rb:5
  #1    <main> at test.rb:12
(rdbg) ls Object -g to    # outline command
eval error: (rdbg)/test.rb:1: syntax error, unexpected local variable or method, expecting `do' or '{' or '('
Object -g to
          ^~
(rdbg)
```

lib/irb/cmd/debug.rb
st0012: > Yes, I knew about the UI_* classes, and we were talking about the same thing

I see, sorry. I didn't know you expect it to take much more effort. What obstacles do you currently see with this approach? Just curious.

st0012:
^ Sorry that I probably shouldn't post issues command-by-command. It's pretty late here and I didn't think it though 😅 

I'll test other commands and post what I found together.

k0kubun:
> After using the next command, the debug seems to take over the session completely

Yes, it's the intended behavior. It's meant to be a superset of `debug` command. If you can't stay in the debug session just after a single `next` command execution, it wouldn't be useful. Honestly I don't get why you left a comment about it.

lib/irb/cmd/debug.rb
k0kubun: I think the main problem for me is that I still don't have much understanding of the debug.gem implementation; I only spent a couple of nights playing with it.

I already know enough to write up this PR within a few minutes, but I'm sure it would take more than a few minutes to learn how the UI is integrated and write an original UI class for IRB. It's probably easy for you, but I assume I'm the only person contributing to this feature since you seem to want to spend more time on auto-completion and documentation. And the ratio of my OSS time that I can spend for IRB is not large enough to accomplish the other implementation. But I also believe that the quality of the current design is as good as `debug` command, and we've effectively agreed that `debug` command is a Ruby 3.2-worthy quality.

k0kubun:
Let me get this straight. So far we've discussed two ways to introduce debugger commands to IRB:

1. Change the session to the debugger like `debug` command. Then run the command on it that was originally passed to IRB.
2. Implement a `UI_*` class in IRB and use that to start the debugger session. You could pass rich commands to the debugger instead of a String. Unlike `debug` command, the debugger session doesn't take over and you can keep talking to the debugger while staying in an IRB session.

And I'm proposing to implement only option 1 in Ruby 3.2, and potentially improving it to option 2 in Ruby 3.3 or later. I _already_ spent time experimenting with option 2 and then left this comment https://github.com/ruby/irb/issues/425#issuecomment-1318264770. I already decided to give up option 2 in Ruby 3.2. However, after that effort and playing with your `debug` implementation, I noticed we could implement option 1 with a reasonable cost. So here we're.

lib/irb/cmd/debug.rb
k0kubun: I revised the implementation and I'd say it's working really well now. Everything is working as intended, so I don't think we need to work on the UI thing.

k0kubun:
I addressed all your concerns. This is working really well, even more than I initially expected. I don't see any reason to avoid shipping this in Ruby 3.2 at this point. Please have another look :)

k0kubun:
> I think I get it now: these commands are to save typing debug, <cmd> to just <cmd>. And then the user would stay in the debug session?
> (When I saw "Seamlessly integrate", my immediate expectation was pry-byebug experience and I stuck with it sweat_smile)

Yeah, sorry about the confusion. 

> If my current understanding is correct, I think we can also add:

done https://github.com/ruby/irb/pull/449/commits/c347b4e3a209b53eaf2988f7ff23d0be19dffd57

k0kubun:
@st0012 I also fixed a conflict after merging https://github.com/ruby/irb/pull/453. Do you have any final feedback?

st0012 (commented):
I think the implementation looks good 👍 

But I'm concerned about testing the commands with [`yamatanooroti`](https://github.com/aycabta/yamatanooroti). It and its dependency [vterm](https://github.com/aycabta/vterm-gem) are both not under `ruby`'s control, so if they have problems going forward, we won't be able to fix them. 

I think it's fine to keep and maintain the current tests but ideally we shouldn't add new tests with them, especially when these tests seem to be less-demanding on terminal setup. WDYT?

Also, can you help me setup `vterm` on Mac (with `libvterm` installed)? I got these errors when installing it:

```
current directory: /Users/hung-wulo/.gem/ruby/3.1.2/gems/vterm-0.0.5/ext/vterm
/opt/rubies/3.1.2/bin/ruby -I /opt/rubies/3.1.2/lib/ruby/3.1.0 -r ./siteconf20221120-15782-673e37.rb extconf.rb
checking for vterm.h... no
checking for vterm_new() in -lvterm... no
creating Makefile

current directory: /Users/hung-wulo/.gem/ruby/3.1.2/gems/vterm-0.0.5/ext/vterm
make DESTDIR\= clean

current directory: /Users/hung-wulo/.gem/ruby/3.1.2/gems/vterm-0.0.5/ext/vterm
make DESTDIR\=
compiling vterm.c
vterm.c:10:10: fatal error: 'vterm.h' file not found
#include "vterm.h"
         ^~~~~~~~~
1 error generated.
make: *** [vterm.o] Error 1

make failed, exit code 2
```

Issue #450, opened 2022-11-18
Proposal: Rename the current `help` command so it can be used to display command documentation


st0012:
This proposal includes 3 changes:

1. Add documentation API to command classes. For example, if we want to have [pry-style APIs](https://github.com/pry/pry/blob/master/lib/pry/commands/ls.rb#L29-L50):

    ```rb
    module ExtendCommand
      class Debug < Nop
        description "Load `debug` gem and starts a debugging session"

        banner <<~MSG
          Usage: debug

          This command loads the `debug` gem and start a debugging session at the current location.
          ....
        MSG
      end
    end
    ```
2. Rename the current `help` command (showing API documents) to something else, like `doc`
    - In `pry`, `byebug` and `debug`, the `help` command are all used for showing command documents. So I think it's better to align with them.

3. Support the new `help` command to list command documents:
   - `help` - list all available commands
       - If we follow what `pry` or `byebug` do, we only display commands' descriptions.
       - If we follow `debug`, we'll display full documentation for all commands. So in this case we won't need the `description` attribute.
       - Question: this means we'll be loading command classes by default. Is this a good reason to do that?
   - `help <cmd>` - display individual commands' full usage (banner).

Since this will be a breaking change (renaming a command), we'll need to bump the version that includes this feature to `2.0`.

k0kubun:
While the two of us are allowed to add/modify IRB commands by just reviewing each other's PR, renaming/replacing an existing command would probably not be considered the same kind of change. It'd be nice to file a ticket on Redmine if you want to land this feature as `help`.

st0012:
I see. I'll first add something like `cmd-help` or `commands` before proposing the renaming.
But do you agree to eagerly load commands so we can store their documents and descriptions as attributes? The current lazy-load design was introduced [since 0.9](https://github.com/ruby/irb/commit/6806669d178f90f38b99c144bdfee06bdb93a229#diff-707785dd86b785fc2279e00b12e4f4c316926e6e633de2bdd374d39b91f28ec5) but I didn't find the reason behind it.

k0kubun:
I think there was a moment when starting irb CLI was very slow, which could be related to it. We even have a benchmark for it under ruby/ruby/benchmark (irb_exec or something like that), so you may try it with and without lazy load.

Though, I don't get why you can't maintain lazy load. You'd only need to lazily load everything when you need to show/generate docs.

st0012:
[...]

PR #451, opened 2022-11-19
Document a full list of commands

k0kubun:
Partly address this problem https://github.com/ruby/irb/pull/449#issuecomment-1320589869:

> How can we help users discover them? debug has help command to help users learn all the commands' usages. But irb doesn't have it and we also don't have good documentation atm. 

With the current way of README format, I haven't noticed IRB had a long document at https://docs.ruby-lang.org/en/master/IRB.html until today. I find it more discoverable to just write them down in README like debug.gem https://github.com/ruby/debug#debug-command-on-the-debug-console.

So, while I added the list to the official documentation for its readers, I also added it to the README as well. This should be useful even after https://github.com/ruby/irb/issues/450 since you'll somehow need to discover `help` command itself. Having that in README could be a good help.

Note that this list is based on [this comment](https://github.com/ruby/irb/blob/ff8b6fe4d49b746175961fd60675112e5c250a58/lib/irb/extend-command.rb#L170-L187). I think we could document this even better, e.g. have categories or show commonly-used commands first, but that shouldn't stop this PR from being merged. I'd like to focus on having something first, and then improve it in other PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,48 @@ irb(main):006:1> end
 
 The Readline extension module can be used with irb. Use of Readline is default if it's installed.
 
+## Commands
+
+The following commands are available on IRB.
+
+* `cwws`
+  * Show the current workspace.
+* `cb`, `cws`, `chws`
+  * Change the current workspace to an object.
+* `bindings`, `workspaces`
+  * Show workspaces.
+* `pushb`, `pushws`
+  * Push an object to the workspace stack.
+* `popb`, `popws`
+  * Pop a workspace from the workspace stack.
+* `load`
+  * Load a Ruby file.
+* `require`
+  * Require a Ruby file.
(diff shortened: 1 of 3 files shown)


st0012 (approved):
Nice 👍 

I think after #450 we can make a readme template like [`debug`'s](https://github.com/ruby/debug/blob/master/misc/README.md.erb#L510) and just generate them automatically.

README.md
```diff
@@ -40,6 +40,48 @@ irb(main):006:1> end
 
 The Readline extension module can be used with irb. Use of Readline is default if it's installed.
 
+## Commands
```
st0012, line 43: Actually, we missed `debug` 😄

k0kubun: :+1: fixed

PR #453, opened 2022-11-19
Add edit command

st0012:
## Name

I originally wanted to name it `open`. But since `ruby/debug` has an `edit` command (the inspiration of this feature), I think it's better to match it.

## Usages

**Editor is determined by `ENV["EDITOR"]`**

- `edit` - opens the current context's file in the editor
- `edit foo.rb` - opens `foo.rb` in the editor if it exists

It also supports `show_source` compatible arguments, like

- `edit Foo`
- `edit Foo#bar`
- `edit Foo.baz`

In these cases, it'll use `show_source` to resolve the argument's path and open them in the editor.

### Exceptions

- If `irb` can't find a file with the argument, nor can `ShowSource#find_source` resolve its source, it prints a message and stops.
- If `ENV["EDITOR"]` is not defined, the command also prints a message and stops.

### DIFF
--- a/lib/irb/cmd/edit.rb
+++ b/lib/irb/cmd/edit.rb
@@ -33,11 +33,10 @@ module IRB
 
         if !File.exist?(path)
           require_relative "show_source"
-          show_source_cmd = ShowSource.new(@irb_context)
 
           source =
             begin
-              show_source_cmd.send(:find_source, path)
+              ShowSource.find_source(path, @irb_context)
             rescue NameError
               # if user enters a path that doesn't exist, it'll cause NameError when passed here because find_source would try to evaluate it as well
               # in this case, we should just ignore the error
--- a/lib/irb/cmd/edit.rb
+++ b/lib/irb/cmd/edit.rb
@@ -1,0 +1,66 @@
+require 'shellwords'
+require_relative "nop"
+
+module IRB
+  # :stopdoc:
+
+  module ExtendCommand
+    class Edit < Nop
+      class << self
+        def transform_args(args)
+          # Return a string literal as is for backward compatibility
+          if args.nil? || args.empty? || string_literal?(args)
+            args
+          else # Otherwise, consider the input as a String for convenience
+            args.strip.dump
+          end
+        end
+
+        private
+
+        def string_literal?(args)
+          sexp = Ripper.sexp(args)
+          sexp && sexp.size == 2 && sexp.last&.first&.first == :string_literal
+        end
+      end
+
+      def execute(*args)
+        path = args.first
+
+        if path.nil? && (irb_path = @irb_context.irb_path)
+          path = irb_path
+        end
+
+        if !File.exist?(path)
+          require_relative "show_source"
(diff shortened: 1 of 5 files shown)


lib/irb/cmd/edit.rb
```diff
@@ -0,0 +1,66 @@
...
+          path = irb_path
+        end
+
+        if !File.exist?(path)
+          require_relative "show_source"
+          show_source_cmd = ShowSource.new(@irb_context)
+
+          source =
+            begin
+              show_source_cmd.send(:find_source, path)
```
k0kubun, line 39: Can we change `find_source` to a public singleton method that also takes `irb_context` to clarify that it's now a method shared with other classes? It's hard to notice that the method is shared if it's kept private.

k0kubun (approved):
I merged https://github.com/ruby/irb/pull/451. Let's add `edit` to it too :+1:

lib/irb/cmd/edit.rb
```diff
@@ -0,0 +1,66 @@
...
+            puts "Can not find file: #{path}"
+            return
+          end
+        end
+
+        if editor = ENV['EDITOR']
+          puts "command: '#{editor}'"
+          puts "   path: #{path}"
+          require 'shellwords'
+          system(*Shellwords.split(editor), path)
```
k0kubun, line 57: It's probably safer to require "shellwords" at the top of this file.

lib/irb/cmd/edit.rb
st0012: Updated

lib/irb/cmd/edit.rb
st0012: Updated 👍

lib/irb/cmd/show_source.rb
```diff
@@ -19,6 +19,48 @@ def transform_args(args)
...
+          when /\A((?<receiver>.+)(\.|::))?(?<method>[^ :.]+)\z/ # method, receiver.method, receiver::method
+            receiver = eval(Regexp.last_match[:receiver] || 'self', irb_context.workspace.binding)
+            method = Regexp.last_match[:method]
+            file, line = receiver.method(method).source_location if receiver.respond_to?(method)
+          end
+          if file && line
+            Source.new(file: file, first_line: line, last_line: find_end(file, line))
+          end
+        end
+
```
k0kubun, line 43:
```suggestion

        private

```

lib/irb/cmd/show_source.rb
```diff
@@ -19,6 +19,50 @@ def transform_args(args)
...
+            continue = lex.process_continue(prev_tokens)
+            code_block_open = lex.check_code_block(code, prev_tokens)
+            if !continue && !code_block_open
+              return first_line + lnum
+            end
+          end
+          first_line
+        end
+
         private
```
k0kubun, line 66:
```suggestion
```

Remove duplicated private

k0kubun (approved):
:tada:

PR #454, opened 2022-11-21
Try using skip_path

k0kubun:


k0kubun:
Asked by ko1 to try this, but not working.

st0012:
Yeah because `skip_path` is invoked much later. Another solution could be letting `capture_frames` take multiple [skip_path_prefix](https://github.com/ruby/debug/blob/master/ext/debug/debug.c#L87) instead of just one.

PR #456, opened 2022-11-21
Copy and use debug.gem's test framework

k0kubun:
The idea proposed at https://github.com/ruby/irb/pull/449#pullrequestreview-1187255149


k0kubun:
addressed by https://github.com/ruby/irb/pull/464

PR #458, opened 2022-11-26
Fix the debug.gem force-loader for Ruby 3.2

k0kubun:
Follows up https://github.com/ruby/irb/pull/448.

Ruby 3.1.3 has a debug.gem installation like:

```
$ find /opt/rubies/3.1.3 |grep "debug\.so"
/opt/rubies/3.1.3/lib/ruby/gems/3.1.0/extensions/x86_64-linux/3.1.0-static/debug-1.6.3/debug/debug.so
/opt/rubies/3.1.3/lib/ruby/gems/3.1.0/gems/debug-1.6.3/ext/debug/debug.so
/opt/rubies/3.1.3/lib/ruby/gems/3.1.0/gems/debug-1.6.3/lib/debug/debug.so
```

On the other hand, Ruby 3.2.0-preview3 has only:

```
$ find /opt/rubies/3.2.0-preview3 |grep "debug\.so"
/opt/rubies/3.2.0-preview3/lib/ruby/gems/3.2.0+3/extensions/x86_64-linux/3.2.0+3-static/debug-1.6.3/debug/debug.so
```

We'd need to support the `extensions` path for Ruby 3.2.

### DIFF
--- a/lib/irb/cmd/debug.rb
+++ b/lib/irb/cmd/debug.rb
@@ -82,14 +82,22 @@ module IRB
       # it's a bundled gem. This method tries to activate and load that.
       def load_bundled_debug_gem
         # Discover latest debug.gem under GEM_PATH
-        debug_gem = Gem.paths.path.map { |path| Dir.glob("#{path}/gems/debug-*") }.flatten.select do |path|
+        debug_gem = Gem.paths.path.flat_map { |path| Dir.glob("#{path}/gems/debug-*") }.select do |path|
           File.basename(path).match?(/\Adebug-\d+\.\d+\.\d+\z/)
         end.sort_by do |path|
           Gem::Version.new(File.basename(path).delete_prefix('debug-'))
         end.last
         return false unless debug_gem
 
+        # Discover debug/debug.so under extensions for Ruby 3.2+
+        debug_so = Gem.paths.path.flat_map do |path|
+          Dir.glob("#{path}/extensions/**/#{File.basename(debug_gem)}/debug/debug.so")
+        end.first
+
         # Attempt to forcibly load the bundled gem
+        if debug_so
+          $LOAD_PATH << debug_so.delete_suffix('/debug/debug.so')
+        end
         $LOAD_PATH << "#{debug_gem}/lib"
         begin
           require "debug/session"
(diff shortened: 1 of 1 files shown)


lib/irb/cmd/debug.rb
```diff
@@ -82,14 +82,22 @@ def DEBUGGER__.capture_frames(*args)
...
-          File.basename(path).match?(/\Adebug-\d+\.\d+\.\d+\z/)
+        debug_gem = Gem.paths.path.flat_map { |path| Dir.glob("#{path}/gems/debug-*") }.select do |path|
+          File.basename(path).match?(/\Adebug-\d+\.\d+\.\d+(\w+)?\z/)
         end.sort_by do |path|
           Gem::Version.new(File.basename(path).delete_prefix('debug-'))
         end.last
         return false unless debug_gem
 
+        # Discover debug/debug.so under extensions for Ruby 3.2+
+        debug_so = Gem.paths.path.flat_map do |path|
```
st0012, line 93: Do we need to avoid triggering this path for Ruby 3.2-?

lib/irb/cmd/debug.rb
```diff
@@ -82,14 +82,22 @@ def DEBUGGER__.capture_frames(*args)
       # it's a bundled gem. This method tries to activate and load that.
       def load_bundled_debug_gem
         # Discover latest debug.gem under GEM_PATH
-        debug_gem = Gem.paths.path.map { |path| Dir.glob("#{path}/gems/debug-*") }.flatten.select do |path|
-          File.basename(path).match?(/\Adebug-\d+\.\d+\.\d+\z/)
+        debug_gem = Gem.paths.path.flat_map { |path| Dir.glob("#{path}/gems/debug-*") }.select do |path|
```
st0012, line 85: Nitpick: It looks like `Gem.paths.path` can be stored in a local?

lib/irb/cmd/debug.rb
k0kubun: No. It seems like the original path, and it's intentional that it uses that path for older versions as well.

lib/irb/cmd/debug.rb
k0kubun: What do you mean? Are you talking about assigning it to a local variable? `Gem.paths` is memoized and `.path` is just an attr_reader. Do you still want to do that?

st0012: Ah it's memoized. There's no need then 👍

st0012 (approved):

PR #459, opened 2022-11-26
Activate yamatanooroti tests on CI

st0012:
The env is also needed for running the tests, otherwise it won't load the `vterm` gem and won't do anything.

### DIFF
--- a/test/irb/yamatanooroti/test_rendering.rb
+++ b/test/irb/yamatanooroti/test_rendering.rb
@@ -297,9 +297,9 @@ begin
       write("continue\n")
       write("delete 0\n")
       close
-      assert_include_screen(<<~EOC)
+      assert_include_screen(<<~EOC.strip)
         (rdbg:irb) delete 0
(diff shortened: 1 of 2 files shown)


test/irb/yamatanooroti/test_rendering.rb
```diff
@@ -297,9 +297,9 @@ def test_delete
       write("continue\n")
       write("delete 0\n")
       close
-      assert_include_screen(<<~EOC)
+      assert_include_screen(<<~EOC.strip)
         (rdbg:irb) delete 0
-        deleted: #0  BP - Line  #{@ruby_file}:5 (line)
+        deleted: #0  BP - Line
```
st0012, line 302: We need to loose this expectation because `debug` can show additional `(pending)` info based on Ruby versions (depending on iseq support).
Since we just want to test if `debug` receives the command and argument correctly, matching only this part of the message should be good enough.

k0kubun (approved):
:man_facepalming:

PR #464, opened 2022-12-01
Test debug commands without yamatanooroti

st0012:
As discussed in #449, it's better to not rely on `yamatanooroti` for new tests. So in this PR, I added debug command tests using approach similar to `debug`'s (with `pty`) and removed the ones added in #449.

Note: There seems to be an issue in `debug 1.6.3` that causes `delete` to fail in Ruby 2.6~3.0 ([example](https://github.com/ruby/irb/actions/runs/3597199737/jobs/6058705262)), but not 3.1+ ([example](https://github.com/ruby/irb/actions/runs/3597199737/jobs/6058705029)). Given that it seems addressed in the latest `debug`, we should test with this version until it's officially released.


k0kubun:
> which adds 2 lines of boilerplate code per-test.

Can we possibly set up an irbrc on setup and remove it on teardown?

st0012:
FWIW, it can also be used by users to make production setup easier: https://github.com/ruby/irb/issues/351#issuecomment-1332036964

Instead of:

> heroku run "rails console -- --nomultiline" -a my-app-name

User can set `USE_SINGLELINE=true` in Heroku's envs panel and just use `heroku run "rails c"`.

But it probably worth a separate discussion. I'll try using `rc` file for now.

k0kubun (approved):
:heart:

PR #466, opened 2022-12-02
Disable debug cmd tests based on project structure instead of env

st0012:
It's hard to find an env var that's universally set in all Ruby CI environments but not in local. Checking gemspec seems to be a better way as `syntax_suggest` already uses it for a while.

https://github.com/ruby/syntax_suggest/blob/d8f1bca297a392b80b5d167e2be07cf6ac4feae3/spec/spec_helper.rb#L47

### DIFF
--- a/test/irb/test_debug_cmd.rb
+++ b/test/irb/test_debug_cmd.rb
@@ -16,13 +16,16 @@ module TestIRB
     }
 
     def setup
-      if ENV["GITHUB_ACTION_REPOSITORY"] != "ruby/irb"
-        omit "This test works only on ruby/irb CI"
+      if ruby_core?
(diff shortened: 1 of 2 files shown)


test/irb/test_debug_cmd.rb
```diff
@@ -16,13 +16,16 @@ class DebugCommandTestCase < TestCase
     }
 
     def setup
-      if ENV["GITHUB_ACTION_REPOSITORY"] != "ruby/irb"
```
st0012, line 19: Using this env means it's disabled locally as well. But AFIAK, adding `ENV["CI"]` to the condition may not work because not all Ruby CI environments have it set.

k0kubun (approved):

PR #469, opened 2022-12-03
Allow disabling autocompletion with `IRB_USE_AUTOCOMPLETE=false` 

st0012:
Currently, the only 2 ways to disable autocompletion are:

1. Create `.irbrc` and set `IRB.conf[:USE_AUTOCOMPLETE] = false`
2. Add the `--noautocomplete` flag when using the `irb` executable

Both of them are less convenient than setting a env var and are lesser known to devs.

And given the number of problems the autocompletion has (see #445), I think we should allow disabling it with a simple `IRB_USE_AUTOCOMPLETE=false`.

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -45,7 +45,7 @@ module IRB # :nodoc:
 
     @CONF[:USE_SINGLELINE] = false unless defined?(ReadlineInputMethod)
     @CONF[:USE_COLORIZE] = (nc = ENV['NO_COLOR']).nil? || nc.empty?
-    @CONF[:USE_AUTOCOMPLETE] = true
+    @CONF[:USE_AUTOCOMPLETE] = ENV.fetch("IRB_USE_AUTOCOMPLETE", "true") != "false"
     @CONF[:INSPECT_MODE] = true
     @CONF[:USE_TRACER] = false
     @CONF[:USE_LOADER] = false
--- a/README.md
+++ b/README.md
@@ -93,6 +93,15 @@ The following commands are available on IRB.
 * `break`, `delete`, `next`, `step`, `continue`, `finish`, `backtrace`, `info`, `catch`
   * Start the debugger of debug.gem and run the command on it.
 
+## Configuration
+
+### Environment Variables
+
+- `NO_COLOR`: Assigning a value to it disables IRB's colorization.
+- `IRB_USE_AUTOCOMPLETE`: Setting it to `false` disables IRB's autocompletion.
+- `EDITOR`: Its value would be used to open files by the `edit` command.
+- `IRBRC`: The file specified would be evaluated as IRB's rc-file.
+
 ## Documentation
 
 https://docs.ruby-lang.org/en/master/IRB.html
--- a/test/irb/test_init.rb
+++ b/test/irb/test_init.rb
@@ -96,6 +96,30 @@ module TestIRB
       IRB.conf[:USE_COLORIZE] = orig_use_colorize
     end
 
+    def test_use_autocomplete_environment_variable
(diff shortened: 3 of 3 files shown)


st0012:
cc @k0kubun

k0kubun:
> Both of them are less convenient than setting a env var and are 

It depends. `~/.irbrc` is scoped to IRB, and `--noautocomplete` is scoped to each IRB invocation. Having `USE_AUTOCOMPLETE` in your shell could impact any other non-IRB program. Global variables are "convenient" but it doesn't necessarily mean it improves maintainability.

I like that should chose a name consistent with irbrc, but when it comes to environment variables, I would choose a name that includes `IRB` to avoid possible namespace conflicts.

> lesser known to devs.

How will it be easier to know of this environment variable than `.irbrc` or `--noautocomplete`?

> And given the number of problems the autocompletion has (see https://github.com/ruby/irb/issues/445), I think we should allow disabling it with a simple USE_AUTOCOMPLETE=false.

If it's that problematic, we might want to consider disabling it by default until all the issues are addressed. The most annoying part of auto-completion for me is the fact that the prompt position randomly moves depending on the completion size when you're at the bottom of the window, but currently nobody even discusses it maybe because there are more annoying problems for other people.

I appreciate that reline-powered IRB supports multi-line editing and can be used like an editor, but I still use one-liner inputs on IRB most of the time. It's never easier to edit a multi-line long script on reline than your own actual editor. If you mostly use a one-liner, it feels more like a normal REPL than an editor. It's normal to have auto-completion on an editor, but not on a REPL. While it might be useful for some people to have this capability, I don't think this needs to be enabled by default, especially when it's a factor that annoys IRB users.

st0012:
> I like that you chose a name consistent with irbrc, but when it comes to environment variables, I would choose a name that includes IRB to avoid possible namespace conflicts.

We can use the same prefix convention `debug` uses, which would be `RUBY_IRB_USE_AUTOCOMPLETE`.

> How will it be easier to know of this environment variable than .irbrc or --noautocomplete?

I think I should phrase it differently: in my experience, env var is a common way to control features like this (such as `NO_COLOR`, which IRB also supports). So if given no guidance on how to configure the feature, I'd look for related env vars first.

> If it's that problematic, we might want to consider disabling it by default until all the issues are addressed. 

I thought about it before. But based on my polling result in #445, we have around 28% of users satisfied with it as it is today. And only 17% of them dislike it enough to disable it. So if we disable it now, it'd be a breaking change for the 28% of users, which is a big number.

Therefore, I think we should keep it enabled, but also make it easier to disable. That's why I want to propose this env var.

st0012:
`IRB_USE_AUTOCOMPLETE` sounds good, updated 👍

k0kubun (approved):

lib/irb/init.rb
```diff
@@ -45,7 +45,7 @@ def IRB.init_config(ap_path)
 
     @CONF[:USE_SINGLELINE] = false unless defined?(ReadlineInputMethod)
     @CONF[:USE_COLORIZE] = (nc = ENV['NO_COLOR']).nil? || nc.empty?
-    @CONF[:USE_AUTOCOMPLETE] = true
+    @CONF[:USE_AUTOCOMPLETE] = !(ENV.fetch("IRB_USE_AUTOCOMPLETE", "true") == "false")
```
k0kubun, line 48: One last very trivial comment since I got confused when I was about to push the merge button, I wonder why it's not `ENV.fetch("IRB_USE_AUTOCOMPLETE", "true") != "false"` or `(ENV.fetch("IRB_USE_AUTOCOMPLETE", "true") != "false")`. You generally like `!(a == b)` than `a != b`?

st0012: No it's not any special preference. I've updated 👍

PR #471, opened 2022-12-05
Kill PTY process after test is finished

st0012:
The killing/waiting logic is borrowed from ruby/debug:

https://github.com/ruby/debug/blob/ec5ae5aebd61a99dc84028d8dffa8e7e165c1ec6/test/support/test_case.rb#L107-L136


st0012:
cc @k0kubun

k0kubun (approved):

PR #472, opened 2022-12-07
Lazily load the multi-irb extension

st0012:
We now have plan to implement a command that prints all commands'
information, which will need to load all commands files without actually
running them.

But because the `multi-irb` extension patches IRB's top-level methods,
loading it would cause unintentional side-effects.

So this commit moves related requires into command execution to avoid the problem.

### DIFF
--- a/lib/irb/cmd/subirb.rb
+++ b/lib/irb/cmd/subirb.rb
@@ -10,31 +10,42 @@
 #
 
 require_relative "nop"
-require_relative "../ext/multi-irb"
 
 module IRB
   # :stopdoc:
 
   module ExtendCommand
-    class IrbCommand < Nop
+    class MultiIRBCommand < Nop
+      def initialize(conf)
+        super
+        extend_irb_context
+      end
+
+      def extend_irb_context
+        # this extension patches IRB context like IRB.CurrentContext
+        require_relative "../ext/multi-irb"
+      end
+    end
+
+    class IrbCommand < MultiIRBCommand
       def execute(*obj)
         IRB.irb(nil, *obj)
       end
     end
 
-    class Jobs < Nop
+    class Jobs < MultiIRBCommand
       def execute
         IRB.JobManager
       end
     end
 
-    class Foreground < Nop
+    class Foreground < MultiIRBCommand
       def execute(key)
         IRB.JobManager.switch(key)
       end
     end
 
-    class Kill < Nop
+    class Kill < MultiIRBCommand
       def execute(*keys)
         IRB.JobManager.kill(*keys)
       end
--- a/lib/irb/cmd/subirb.rb
+++ b/lib/irb/cmd/subirb.rb
@@ -21,6 +21,8 @@ module IRB
         extend_irb_context
       end
 
+      private
+
       def extend_irb_context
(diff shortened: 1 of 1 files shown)


lib/irb/cmd/subirb.rb
```diff
@@ -10,32 +10,35 @@
...
     class Foreground < Nop
       def execute(key)
+        require_relative "../ext/multi-irb"
         IRB.JobManager.switch(key)
       end
     end
 
     class Kill < Nop
       def execute(*keys)
+        require_relative "../ext/multi-irb"
```
k0kubun, line 41: The overhead of `require_relative` would be hopefully negligible, but it seems ugly that you need to repeat this for every command in this file, especially when this file itself is lazily loaded for non-`show_cmds` stuff.

Is it really important to avoid loading `../ext/multi-irb` on `show_cmds`? The implementation complexity by this seems to outweigh the benefit of not loading them if it's just about the initial loading speed and there's no side effect besides method definitions.

st0012: No it's not for speed. It's because these extensions, like `multi-irb`, [overrides some top-level methods](https://github.com/ruby/irb/blob/master/lib/irb/ext/multi-irb.rb#L180-L262).
So if `show_cmds` loads it, the users would have IRB's core components/top-level methods patched without noticing it.

k0kubun: I wouldn't call methods on `IRB` as top-level methods, but I agree that it's a surprising side effect.

The danger is hard to notice from reading the current code. Maybe worth defining a helper method for doing `require_relative "../ext/multi-irb"`, indicating its side effect from the method name?

lib/irb/cmd/chws.rb
```diff
@@ -26,6 +25,7 @@ def execute(*obj)
 
     class ChangeWorkspace < Nop
       def execute(*obj)
+        require_relative "../ext/change-ws"
```
k0kubun, line 28: Unlike `"../ext/multi-irb"`, I'm not sure if we need this. This doesn't override anything, and it impacts only IRB implementation, not the actual context users would eval expressions on IIUC.

lib/irb/cmd/pushws.rb
```diff
@@ -11,27 +11,29 @@
...
 require_relative "nop"
-require_relative "../ext/workspaces"
 
 module IRB
   # :stopdoc:
 
   module ExtendCommand
     class Workspaces < Nop
       def execute(*obj)
+        require_relative "../ext/workspaces"
```
k0kubun, line 21: ditto. This doesn't seem too dangerous.

lib/irb/cmd/subirb.rb
```diff
@@ -10,31 +10,42 @@
...
 
-    class Jobs < Nop
+    class Jobs < MultiIRBCommand
       def execute
         IRB.JobManager
       end
     end
 
-    class Foreground < Nop
+    class Foreground < MultiIRBCommand
```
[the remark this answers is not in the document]
st0012: Sorry it's just bad error handling 🤦‍♂️

PR #473, opened 2022-12-08
Gracefully handle missing command argument

st0012:
Currently, if users fail to provide a command's required argument(s), IRB treats it as normal `ArgumentError` and prints confusing message with full stacktrace:

```
irb(main):001:0> fg
/Users/hung-wulo/src/github.com/ruby/irb/lib/irb/extend-command.rb:242:in `irb_fg_': wrong number of arguments (given 0, expected 1) (ArgumentError)
        from /Users/hung-wulo/src/github.com/ruby/irb/lib/irb/extend-command.rb:247:in `irb_fg'
        from (irb):1:in `<main>'
        from /Users/hung-wulo/src/github.com/ruby/irb/lib/irb/workspace.rb:119:in `eval'
        from /Users/hung-wulo/src/github.com/ruby/irb/lib/irb/workspace.rb:119:in `evaluate'
        from /Users/hung-wulo/src/github.com/ruby/irb/lib/irb/context.rb:502:in `evaluate'
        .....
        from /Users/hung-wulo/.gem/ruby/3.1.2/bin/bundle:25:in `load'
        from /Users/hung-wulo/.gem/ruby/3.1.2/bin/bundle:25:in `<main>'
Maybe IRB bug!
```

This is not helpful because:

1. It's the same message users would see when they do hit an IRB bug.
1. The message doesn't tell users how to avoid the error.

So I want to introduce a new `CommandArgumentError` class as a separation from `ArgumentError`.
And commands with required argument(s), they should raise this error specifically when it's not supplied with clear message.

```
irb(main):001:0> fg
Please specify the id of target IRB job (listed in the `jobs` command).
=> nil
```

### DIFF
--- a/lib/irb/cmd/load.rb
+++ b/lib/irb/cmd/load.rb
@@ -17,24 +17,29 @@ module IRB
   # :stopdoc:
 
   module ExtendCommand
-    class Load < Nop
+    class LoaderCommand < Nop
       include IrbLoader
 
-      category "IRB"
-      description "Load a Ruby file."
-
-      def execute(file_name, priv = nil)
-        return irb_load(file_name, priv)
+      def raise_cmd_argument_error
+        raise CommandArgumentError.new("Please specify the file name.")
       end
     end
 
-    class Require < Nop
-      include IrbLoader
+    class Load < LoaderCommand
+      category "IRB"
+      description "Load a Ruby file."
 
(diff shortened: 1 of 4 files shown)


st0012:
cc @k0kubun

k0kubun (approved):

PR #476, opened 2022-12-08
Update readme's commands section

st0012:
I feel it's kinda nicer to read and A LOT easier to maintain if we just put the `show_cmds`'s output in the readme. WDYT? @k0kubun


k0kubun (approved):
I was thinking about the same thing actually. Perfect :+1:

PR #477, opened 2022-12-09
Fix step command

st0012:
The current `next` pre-command workaround on IRB source stepping moves the location by 1 extra line. 

```
From: test.rb @ line 1 :

 => 1: binding.irb
    2:
    3: c = 3
    4: d = 4

irb(main):001:0> step
(rdbg:irb) next
[1, 4] in test.rb
     1| binding.irb
     2|
=>   3| c = 3
     4| d = 4
=>#0    <main> at test.rb:3
(rdbg:irb) step
[1, 4] in test.rb
     1| binding.irb
     2|
     3| c = 3
=>   4| d = 4
=>#0    <main> at test.rb:4
(rdbg)
```

A better way is to make `debug` skip IRB frames completely. This PR does it by making `debug` think IRB paths as internal. This ensures users won't accidentally stepping into IRB frames in any way.

It also fixes the step command's test. The `|` in regexp was not escaped so it was always incorrectly matched.

### DIFF
--- a/lib/irb/cmd/debug.rb
+++ b/lib/irb/cmd/debug.rb
@@ -55,6 +55,13 @@ module IRB
         end
       end
 
+      module SkipPathHelperForIRB
+        def skip_internal_path?(path)
+          # The latter can be removed once https://github.com/ruby/debug/issues/866 is resolved
+          super || path.match?(IRB_DIR) || path.match?('<internal:prelude>')
+        end
+      end
+
       def setup_debugger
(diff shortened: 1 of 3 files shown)


st0012:
@k0kubun I think I found a bug of the `step` command and this PR should fix it and its test.

k0kubun (approved):
:+1: confirmed the behavior locally, and this looks better. Thank you!

PR #478, opened 2022-12-11
`show_doc` command should take non-string argument too

st0012:
Given that `show_doc` already supports syntax like `String#gsub`, it should be able to take it in non-string form too, like `edit` and `show_source` do. This ensures users can have a consistent syntax on argument between different commands.

- Before: `show_doc 'String#gsub'` 
- After: `show_doc 'String#gsub'`  and `show_doc String#gsub`

### DIFF
--- a/lib/irb/cmd/edit.rb
+++ b/lib/irb/cmd/edit.rb
@@ -18,13 +18,6 @@ module IRB
             args.strip.dump
           end
         end
-
-        private
-
-        def string_literal?(args)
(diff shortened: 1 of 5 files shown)


st0012:
@k0kubun To this single command, it's an enhancement. But to the overall user-experience, I think it's a fix on inconsistency. WDYT?

k0kubun (approved):
:+1:

PR #479, opened 2022-12-12
Group show_doc tests and update the expectation

st0012:
I created `ShowDocTest` because I see the tests [leaked to other test classes](http://rubyci.s3.amazonaws.com/osx1013/ruby-master/log/20221212T174503Z.fail.html.gz). I know it doesn't solve the root problem but it should reduce the chance of it.

And the tests are failing because I updated the expectation with document from  `Ruby 2.7.6` 🤦‍♂️ 

Now the new expectation should work with both Ruby head and Ruby 2.7.

### DIFF
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -376,38 +376,6 @@ module TestIRB
       assert_match(/Please specify the file name./, out)
     end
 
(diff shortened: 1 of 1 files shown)


st0012:
cc @k0kubun

k0kubun (approved):

PR #480, opened 2022-12-15
Prefer to use File.open instead of Kernel.open

hsbt:
CodeQL warns `Kernel.open` today. I replaced it to `File.open`.


st0012 (approved):
👍

PR #482, opened 2022-12-26
Fix wrong conf path with XDG_CONFIG_HOME.

hsbt:
It should be under the HOME directory, not current directory.

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -407,9 +407,9 @@ module IRB # :nodoc:
     end
     if home = ENV["HOME"]
       yield proc{|rc| home+"/.irb#{rc}"}
+      yield proc{|rc| home+"/.config/irb/irb#{rc}"}
     end
     current_dir = Dir.pwd
-    yield proc{|rc| current_dir+"/.config/irb/irb#{rc}"}
     yield proc{|rc| current_dir+"/.irb#{rc}"}
     yield proc{|rc| current_dir+"/irb#{rc.sub(/\A_?/, '.')}"}
     yield proc{|rc| current_dir+"/_irb#{rc}"}


lib/irb/init.rb
```diff
@@ -407,9 +407,9 @@ def IRB.rc_file_generators
     end
     if home = ENV["HOME"]
       yield proc{|rc| home+"/.irb#{rc}"}
+      yield proc{|rc| home+"/.config/irb/irb#{rc}"}
```
st0012, line 410: I see the specification says:

> If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

And given that `ENV["XDG_CONFIG_HOME"]` is handled above

```rb
    if xdg_config_home = ENV["XDG_CONFIG_HOME"]
      irb_home = File.join(xdg_config_home, "irb")
      if File.directory?(irb_home)
        yield proc{|rc| irb_home + "/irb#{rc}"}
      end
    end
````

Should we skip this line when `ENV.fetch("XDG_CONFIG_HOME", "").empty?` so when users change their config home we don't evaluate both?

hsbt: `XDG_CONFIG_HOME` is not always assigned on all of platform. Some of application enforce to use this directory without `XDG_CONFIG_HOME`.

This pull request is the simple bugfix. We should discuss about XDG specification and configuration policy of IRB in the another place, I think.

st0012 (approved):

lib/irb/init.rb
st0012: Fair enough 👍 I'm not familiar `XDG_CONFIG_HOME` so I just checked the specification and came up with that question. But I agree that it's not a blocker by any means.

PR #484, opened 2022-12-26
Improve RubyLex's tests

st0012:
When doing code reviews about RubyLex, it often took me a while to understand the test helpers' usage, and then the test's intention. So I think it may be worth refactoring the helpers a bit to make it easier for everyone.

It'd be easier to review/test the changes by commits.

(I know there are further refactors we can do, but I want to see how you think about it first)

### DIFF
--- a/test/irb/test_ruby_lex.rb
+++ b/test/irb/test_ruby_lex.rb
@@ -82,12 +82,6 @@ module TestIRB
       assert_equal(row.new_line_spaces, actual_next_line_spaces, error_message)
     end
 
-    def assert_indenting(lines, expected_space_count, add_new_line)
-      auto_indent_space_count = calculate_indenting(lines, add_new_line)
-      error_message = "Calculated the wrong number of spaces for:\n #{lines.join("\n")}"
-      assert_equal(expected_space_count, auto_indent_space_count, error_message)
-    end
-
     def assert_nesting_level(lines, expected, local_variables: [])
       ruby_lex = ruby_lex_for_lines(lines, local_variables: local_variables)
       error_message = "Calculated the wrong number of nesting level for:\n #{lines.join("\n")}"
@@ -166,8 +160,7 @@ module TestIRB
       lines = []
       input_with_correct_indents.each do |row|
         lines << row.content
-        assert_indenting(lines, row.current_line_spaces, false)
-        assert_indenting(lines, row.new_line_spaces, true)
+        assert_row_indenting(lines, row)
       end
     end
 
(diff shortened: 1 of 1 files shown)


hasumikin:
@st0012 
My poor knowledge of RubyLex can't help me to judge whether this refactoring is effective.
If you want to make things fast, I'm OK that you go ahead.
If you don't hurry, give me more several weeks to catch up :)

st0012:
It's not urgent by any means. My intention is just to help whoever works on `RubyLex` feels easier to write its tests.

st0012:
@tompng Given you're most familiar with RubyLex and its tests, could you give this change a look and let me know if it'll help your work?

tompng (changes requested):
The change looks good!
We always test both current_line_spaces and new_line_spaces, so the test seems clearer than before :+1:

test/irb/test_ruby_lex.rb
```diff
@@ -29,22 +27,60 @@ def teardown
...
       context.auto_indent_mode = true
+
       ruby_lex = RubyLex.new(context)
       io = MockIO_AutoIndent.new([lines, last_line_index, byte_pointer, add_new_line]) do |auto_indent|
         error_message = "Calculated the wrong number of spaces for:\n #{lines.join("\n")}"
         assert_equal(correct_space_count, auto_indent, error_message)
       end
+
       ruby_lex.set_input(io)
       ruby_lex.set_auto_indent
```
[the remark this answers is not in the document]
st0012: Good call 👍 Updated

test/irb/test_ruby_lex.rb
```diff
@@ -29,20 +30,59 @@ def teardown
...
       lines = lines + [""] if add_new_line
       last_line_index = lines.length - 1
       byte_pointer = lines.last.length
 
       context = build_context
       context.auto_indent_mode = true
+
       ruby_lex = RubyLex.new(context)
-      io = MockIO_AutoIndent.new([lines, last_line_index, byte_pointer, add_new_line]) do |auto_indent|
+      mock_io = MockIO_AutoIndent.new(lines, last_line_index, byte_pointer, add_new_line) do |auto_indent|
```
tompng, line 42: The block passed to `MockIO_AutoIndent.new` is not used because `&assertion` is deleted from parameter

st0012: 🤦‍♂️ Because I've moved the assertion out to `assert_row_indenting` and this is not needed.

tompng (approved):

PR #488, opened 2022-12-27
Remove unused context argument from Worksapce#evaluate

st0012:
The context argument was introduced in this change:

https://github.com/ruby/irb/commit/6806669d178f90f38b99c144bdfee06bdb93a229#diff-296327851fb7a2c307c2d0693b769f58c01aaf315972f290500d10081ee200a9

perhaps for potential usages. But after that it's never used.


hasumikin:
LGTM

PR #490, opened 2022-12-28
Update README's installation and usage sections.

st0012:
To my surprise, some devs that I've talked to didn't know `binding.irb` exists. So I think it may be worth mentioning it in the readme. (I know not many people would check this readme, but that's a topic for another day)

And the second commit is to clarify that `IRB` should already be installed by default in the recent Rubies.

I recommend reviewing the changes directly from the [branch](https://github.com/ruby/irb/tree/update-readme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,23 +22,50 @@ Or install it yourself as:
 
 ## Usage
 
-Use of irb is easy if you know Ruby.
+### The `irb` Executable
 
-When executing irb, prompts are displayed as follows. Then, enter the Ruby expression. An input is executed when it is syntactically complete.
+You can start a fresh IRB session by typing `irb` in your terminal.
 
-```
+In the session, you can evaluate Ruby expressions or even prototype a small Ruby script. An input is executed when it is syntactically complete.
+
+```shell
 $ irb
-irb(main):001:0> 1+2
-#=> 3
-irb(main):002:0> class Foo
-irb(main):003:1>   def foo
-irb(main):004:2>     print 1
-irb(main):005:2>   end
-irb(main):006:1> end
-#=> nil
+irb(main):001:0> 1 + 2
+=> 3
+irb(main):002:1* class Foo
+irb(main):003:2*   def foo
+irb(main):004:2*     puts 1
+irb(main):005:1*   end
+irb(main):006:0> end
+=> :foo
+irb(main):007:0> Foo.new.foo
+1
+=> nil
 ```
 
-The Readline extension module can be used with irb. Use of Readline is default if it's installed.
+### The `binding.irb` Breakpoint
+
+If you use Ruby 2.5+, you can also use `binding.irb` in your program as breakpoints.
+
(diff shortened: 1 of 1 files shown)


st0012:
@k0kubun would you mind giving it a look too? I used your [IRB article](https://qiita.com/k0kubun/items/4600e52a1a245495dfaa) as the version reference.

README.md
```diff
@@ -6,39 +6,76 @@ The `irb` command from your shell will start the interpreter.
 
 ## Installation
 
-Add this line to your application's Gemfile:
+> **Note**
+>
+> If you use Ruby 2.4 or later versions, a bundled version of IRB should already be installed with it by default.
```
k0kubun, line 11: What does this mean? I'm not sure if I said this in that article. IRB has been a standard library in versions older than 2.4 (`binding.irb` is just a new feature added in that version), and it's converted to a (default) gem in Ruby 2.6.

k0kubun: In other words, `gem install irb` or `gem 'irb'` works only for Ruby 2.6+.

st0012: Ah sorry I misread the article. What I want to express is that:

1. Each version of Ruby comes with a certain version of IRB
2. But users can chose to upgrade the version using the installation instructions

k0kubun (approved):
:+1:

README.md
```diff
@@ -6,39 +6,76 @@ The `irb` command from your shell will start the interpreter.
...
 ## Installation
 
-Add this line to your application's Gemfile:
+> **Note**
+>
+> IRB is a default gem of Ruby so you shouldn't need to install it separately.
+>
+> But if you're using Ruby 2.6 or later and want to upgrade/install a specific version of IRB, please follow these steps.
+
+To install it with `bundler`, add this line to your application's Gemfile:
```
ima1zumi, line 15: Since current IRB supports Ruby 2.6 or later, I thought you might as well remove the installation section. I will leave it to you.

README.md
```diff
@@ -6,39 +6,76 @@ The `irb` command from your shell will start the interpreter.
...
-irb(main):006:1> end
-#=> nil
+irb(main):001:0> 1 + 2
+=> 3
+irb(main):002:1* class Foo
+irb(main):003:2*   def foo
+irb(main):004:2*     puts 1
+irb(main):005:1*   end
+irb(main):006:0> end
+=> :foo
```
ima1zumi, line 50: 👍

ima1zumi (approved):

README.md
st0012: Sorry I don't get it. I think installation section is still helpful for users with Ruby 2.6+ to upgrade their IRB?
My thinking is, we can encourage power users using IRB as a gem. Because the more people doing that, the faster the feedback loop would be, which would help our development.

ima1zumi: Oh, sorry. I thought we didn't have to write about how to update default gems because it is the same for default gems, not just IRB. But it's easier to understand if you wrote it. I think you don't need to change it.

PR #491, opened 2022-12-28
Group command test cases with class

st0012:
Currently, IRB has 35 test cases for 10 non-debugging commands, with the rest 10 commands untested. So in the long-term, it could have around 70 test cases for all existing commands.

With this number of test cases, I think it's easier to manage them by grouping them in classes (by command).

Benefits:

- Isolating each command's test setup becomes easier, which is likely needed for testing workspaces related commands.
- It's easier to fold and outline test cases with editor features (see below).

**File Outline Before The Change**

(This outline is powered by [`ruby-lsp`](https://github.com/Shopify/ruby-lsp) 😉)

Screenshot 2022-12-28 at 21 42 33

**File Outline After The Change**

Screenshot 2022-12-28 at 21 41 36

### DIFF
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -7,23 +7,6 @@ require_relative "helper"
 
 module TestIRB
   class CommandTestCase < TestCase
-    def execute_lines(*lines, conf: {}, main: self, irb_path: nil)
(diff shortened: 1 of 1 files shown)


ima1zumi (approved):
Thank you for the refactoring!

PR #500, opened 2023-01-09
Rewrite RubyLex to fix some bugs and make it possible to add new features easily

### LINKED ISSUE #499: # RubyLex cannot handle some edge-case code correctly
## Description

While reading ruby-lex.rb code, I found these bugs.

### oneliner def

```ruby
irb(main):001:1* if 0
irb(main):002:2*   def f()
irb(main):003:2*     = 1 # This is oneliner def
irb(main):004:1*     2 # but RubyLex thinks it's a normal method def
irb(main):005:1*     3

irb(main):001:3* if 1;if 1;if 1;
irb(main):002:2*       (a,b)=1 # RubyLex thinks (a,b)=1 is oneliner def
irb(main):003:1*       (a,*)=1 # See the prompt. nesting leven is 1
irb(main):004:0*       (a,)=1 # nesting level is 0
irb(main):005:-*       (*)=1 # nesting level is -1
```

### keywords

```ruby
irb(main):001:1* def a.if
irb(main):002:2*     1
irb(main):003:1> end

irb(main):001:1* def a.end
irb(main):002:0* 1

irb(main):001:2* def ((a;b())).do
irb(main):002:2*     1
irb(main):003:1> end
irb(main):004:0>
```

### case in, pattern matching

```ruby
irb(main):001:1* case 1
irb(main):002:1* when 1
irb(main):003:2*   def f
irb(main):004:2*   1 in 2
irb(main):005:1*   end
irb(main):006:0> end

irb(main):001:2* def self.case
irb(main):002:2*     1
irb(main):003:2* 1 in 2
irb(main):004:2*   2
irb(main):005:1> end

irb(main):001:1* if :case
irb(main):002:1* 1 in 2
irb(main):003:0> end

irb(main):001:1* def f
irb(main):002:1*   1 in 2
irb(main):003:1*     2 in 3
```

### while do, until do, for do

```ruby
irb(main):001:1* while
irb(main):002:2*   p do
irb(main):003:2*     1
irb(main):004:1>   end

irb(main):001:1* until
irb(main):002:2*   p do
irb(main):003:2*     1
irb(main):004:1>   end

irb(main):001:1* for i in
irb(main):002:2*   p do
irb(main):003:2*     1
irb(main):004:1>   end

irb(main):001:1* for
irb(main):002:2*   i in j do
irb(main):003:2*       1

irb(main):030:1* if 1
irb(main):031:1*   for i in (p do end)
irb(main):032:1* end

irb(main):001:1* if :for.tap do
irb(main):002:1*   1
irb(main):003:0* end
irb(main):004:-> end

irb(main):001:0* for i in -> do end do
irb(main):002:0* 1
irb(main):003:-> end

irb(main):001:1* for i in :\
irb(main):002:1:   do
irb(main):003:1:     1
```

### heredoc

```ruby
irb(main):001:0" <<~A
irb(main):002:2" #{((
irb(main):003:1"     )
irb(main):004:0"     )
irb(main):005:0"     }
irb(main):006:0>     A
```

### unknown

```ruby
irb(main):001:1* if 1
irb(main):002:1*   (case 1;when 2;end)
irb(main):003:1*     (case 1;when 2;end)
irb(main):004:0> end

irb(main):001:1* if 1
irb(main):002:0*   (p do end)
irb(main):003:->   end

irb(main):001:1* if 1
irb(main):002:1*   if (p do end)
irb(main):003:0* end
```

## Result of irb_info
```
Ruby version: 3.2.0                                                                  
IRB version: irb 1.6.2 (2022-12-13)                                                  
InputMethod: RelineInputMethod with Reline 0.3.2                                     
.irbrc path: /Users/tomoya/.irbrc                                                    
LANG env: ja_JP.UTF-8                                                                
East Asian Ambiguous Width: 1      
```

## Terminal Emulator

macOS Terminal.app

tompng:
## Description

Fixes #499
In RubyLex, there are several parser-like methods listed below. To fix bugs, almost all of these methods needs change.

- process_nesting_level
- check_corresponding_token_depth
- check_newline_depth_difference
- check_string_literal
- is_the_in_correspond_to_a_for
- take_corresponding_syntax_to_kw_do
- in_keyword_case_scope?
- is_method_calling?
- heredoc_scope?

I combined them all into a single parser that calculates open tokens for each line.
Duplicated codes are reduced. Additions to ruby-lex.rb is about +310 and deletions is about -500.

### Rewrite nesting parser

Create `RubyLex::NestingParser` and use it from several methods.

```ruby
# Example of open tokens for each line
if true # open tokens: ['if']
  puts( # open tokens: ['if', '(']
    1   # open tokens: ['if', '(']
  )     # open tokens: ['if']
end     # open tokens: []
```

```ruby
each_line_result = RubyLex::NestingParser.parse_line(tokens)
line_tokens, open_tokens_before_line, open_tokens_after_line, minimum_token_depth_in_line = each_line_result[line_index]

# Example
puts((
    1 + 2
  ) + [ # HERE
    1
  ].size
)
# line_tokens: ['  ', ')', ' ', '+', ' ', '[', ' ', "# HERE\n"]
# open tokens changes: '((' → '(' (minimum) → '(['
# open_tokens_before_line: ['(', '(']
# open_tokens_after_line: ['(', '[']
# minimum_token_depth_in_line: 1
```

Prompt(check_corresponding_token_depth, check_string_literal), indent(process_indent_level, check_newline_depth_difference) and termination can be calculated from open tokens.

### Update test

- Add test for NestingParser
- Delete some test which will be tested in newly added test
- Fix test which was testing broken feature

### Refactor

I've done a minimal refactoring in ruby-lex.rb

- Reduce instance variables
- Split `readmultiline` from `each_top_level_statement`
- Refactor free indentation inside heredoc feature because it needs to be implemented in both `process_indent_level` and `check_corresponding_token_depth`

### Testability

We can now test these functionalities separately.

- calculate nesting for many ruby syntax
- open tokens to indent conversion
- open tokens to prompt conversion
- actual indentation

### Indent

We can now implement heredoc indent like the code below.
Previously, we cannot implement it because indent was using `indent += 1` and `indent -= 1`.

```ruby
if true
  if true
    s = <<HEREDOC
#{ # nesting level gets deeper but indent gets shallower
  (
    1
  )
}
heredoc
HEREDOC
    puts s # restores indent
  end
end
```

Parsing logic and indent calculation logic is separated, so we can now easily update indent of specific syntax.

```ruby
# easy to implement this indent
words = %w[
  irb
  reline
]
```

### Prompt

Prompt string is calculated from `ltype`, `indent`, `continue` and `line_no`.
We can now easily change it.

```
> string = "hello #{[
>   1, # ltype is currently `"` here because it is restricted to string-like literal
>   2, # changing ltype to `]` might be better
> ].join} world"
```

### Completion

Currently, completion is implemented using regexp. `array[index].???` shows `Array` methods and `array.map{}.???` shows `Hash | Proc` methods.
We can get an S-expression or a syntax tree of incomplete code and use it for accurate completion.

```ruby
closing_tokens_calculated_from_open_tokens = open_tokens.map { |t| closing_token_from(t) }
Ripper.sexp(incomplete_code + closing_tokens_calculated_from_open_tokens.join)
```


tompng:
changed to draft because it got infinite loop in ruby2.7. trying to fixing it now

st0012 (commented):
Thank you so much for the rewrite :pray:  I believe it’ll be a tremendous improvement to IRB.
But given the size of the change, we may need to go through this slowly with multiple reviews. I hope that’ll be ok.

I’ve given it a couple of scans and I have a question: Should we make `context` an instance variable?

From what I see, `RubyLex` can be used as an instance (e.g. `scanner.set_input`) or as a helper class (e.g. `RubyLex.generate_local_variables_assign_code`). And currently both usages would take `context` as an argument.

But for the former usages, the invocations could be largely simplified if `context` could be stored as an instance variable. So I did a quick search on `RubyLex.new` and I think it should be possible to do:

- In [`ShowSource`](https://github.com/ruby/irb/blob/master/lib/irb/cmd/show_source.rb#L51) it could access the context through `irb_context` (provided by [`Nop`](https://github.com/ruby/irb/blob/master/lib/irb/cmd/nop.rb))
- In [`Irb.initialize`](https://github.com/ruby/irb/blob/master/lib/irb.rb#L473-L476) it has context defined just a couple lines prior.
- In [`RubyLex.set_input`](https://github.com/ruby/irb/blob/master/lib/irb/ruby-lex.rb#L56) it also has access to context through itself. I wonder why we need to initialise another `RubyLex` instance here though.

If the answer is yes, I will to do that refactor before merging this PR, which is likely to cause some conflicts. So this is why I’m asking here :slightly_smiling_face:

tompng:
@st0012 

> Should we make context an instance variable?

Thanks for the explanation. I think it's a very good idea too. My answer is yes.

> likely to cause some conflicts

I'll resolve it when it's needed :smile:

lib/irb/ruby-lex.rb
```diff
@@ -47,14 +44,19 @@ def self.compile_with_errors_suppressed(code, line_no: 1)
     result
   end
 
+  def check_command(code, context:)
```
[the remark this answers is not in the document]
tompng: It returns boolean so name ends with `?` makes sense. 
How about just `command?` because rubocop Naming/PredicateName warns method name that starts with `is_`.

I also want to change `RubyLex#check_termination` to `terminated?`

```ruby
# Line 69~74
elsif check_command(code, context: context) # → is_command?(code, context: context) or command?(code, context: context)
  true
else
  code.gsub!(/\s*\z/, '').concat("\n")
  check_termination(code, context: context) # → terminated?(code, context: context)
end
```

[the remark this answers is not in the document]
st0012: Yeah that sounds good 👍

st0012:
I ended up doing 3 types of refactor:
- #502 
- #503 
- #504

I hope once they're all merged, it'll make the rewrite even simpler 🙂

st0012:
@tompng All 3 refactor PRs have been merged 😄

tompng:
Thanks, I rebased it

lib/irb/ruby-lex.rb
```diff
@@ -176,50 +151,26 @@ def self.ripper_lex_without_warning(code, context: nil)
     $VERBOSE = verbose
   end
 
-  def find_prev_spaces(line_index)
```
tompng, line 179: I dropped prev_spaces feature.

paste this code and type `\nend\nend`
```ruby
if 1
    def f
        abc
        de
        fg
```
then
```ruby
irb(main):001:1* if 1
irb(main):002:2*     def f
irb(main):003:2*         abc
irb(main):004:2*         de
irb(main):005:2*         fg
irb(main):006:1*     end
irb(main):007:1* end
=> :f
```

It is changed to
```ruby
irb(main):001:1* if 1
irb(main):002:2*     def f
irb(main):003:2*         abc
irb(main):004:2*         de
irb(main):005:2*     fg
irb(main):006:1*   end
irb(main):007:0> end
=> :f
```

This feature is not tested, and I thought it was broken because of the behavior when inserting newline between `a` and `bc`. But what I saw was a combination of another bug (#515 will fix it).
If it's not acceptable to drop in this pull request, I'll try not to drop.
If it's acceptable, I want to try re-implement it after #515.

st0012: I think as long as we ship this PR, #515, and the new implementation in the same release, it should be fine 👍  
Can you open an issue for this so we don't forget?

lib/irb/nesting_parser.rb
```diff
@@ -0,0 +1,205 @@
+module IRB
+  module NestingParser
+    IGNORE_TOKENS = %i[on_sp on_ignored_nl on_comment on_embdoc_beg on_embdoc on_embdoc_end]
+    def self.scan(tokens)
```
st0012, line 4: From its callers, it seems to return `opens`? Which I assume is an array of tokens that's "unclosed", like `(`?

If it is, I have some suggestions:

- Can you write some comments above the method as it's not super easy to understand due to its complexity
- Can we name it more explicitly, like `scan_opens`?

lib/irb/nesting_parser.rb
```diff
@@ -0,0 +1,205 @@
...
+        if pending_heredocs.any? && t.tok.include?("\n")
+          pending_heredocs.reverse_each { |t| opens << [t, nil] }
+          pending_heredocs = []
+        end
+        yield t, opens if block_given?
+      end
+      opens.map(&:first) + pending_heredocs.reverse
+    end
+
+    def self.parse_by_line(tokens)
```
st0012, line 179: Can we document this method's expected return values and how to interpret them?

lib/irb/ruby-lex.rb
```diff
@@ -65,68 +65,43 @@ def configure_io(io)
           else
             false
           end
-        else
+        elsif single_line_command?(code)
```
st0012, line 68: I wonder if `single_line_command` related changes could be extracted from the rest?

lib/irb/ruby-lex.rb
```diff
@@ -372,365 +321,74 @@ def check_code_block(code, tokens)
...
-    open_brace_on_line = 0
-    in_oneliner_def = nil
-
-    if heredoc_scope?
+  def check_corresponding_token_depth(tokens, lines, line_index)
+    line_results = IRB::NestingParser.parse_by_line(tokens)
+    result = line_results[line_index]
+    return unless result
+    _tokens, prev_opens, opens, min_depth = result
+    depth, = calc_nesting_depth(opens.take(min_depth))
```
st0012, line 370: What's the difference between `depth` and `min_depth`? Can we have a comment for it?

lib/irb/ruby-lex.rb
```diff
@@ -65,68 +65,43 @@ def configure_io(io)
...
-              ltype, indent, continue, code_block_open = check_state(code, partial_tokens)
-              result << @prompt.call(ltype, indent, continue || code_block_open, @line_no + line_num_offset)
-              line_num_offset += 1
-            end
-            unprocessed_tokens = []
-          else
-            code << t.tok
+        line_results = IRB::NestingParser.parse_by_line(tokens)
+        tokens_until_line = []
+        line_results.map.with_index do |(line_tokens, _prev_opens, next_opens), line_num_offset|
```
st0012, line 84: Nitpick: can we add `_min_depth` for clarity? I was confused a bit by the number of params being different 😅

lib/irb/ruby-lex.rb
```diff
@@ -237,11 +188,12 @@ def readmultiline
       # Accept any single-line input for symbol aliases or commands that transform args
       return code if single_line_command?(code)
 
-      ltype, indent, continue, code_block_open = check_code_state(code)
-      return code unless ltype or indent > 0 or continue or code_block_open
+      check_target_code, tokens, opens = check_code_state(code)
+      continue = process_continue(tokens)
```
st0012, line 192: I know continue is a boolean state but it's not very clear what it means. Would you mind explain it a bit?
Also, I feel `is_continue?` or `should_continue?` would probably be a much better name. Do you mind opening a PR to rename it first? It should help others understand this PR a bit easier.

lib/irb/ruby-lex.rb
```diff
@@ -65,68 +65,43 @@ def configure_io(io)
...
-          if ltype or indent > 0 or continue or code_block_open
-            false
-          else
-            true
-          end
+          true
+        else
+          check_target_code, tokens, opens = check_code_state(code)
+          continue = process_continue(tokens)
+          opens.empty? && !continue && !check_code_block(check_target_code, tokens)
```
st0012, line 74: I think these 3 lines could be extracted to a `code_terminated?` method? It can be use in `readmultiline` too and the intention would be clearer.

test/irb/test_nesting_parser.rb
```diff
@@ -0,0 +1,299 @@
...
+          until 1; end
+          for i in j; end
+          case 1; when 2; end
+          puts(1, 2, 3)
+          loop{|i|}
+          loop do |i| end
+        end
+      EOS
+      line_results = parse_by_line(code)
+      line_results[1...-1].each {|result| assert_equal(['class'], result[2].map(&:tok)) }
```
st0012, line 81: Can we use `1..<n>` here so we know exactly how many line results to expect? I think `1..-1` means it'll still pass even if we only get 1 result back?

test/irb/test_nesting_parser.rb
```diff
@@ -0,0 +1,299 @@
...
+        bbb
+        "
+        <<A
+        aaa
+        bbb
+        A
+      EOS
+      line_results = parse_by_line(code)
+      string_content_line, string_opens = line_results[1]
+      assert_equal("\naaa\nbbb\n", string_content_line.map(&:first).map(&:tok).join)
```
st0012, line 97: Since `string_content_line` only has 1 item, can we do `string_content_line.first.first.tok` instead?

test/irb/test_nesting_parser.rb
```diff
@@ -0,0 +1,299 @@
...
+      assert_equal("aaa\n", string_content_line.map(&:last).join)
```
st0012, line 98: `string_content_line.first.last`?

st0012 (commented):
I like that we're getting more detailed coverage with this rewrite and I think some small tweaks can make the new tests even more approachable 👍

lib/irb/nesting_parser.rb
tompng: I changed it to `scan_opens`. I also add `open_tokens` method that doesn't receive block and just return open tokens because `scan_opens` with block is only used inside nesting_parser.

lib/irb/ruby-lex.rb
tompng: I revert it

test/irb/test_nesting_parser.rb
tompng: Add line_results.size assertion,
Change `line_results[1...-1]` to `class_open, *inner_line_results, class_close = line_results` and add assertion for `class_open` and `class_close`

lib/irb/ruby-lex.rb
tompng: In readmultiline, opens and tokens are also used after checking for code_termination, so I've done some other refactor that check_code_state also returns termination state

test/irb/test_nesting_parser.rb
```diff
@@ -0,0 +1,303 @@
...
+          puts(1, 2, 3)
+          loop{|i|}
+          loop do |i| end
+        end
+      EOS
+      line_results = parse_by_line(code)
+      assert_equal(code.lines.size, line_results.size)
+      class_open, *inner_line_results, class_close = line_results
+      assert_equal(['class'], class_open[2].map(&:tok))
+      inner_line_results.each {|result| assert_equal(['class'], result[2].map(&:tok)) }
```
st0012, line 83: Just curious: why do we pick `class` as the target for checking Ruby syntax? Is the assumption: "if the class is not accidentally closed by any of the complicated syntax inside it, then we assume the nesting parser is working correctly"?

test/irb/test_nesting_parser.rb
```diff
@@ -0,0 +1,303 @@
+# frozen_string_literal: false
+require 'irb'
+
+require_relative "helper"
+
+module TestIRB
+  class TestNestingParser < TestCase
```
st0012, line 7: Nit: can we put `Test` as postfix instead of prefix?

lib/irb/ruby-lex.rb
```diff
@@ -69,64 +69,38 @@ def configure_io(io)
...
-              line_num_offset += 1
-            end
-            unprocessed_tokens = []
-          else
-            code << t.tok
+        line_results = IRB::NestingParser.parse_by_line(tokens)
+        tokens_until_line = []
+        line_results.map.with_index do |(line_tokens, _prev_opens, next_opens, _min_depth), line_num_offset|
+          line_tokens.each do |token, _s|
+            tokens_until_line << token if token != tokens_until_line.last
```
st0012, line 85: Why do we need this if condition? Looking at `process_continue`, having this token or not should make a difference. But even if I removed this line, all tests still pass.

lib/irb/ruby-lex.rb
```diff
@@ -372,365 +324,79 @@ def check_code_block(code, tokens)
...
-        pending_heredocs = []
-      end
-      i += 1
+  def ltype_from_open_tokens(opens)
+    start_token = opens.reverse_each.find do |tok|
+      %i[
+        on_heredoc_beg on_tstring_beg on_symbeg on_regexp_beg
+        on_symbols_beg on_qsymbols_beg
+        on_words_beg on_qwords_beg
+      ].include?(tok.event)
```
st0012, line 397: Can this array be a constant so we don't need to create a new one for every iteration?

st0012 (commented):
This is an amazing work, I think we're pretty close to merging it 👍

test/irb/test_nesting_parser.rb
tompng: Yes. the intention of the test is to ensure "class is not accidentally closed".
We can also use `if true` or other nesting syntax instead of `class A`

lib/irb/ruby-lex.rb
tompng: For this code
```ruby
if 1
  "aaa
  bbb
  ccc"
end
```
tokens on line 1 are: `["if", " ", "1", "\n"]`
tokens on line 2 are: `["  ", '"', "aaa\n  bbb\n  ccc"]`
tokens on line 3 are: `["aaa\n  bbb\n  ccc"]`
tokens on line 4 are: `["aaa\n  bbb\n  ccc", '"', "\n"]`
tokens on line 5 are: `["end"]`
It has duplication because token `on_tstring_content` can span multiline.

If I remove this condition, tokens_until_line will be:
`["if", " ", "1", "\n", "  ", "\"", "aaa\n  bbb\n  ccc", "aaa\n  bbb\n  ccc", "aaa\n  bbb\n  ccc", "\"", "\n", "end"]`
`"aaa\n  bbb\n  ccc"` is duplicated three times.

It will work without this condition because duplication does not affect the result of `process_continue.

lib/irb/ruby-lex.rb
tompng: `!process_continue(...) && !check_code_block(...)` is checking if the input still continues or not.

`process_continue` it is checking if the line ends with `\`, means user wants to write next line, and doing more thing.
`check_code_block` is checking if syntax-error code is recoverable or not, and doing more thing.

I actually don't know what it is doing on each method. I think there's overlap or duplication. IMO, later part of check_code_block where last_lex_state is used looks like it should belong to process_continue.

st0012: Thanks for the explanation. And yeah refactoring it later seems better.

lib/irb/ruby-lex.rb
st0012: Thx!

lib/irb/ruby-lex.rb
st0012: Given it's for optimisation purpose and currently can't be covered by tests, I think we can add a comment like:

```suggestion
            # avoid appending duplicated last tokens to keep the list small
            tokens_until_line << token if token != tokens_until_line.last
```

st0012: Oh I was confused by this:
>  but It will work without this condition because duplication does not affect the result of process_continue.

That comment looks good 👍

PR #501, opened 2023-01-10
Add timeout minutes for CI

ima1zumi:
Because it may run up to the default timeout of 6 hours.
https://github.com/ruby/irb/actions/runs/3874332006

The most recent run time was about 5 minutes; I don't think it will run for 30 minutes.


st0012 (approved):
This will make issues surface faster, thanks 👍 

(I required my review by mistake 😅)

PR #502, opened 2023-01-10
Remove redundant argument defaults from some RubyLex methods

st0012:
First 2 commits' argument defaults are not necessary because all of their callers specify the argument's value.
Given that these methods are only used within IRB, I feel leaving these defaults could be misleading.

The last commit is a similar cleanup by making one caller generating tokens beforehand.


hasumikin (approved):
LGTM👍

PR #503, opened 2023-01-10
Store context in RubyLex

st0012:
Some background/assumptions for this refactor:

1. Through a RubyLex instance's lifetime, the context passed to its methods should be the same. Given that `Context` is only initialised once in `Irb#initialize`, this should be true.

2. When `RubyLex` is initialised, the context object should be accessible. This is also true in all 3 of `RubyLex.new`'s invocations.

With the above observations, we should be able to store the context in `RubyLex` as an instance variable. And doing so will make `RubyLex`'s instance methods easier to use and maintain.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -468,7 +468,7 @@ module IRB
       @context = Context.new(self, workspace, input_method)
       @context.main.extend ExtendCommandBundle
       @signal_status = :IN_IRB
-      @scanner = RubyLex.new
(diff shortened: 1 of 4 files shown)


hasumikin (approved):
LGTM

I made a related PR https://github.com/ruby/reline/pull/498

st0012:
@hasumikin That's awesome, thank you 👍

PR #505, opened 2023-01-11
Remove needless headers

hsbt:
The current `irb` files keep some headers for old VCS system. The users confused by `0.9.6` revision different with `irb` version. 

We should remove them for the future development. And I styled their headers.


st0012 (approved):
Nice cleanup. Thanks 👍

ima1zumi (approved):
Thanks!

PR #507, opened 2023-01-12
Allow alias warnings to be silenced

shreeve:
Allow a user to silence the warning that certain methods cannot be aliased. The use case is for tools that use `irb` and are already aware of the warning, but deem it acceptable and just want to suppress it.

Code would need to do the following to hush these warnings:

```
IRB::ExtendCommandBundle::HushAliasWarning = true
```

### DIFF
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -306,7 +306,7 @@ module IRB # :nodoc:
           end
           alias_method to, from
         }
-      else
+      elsif !defined?(HushAliasWarning)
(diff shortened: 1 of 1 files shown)


shreeve:
An example of these warning is:

```
irb: warn: can't alias debug from irb_debug.
irb: warn: can't alias delete from irb_delete.
irb: warn: can't alias backtrace from irb_backtrace.
(main):001:0>
```

tompng:
With #824, method aliasing mechanism and this warning message is removed

PR #514, opened 2023-02-03
Make tests more compatible with TruffleRuby

st0012:
1. Some test cases don't fail on TruffleRuby anymore so we can remove the omit/pend conditions.
2. Some test cases can work with TruffleRuby with small tweaks on the test subject/assertion.

By making more tests run against TruffleRuby, we can help it match CRuby's behaviour quicker (example: https://github.com/oracle/truffleruby/issues/2848).

Thanks @nirvdrum for working with me on this. We'll have regular pairing sessions to remove the rest of the skipping conditions.

### DIFF
--- a/test/irb/test_context.rb
+++ b/test/irb/test_context.rb
@@ -64,7 +64,6 @@ module TestIRB
     end
 
     def test_eval_input
-      pend if RUBY_ENGINE == 'truffleruby'
       verbose, $VERBOSE = $VERBOSE, nil
       input = TestInputMethod.new([
         "raise 'Foo'\n",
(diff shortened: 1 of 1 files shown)


test/irb/test_context.rb
```diff
@@ -50,22 +50,20 @@ def test_evaluate_with_exception
     end
 
     def test_evaluate_with_encoding_error_without_lineno
-      pend if RUBY_ENGINE == 'truffleruby'
-      assert_raise_with_message(EncodingError, /invalid symbol/) {
-        @context.evaluate(%q[{"\xAE": 1}], 1)
+      # TruffleRuby adds 'encoding' in the middle of the message
```
nirvdrum, line 53: This will be fixed with https://github.com/oracle/truffleruby/issues/2850. I expect that'll change fairly quickly, so you may not want to modify the test. Of course, we can always change the test back once the message is changed in TruffleRuby.

st0012: 👍 since the issue has been addressed, I left the assertion unchanged.

PR #515, opened 2023-02-05
Improve indentation: bugfix, heredoc, embdoc, strings

tompng:
## Description 

- Implement heredoc and embdoc indent
- Implement string indent
- Fixes all indentation bug that I know now

left: irb 1.6.2 right: this branch

https://user-images.githubusercontent.com/1780201/216833960-609df48a-f1e2-4784-9fe7-408f839b3c96.mp4

## Heredoc and embdoc indent

I implemented these heredoc and embdoc indent

```ruby
if 1
  x = <<EOS
heredoc # initial indent should be 0
EOS
  puts x
=begin
embdoc # initial indent should be 0
=end
  y = <<~EOS1 + <<~EOS2
    heredoc1 # initial indent should be 4
    # indent of heredoc_end should be 2
  EOS1
    heredoc2
  EOS2
end
```

```ruby
<<~JS
  function f() {
    const x = 1 // any number of indent(>=2) should be accepted
    console.log(x) // next-line's initial indent should be same as prev line
  }
JS
```

## String indent

string, backtick, regexp and symbol

```ruby
def send_response
  # Inside string literal, any number of indent(>=0) should be accepted.
  # In master branch, two spaces will automatically inserted and make http header invalid.
  http_header = "HTTP/1.1 200 OK
Content-Type: text/html
Content-Length: #{body.bytesize}"
  socket.write "#{http_header}\n\n#{body}"
end
```

## Indentation bugfix

These bugs will be fixed
```ruby
> if 1
>   def f() = (█1 * 2)
>   1
# ↓ insert line-break by pressing option+enter
> if 1
>     def f() = (
>     █1 * 2)
>   1
```

```ruby
> if 1
>   a = 1█
>   1
# ↓ insert line-break by pressing option+enter
> if 1
> a = 1
> █
>   1
```

```ruby
if 1
  beginning
  en█ing
# ↓ type d
if 1
  beginning
end█ing
```

## Internal

auto_indent_proc `->(lines, line_index, byte_pointer, is_newline){}` was implemented as:

```ruby
if is_newline
  # next line's indent number
else
  # current line's indent if indent number is decreasing
  # nil (means any number of indent is accepted) if indent number is not decreasing (this will cause a bug)
  # note that it ignores right side of the cursor in the line which also cause a bug
end
```

But it is wrong. In most of the cases, it should return a number to avoid bug when inserting newline.

```ruby
if is_newline
  # next line's indent number
else
  # nil (means any number of indent is accepted) when deleting or adding indent spaces
  # checked by left side of the cursor is all space or not
  # otherwise, returns current line's indent
end
```

Also, special treatment is done for heredoc, embdoc and free-indent-acceptable literal like string.

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -93,6 +93,7 @@ class RubyLex
 
     if @io.respond_to?(:auto_indent) and @context.auto_indent_mode
       @io.auto_indent do |lines, line_index, byte_pointer, is_newline|
+        next nil if lines == [nil] # Workaround for exit IRB with CTRL+d
         next nil if !is_newline && lines[line_index]&.byteslice(0, byte_pointer)&.match?(/\A\s*\z/)
 
         code = lines[0..line_index].map { |l| "#{l}\n" }.join
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -93,16 +93,11 @@ class RubyLex
 
     if @io.respond_to?(:auto_indent) and @context.auto_indent_mode
       @io.auto_indent do |lines, line_index, byte_pointer, is_newline|
-        if is_newline
-          tokens = self.class.ripper_lex_without_warning(lines[0..line_index].join("\n"), context: @context)
-          process_indent_level(tokens, lines)
-        else
-          code = line_index.zero? ? '' : lines[0..(line_index - 1)].map{ |l| l + "\n" }.join
-          last_line = lines[line_index]&.byteslice(0, byte_pointer)
-          code += last_line if last_line
-          tokens = self.class.ripper_lex_without_warning(code, context: @context)
-          check_corresponding_token_depth(tokens, lines, line_index)
-        end
+        next nil if !is_newline && lines[line_index]&.byteslice(0, byte_pointer)&.match?(/\A\s*\z/)
+
+        code = lines[0..line_index].map { |l| "#{l}\n" }.join
+        tokens = self.class.ripper_lex_without_warning(code, context: @context)
+        process_indent_level(tokens, lines, line_index, is_newline)
       end
     end
   end
@@ -364,10 +359,16 @@ class RubyLex
   def calc_nesting_depth(opens)
     indent_level = 0
     nesting_level = 0
-    opens.each do |t|
+    opens.each_with_index do |t, index|
       case t.event
       when :on_heredoc_beg
-        # TODO: indent heredoc
+        if opens[index + 1]&.event != :on_heredoc_beg
+          if t.tok.match?(/^<<[~-]/)
+            indent_level += 1
+          else
+            indent_level = 0
+          end
+        end
       when :on_tstring_beg, :on_regexp_beg, :on_symbeg
         # can be indented if t.tok starts with `%`
       when :on_words_beg, :on_qwords_beg, :on_symbols_beg, :on_qsymbols_beg, :on_embexpr_beg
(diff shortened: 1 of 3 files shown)


test/irb/yamatanooroti/test_rendering.rb
```diff
@@ -80,7 +80,7 @@ def b; true; end
       irb(main):008:0>
       irb(main):009:0> a
       irb(main):010:0>  .a
-      irb(main):011:0>  .b
```
tompng, line 83: This indent was not a feature but a bug.

Paste this code and press return key
```ruby
a
  .b
  .c
if d
```
then you will get
```
irb(main):001:0> a
irb(main):002:0>   .b
irb(main):003:0>   .c
irb(main):004:1*   if d
irb(main):005:1*     
```

lib/irb/ruby-lex.rb
```diff
@@ -329,10 +326,16 @@ def check_code_block(code, tokens)
   def calc_nesting_depth(opens)
```
st0012, line 326: Not directly related to this PR, but it seems `indent_level` and `nesting_level` are always used separately. Given that `nesting_level`'s condition should be much simpler and is only used in `#prompt`, should we split their implementation?

Also, it seems weird that `RubyLex#prompt` passes `nesting_level` to `@prompt`, but in `set_prompt` it names that argument as `indent`.

tompng: I thought the implementation is similar so I combined it together.
But yes, splitting seems better now. I'll do it :+1:

lib/irb/ruby-lex.rb
```diff
@@ -93,16 +93,13 @@ def configure_io(io)
...
-          code = line_index.zero? ? '' : lines[0..(line_index - 1)].map{ |l| l + "\n" }.join
-          last_line = lines[line_index]&.byteslice(0, byte_pointer)
-          code += last_line if last_line
-          tokens = self.class.ripper_lex_without_warning(code, context: @context)
-          check_corresponding_token_depth(tokens, lines, line_index)
-        end
+        next nil if !is_newline && lines[line_index]&.byteslice(0, byte_pointer)&.match?(/\A\s*\z/)
+
+        code = lines[0..line_index].map { |l| "#{l}\n" }.join
+        next nil if code == "\n"
```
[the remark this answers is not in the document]
tompng: Nice catch, it looks like almost-unreachable code.

I think I added this to returns nil if lines are empty, but empty lines are not passed to this proc

lib/irb/ruby-lex.rb
```diff
@@ -382,46 +384,69 @@ def calc_nesting_depth(opens)
...
-    if prev_line_free_indent_tok && prev_line_free_indent_tok != free_indent_tok
-      return indent
-    elsif free_indent_tok
-      return lines[line_index][/^ */].length
+    indent_level, _nesting_level = calc_nesting_depth(prev_opens.take(min_depth))
+    indent = 2 * indent_level
+
+    preserve_indent = lines[line_index - (is_newline ? 1 : 0)][/^ */].size
+
+    if free_indent_token?(prev_opens.last)
```
st0012, line 413: nit: perhaps extracting `prev_opens.last` and `next_opens.last` to a locals given that they appear multiple times below?

lib/irb/ruby-lex.rb
```diff
@@ -382,46 +384,69 @@ def calc_nesting_depth(opens)
...
+          # Accept extra indent spaces inside `<<~` heredoc
+          [indent, preserve_indent].max
+        else
+          # Accept any number of indent inside other heredoc
+          preserve_indent
+        end
+      else
+        # Heredoc close
+        prev_line_indent_level, _prev_line_nesting_level = calc_nesting_depth(prev_opens)
+        tok.match?(/^<<[~-]/) ? 2 * (prev_line_indent_level - 1) : 0
```
st0012, line 445: Just to make sure I understand it correctly: even in the heredoc closing token we can still access the full heredoc content, including the opening syntax?

test/irb/test_ruby_lex.rb
```diff
@@ -98,9 +98,9 @@ def check_state(lines, local_variables: [])
       tokens = RubyLex.ripper_lex_without_warning(lines.join("\n"), context: context)
       opens = IRB::NestingParser.open_tokens(tokens)
       ruby_lex = RubyLex.new(context)
-      indent, _nesting_level = ruby_lex.calc_nesting_depth(opens)
+      indent, nesting_level = ruby_lex.calc_nesting_depth(opens)
       code_block_open = !opens.empty? || ruby_lex.process_continue(tokens)
-      [indent, code_block_open]
+      [indent, nesting_level, code_block_open]
```
st0012, line 103: I think `indent` is not used anymore after this PR?

lib/irb/ruby-lex.rb
tompng: We can access only the opening syntax.

For this code
```
-| code          | prev_opens      | next_opens
1| if true       | []              | [if]
2|   <<A + <<~B  | [if]            | [if, <<~B, <<A]
3| content_a     | [if, <<~B, <<A] | [if, <<~B, <<A]
4| A             | [if, <<~B, <<A] | [if, <<~B]
5|     content_b | [if, <<~B]      | [if, <<~B]
6|   B           | [if, <<~B]      | [if]
7| end           | [if]            | []
```
In heredoc closing line (L4 and L6), `prev_opens.last` (`<<A` for line 4, `<<~B` for line 6) is the corresponding heredoc_beg token.

st0012: Hmm, that's an very interesting behaviour. Though I guess that's required so the parser can later know what'd close the heredoc.
Thanks for explaining 👍

lib/irb/ruby-lex.rb
```diff
@@ -382,46 +384,69 @@ def calc_nesting_depth(opens)
...
+      if prev_opens.last&.event == next_opens.last&.event
+        # Accept any number of indent inside embdoc content
+        preserve_indent
+      else
+        # =begin or =end
+        0
+      end
+    elsif prev_opens.last&.event == :on_heredoc_beg
+      tok = prev_opens.last.tok
+      if prev_opens.size < next_opens.size || prev_opens.last == next_opens.last
```
tompng, line 431: I think I can make this condition simpler  like `if prev_opens.size <= next_opens.size`

When heredoc is closed, `next_opens.size` is always smaller than `prev_opens.size` because only heredoc_end can be on the line.
```ruby
<<A # heredoc_beg
A + ( # heredoc_end + lparen ← this is not possible
A
↑ heredoc ends here
```

st0012 (approved):

PR #516, opened 2023-02-16
Improve actions

hsbt:
* Use reusing workflow for automatically detect Ruby versions
* Use stable branch of `actions/checkout`
* Reduce build matrix by single os value.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,8 @@ jobs:
       matrix:
         ruby: ${{ fromJson(needs.ruby-versions.outputs.versions) }}
         with_latest_reline: [true, false]
+        exclude:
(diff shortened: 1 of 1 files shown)


st0012:
Thank you!

st0012:
Thanks for the reminder 👍 

I haven't had time to check jruby related tests yet, but [the failure in truffleruby](https://github.com/ruby/irb/actions/runs/4190760736/jobs/7264491547) has been addressed from source and just waits for release.

That being said, I'm now working regularly with @nirvdrum to remove skipped truffleruby tests and fix them upstream ([example](https://github.com/ruby/irb/pull/514)). So I'll only test IRB against `truffleruby-head` for now until all fixes have been completed and released.

PR #520, opened 2023-02-19
Cleanup completion tests

st0012:
- The first test was added for an implementation's edge case that's been replaced a few years ago. So I think it's not needed anymore. Please see the commit message for more info.

- The `pend` doesn't seem to be needed as tests can pass without it.


tompng:
LGTM 👍

PR #522, opened 2023-02-24
Remove unused Struct

ima1zumi:
`DefaultEncodings` is not used, so remove it.

```
~/ghq/github.com/ruby/irb remove-defaultencodings
❯ git grep DefaultEncodings

```

It was created with this commit. But it is no longer in use.
https://github.com/ruby/irb/commit/3ee79e89adb8e21b63d796e53bcc86281685076d

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -421,8 +421,6 @@ module IRB # :nodoc:
     end
   end
 
-
-  DefaultEncodings = Struct.new(:external, :internal)
   class << IRB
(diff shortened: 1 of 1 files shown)


st0012 (approved):
I also did an [org-wide search](https://github.com/search?q=org%3Aruby+DefaultEncodings&type=code) and didn't find anything using it either 👍

tompng:
LGTM and thanks for the org-wide search

PR #523, opened 2023-02-25
Improve method completion for string and regexp that includes word break characters

### LINKED ISSUE #314: The completion dialog following the parentheses is wrong

tompng:
## Description

fixes #314

In these case, irb cannot correctly complete string and regexp methods.
- `"()".`
- `"hello world".`
- /hello world/.

This pull request fixes it.

## Background

Readline and Reline has a configuration `completer_quote_characters` and `completer_word_break_characters` of completion target.
It was able to calculate `"()".` completion target correctly.

In this pull request(https://github.com/ruby/irb/pull/212), `Reline.completer_quote_characters = ''` is added, which means reline does not consider quotes in completion target calculation.
Completion target calculated by reline changed from `"hello world".` to `world".` and the regexp needed to be updated.

## Change

I changed string completion regexp to match `anything".` I also changed regexp for regexp completion.
Since array completion regexp matches to `].`, changing string completion regexp to match `".` is acceptable(in my opinion).

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -166,10 +166,10 @@ module IRB
 
     def self.retrieve_completion_data(input, bind: IRB.conf[:MAIN_CONTEXT].workspace.binding, doc_namespace: false)
       case input
-      when /^((["'`]).*\2)\.([^.]*)$/
+      when /^(.*["'`])\.([^.]*)$/
         # String
         receiver = $1
-        message = $3
+        message = $2
 
         if doc_namespace
           "String.#{message}"
@@ -178,7 +178,7 @@ module IRB
           select_message(receiver, message, candidates)
         end
 
-      when /^(\/[^\/]*\/)\.([^.]*)$/
+      when /^(.*\/)\.([^.]*)$/
         # Regexp
         receiver = $1
         message = $2
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -166,6 +166,8 @@ module IRB
 
     def self.retrieve_completion_data(input, bind: IRB.conf[:MAIN_CONTEXT].workspace.binding, doc_namespace: false)
       case input
+      # this regexp only matches the closing character because of irb's Reline.completer_quote_characters setting
+      # details are described in: https://github.com/ruby/irb/pull/523
       when /^(.*["'`])\.([^.]*)$/
         # String
         receiver = $1
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -180,6 +180,8 @@ module IRB
           select_message(receiver, message, candidates)
         end
 
+      # this regexp only matches the closing character because of irb's Reline.completer_quote_characters setting
+      # details are described in: https://github.com/ruby/irb/pull/523
       when /^(.*\/)\.([^.]*)$/
         # Regexp
         receiver = $1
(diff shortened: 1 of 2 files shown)


test/irb/test_completion.rb
```diff
@@ -14,11 +14,19 @@ def setup
     class TestMethodCompletion < TestCompletion
       def test_complete_string
         assert_include(IRB::InputCompletor.retrieve_completion_data("'foo'.up", bind: binding), "'foo'.upcase")
+        assert_not_include(IRB::InputCompletor.retrieve_completion_data("'foo'.up", bind: binding), "'foo'.update")
```
tompng, line 17: I think the test was not working because completion will fallback to listup all methods using ObjectSpace search.
all methods including upcase will be shown, and the test in L16 passes whether string completion is implemented or not.
I add this `assert_not_include` to check that it does not include `Hash#update`

st0012:
> Completion target calculated by reline changed from "hello world". to world".

Is it possible to resolve the issue by assigning proper `completer_quote_characters`? This behaviour doesn't seem intuitive to me.

> If it does not match, it will fallback to showing all-methods that searches ObjectSpace. Avoiding it will improve performance a little bit.

This should probably be updated.

tompng:
> Is it possible to resolve the issue by assigning proper completer_quote_characters? 

I think it's not possible now. If I add `Reline.completer_quote_characters = %('")`, closing quote is automatically added and will be impossible to input. I didn't figure it out whether it's a reline's bug or just configuration missing.

https://user-images.githubusercontent.com/1780201/222971766-83a6848e-b7eb-4d7a-a37b-c6468866c181.mp4

tompng:
> This should probably be updated.

OK :+1: I removed the `assert_not_inlcude`

lib/irb/completion.rb
```diff
@@ -166,10 +166,10 @@ def self.retrieve_files_to_require_relative_from_current_dir
 
     def self.retrieve_completion_data(input, bind: IRB.conf[:MAIN_CONTEXT].workspace.binding, doc_namespace: false)
       case input
-      when /^((["'`]).*\2)\.([^.]*)$/
+      when /^(.*["'`])\.([^.]*)$/
```
st0012, line 169:
```suggestion
      # this regexp only matches the closing character because of irb's Reline.completer_quote_characters setting
      # details are described in: https://github.com/ruby/irb/pull/523
      when /^(.*["'`])\.([^.]*)$/
```

lib/irb/completion.rb
```diff
@@ -178,7 +178,7 @@ def self.retrieve_completion_data(input, bind: IRB.conf[:MAIN_CONTEXT].workspace
           select_message(receiver, message, candidates)
         end
 
-      when /^(\/[^\/]*\/)\.([^.]*)$/
+      when /^(.*\/)\.([^.]*)$/
```
st0012, line 181:
```suggestion
      # this regexp only matches the closing character because of irb's Reline.completer_quote_characters setting
      # details are described in: https://github.com/ruby/irb/pull/523
      when /^(.*\/)\.([^.]*)$/
```

st0012 (commented):
Hmm it looks like a reline's bug, or `completer_quote_characters` has very limited use cases if that's intentional.
Can you add comments before the changed regexps to give readers some context?

st0012 (approved):

PR #526, opened 2023-02-27
Add a comment about Rails' patch on Workspace#filter_backtrace

st0012:


lib/irb/workspace.rb
```diff
@@ -122,6 +122,8 @@ def local_variable_get(name)
     end
 
     # error message manipulator
+    # WARN: Rails patches this method to filter its own backtrace. Be cautious when changing it.
+    # See: https://github.com/rails/rails/blob/main/railties/lib/rails/commands/console/console_command.rb#L8
```
ima1zumi, line 126: Main branch will change in the future, so it is better to refer to commit.
```suggestion
    # See: https://github.com/rails/rails/blob/dccb336c12a8642439addb3a4b3ff7fb081b2e37/railties/lib/rails/commands/console/console_command.rb#L8
```

st0012: I actually intentionally point it to the current `main`. Because if it's not there, we'll need this comment to be updated or removed. If we point it to a commit, we won't know if the implementation has changed and/or this comment has gone outdated.

ima1zumi: I see your point but I have different opinions.

Even if the method definition location has changed, it is easy to find the current method location from past commits. However, it would be difficult to find what was pointed to in the past from the future.

Also, I would not want to change the comments every time the method location changes. I would like to reduce maintenance.

sorah:
```suggestion
    # See: https://github.com/rails/rails/blob/main/railties/lib/rails/commands/console/console_command.rb#L8:~:text=def,filter_backtrace
```

?

ima1zumi: I didn't know how to create text fragments. Cool!
https://developer.mozilla.org/en-US/docs/Web/Text_fragments

st0012: @ima1zumi I agree with your point in general. But in this specific case, the code hasn't been updated for 7 straight years ([blame](https://github.com/rails/rails/blame/main/railties/lib/rails/commands/console/console_command.rb#L8)). And consider we actually don't plan to change it either, I don't worry about the maintenance effort to much.

> Even if the method definition location has changed, it is easy to find the current method location from past commits. However, it would be difficult to find what was pointed to in the past from the future.

I think it depends on the goal. IMO, the purpose for the comment is to help readers check if that specific use case by Rails is still there. Therefore, pointing to a specific comment would require 2 steps to do that:

1. Go to that commit page
2. Select the `main` branch to see if it's still there

In which the reader would eventually arrive the same page I'm pointing now.

I understand that it's not that clear as for what we expect the reader to check, but I think @sorah's amazing tip (thank you 🙌 ) has improved that significantly.

ima1zumi: > But in this specific case, the code hasn't been updated for 7 straight years ([blame](https://github.com/rails/rails/blame/main/railties/lib/rails/commands/console/console_command.rb#L8)). And consider we actually don't plan to change it either, I don't worry about the maintenance effort to much.

Oh, it hasn't been updated in 7 years. In this case, I thought you were right!

ima1zumi:
LGTM 😄  Thanks!

PR #527, opened 2023-02-27
Remove no longer necessary TruffleRuby test exclusions

nirvdrum:
I've been working through the set of TruffleRuby exclusions in the IRB test suite with @st0012. We've fixed a few compatibility issues upstream that have allowed us to remove test exclusions.

This PR needs https://github.com/oracle/truffleruby/pull/2897 to be merged and integrated into a build before the tests will pass.


st0012 (approved):
Thank you for the PR!

PR #528, opened 2023-02-28
Handle long inspect and control character in prompt string

### LINKED ISSUE #153: Wrong cursor position when self.to_s is a multi-line string
## Description

When `self.to_s` is a multi-line string or contains newline characters, the cursor is displayed at the wrong column in the IRB prompt. Also, typing any character causes the prompt redrawn at the wrong line.

```console
$ ruby -e '("a\n"*10).instance_eval { binding.irb }'
```

image

This screenshot shows an IRB prompt with a 10-line string as self and "1" typed in the prompt.
We expected the cursor to be positioned next to the "> " (or next to "1" after typing) and no duplicate prompts.

## Result of irb_info

Built from ruby v3_0_0_rc2 https://github.com/ruby/ruby/commit/a89932799c914d1864fc91f33d6461592dbfd3fb

```
irb(main):001:0> irb_info
=>
Ruby version: 3.0.0
IRB version: irb 1.2.9 (2020-12-22)
InputMethod: ReidlineInputMethod with Reline 0.1.10
```

## Terminal Emulator

- mintty 3.2.0 (Ruby runs on a Linux box over SSH)
- iTerm2 3.4.3 (Ruby runs on macOS)

## Setting Files

Are you using `~/.irbrc` and `~/.inputrc`? : Reproduces without .irbrc.

tompng:
Fixes #153 by replacing `"\x00-\x1F"` to `" "`
I also add truncation because changing workspace to a large array makes input hard.

## Background
Reline fixed #153 by escaping only `\n`. `\\` is not escaped and I think it's an inconsistent behavior.
Other control character should also be escaped or deleted or replaced in IRB side.

## Before
```
irb(main):001:0> cws "c:\\windows\\system32\n"
irb(c:\windows\system32\n):002:0> # only "\n" is escaped, "\\" is not escaped

irb(main):001:0> cws "HTTP/1.1 200 OK\r\n"
):002:0> # broken prompt

irb(main):001:0> cws [1]*100
irb([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1
, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1
, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1
, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1
, 1]):005:0> 
```

## After
```
irb(main):001:0> cws "c:\\windows\\system32\n"
irb(c:\windows\system32 ):002:0>

irb(main):001:0> cws "HTTP/1.1 200 OK\r\n"
irb(HTTP/1.1 200 OK  ):002:0> 

irb(main):001:0> cws [1]*100
irb([1, 1, 1, 1, 1, 1, 1, 1, 1, 1...):003:0> 
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -775,6 +775,15 @@ module IRB
       end
     end
 
+    def truncate_prompt_main(str, size = 32, omission: '...') # :nodoc:
+      str = str.tr("\x00-\x1F", ' ')
+      if str.size <= size
+        str
+      else
+        str[0, size - omission.size] + omission
+      end
+    end
+
     def prompt(prompt, ltype, indent, line_no) # :nodoc:
       p = prompt.dup
       p.gsub!(/%([0-9]+)?([a-zA-Z])/) do
@@ -782,9 +791,9 @@ module IRB
         when "N"
           @context.irb_name
         when "m"
-          @context.main.to_s
+          truncate_prompt_main(@context.main.to_s)
         when "M"
-          @context.main.inspect
+          truncate_prompt_main(@context.main.inspect)
         when "l"
           ltype
         when "i"
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -463,6 +463,10 @@ module IRB
     # be parsed as :assign and echo will be suppressed, but the latter is
     # parsed as a :method_add_arg and the output won't be suppressed
 
+    PROMPT_MAIN_TRUNCATE_LENGTH = 32
+    PROMPT_MAIN_TRUNCATE_OMISSION = '...'.freeze
+    CONTROL_CHARACTERS_PATTERN = "\x00-\x1F".freeze
+
     # Creates a new irb session
     def initialize(workspace = nil, input_method = nil)
       @context = Context.new(self, workspace, input_method)
@@ -775,12 +779,12 @@ module IRB
       end
     end
 
-    def truncate_prompt_main(str, size = 32, omission: '...') # :nodoc:
-      str = str.tr("\x00-\x1F", ' ')
-      if str.size <= size
+    def truncate_prompt_main(str) # :nodoc:
+      str = str.tr(CONTROL_CHARACTERS_PATTERN, ' ')
+      if str.size <= PROMPT_MAIN_TRUNCATE_LENGTH
         str
       else
-        str[0, size - omission.size] + omission
+        str[0, PROMPT_MAIN_TRUNCATE_LENGTH - PROMPT_MAIN_TRUNCATE_OMISSION.size] + PROMPT_MAIN_TRUNCATE_OMISSION
       end
     end
 
--- a/test/irb/test_context.rb
+++ b/test/irb/test_context.rb
@@ -656,6 +656,23 @@ module TestIRB
       $VERBOSE = verbose
     end
 
+    def test_prompt_main_escape
+      irb = IRB::Irb.new(IRB::WorkSpace.new("main\a\t\r\n"))
+      assert_equal("irb(main    )>", irb.prompt('irb(%m)>', nil, 1, 1))
+    end
+
+    def test_prompt_main_inspect_escape
+      main = Struct.new(:inspect).new("main\\n\nmain")
+      irb = IRB::Irb.new(IRB::WorkSpace.new(main))
+      assert_equal("irb(main\\n main)>", irb.prompt('irb(%M)>', nil, 1, 1))
+    end
+
+    def test_prompt_main_truncate
+      irb = IRB::Irb.new(IRB::WorkSpace.new("a" * 100))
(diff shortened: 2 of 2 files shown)


lib/irb.rb
```diff
@@ -775,16 +775,25 @@ def signal_status(status)
       end
     end
 
+    def truncate_prompt_main(str, size = 32, omission: '...') # :nodoc:
```
st0012, line 778: Do we have plan to specify `size` and `omission` in follow up PRs? If not, maybe we can specify them in constants first? (omission can use `.freeze`)

lib/irb.rb
```diff
@@ -775,16 +775,25 @@ def signal_status(status)
...
+      str = str.tr("\x00-\x1F", ' ')
```
st0012, line 779: Let's make it a constant with `.freeze` too.

test/irb/test_context.rb
```diff
@@ -656,6 +656,23 @@ def test_eval_input_with_long_exception
...
+
+    def test_prompt_main_inspect_escape
+      main = Struct.new(:inspect).new("main\\n\nmain")
+      irb = IRB::Irb.new(IRB::WorkSpace.new(main))
+      assert_equal("irb(main\\n main)>", irb.prompt('irb(%M)>', nil, 1, 1))
+    end
+
+    def test_prompt_main_truncate
+      irb = IRB::Irb.new(IRB::WorkSpace.new("a" * 100))
+      assert_match(/irb\(a.+\.\.\.\)>/, irb.prompt('irb(%m)>', nil, 1, 1))
```
st0012, line 672: I'm not good with regexp, so I'd appreciate if we can still use `assert_equal` here to avoid regexp computation in my head 😂 

```rb
assert_equal(
  "irb(aaaaaaaaaaaaaaaaaaaaaaaaaaaaa...)>",
  irb.prompt('irb(%m)>', nil, 1, 1))
)
```

st0012 (commented):
Thank you for the fix!
LGTM, just some small comments.

lib/irb.rb
tompng: >  follow up PRs?

no plan, I'll use constants :+1:

st0012 (approved):

lib/irb.rb
```diff
@@ -775,16 +779,25 @@ def signal_status(status)
       end
     end
 
+    def truncate_prompt_main(str) # :nodoc:
+      str = str.tr(CONTROL_CHARACTERS_PATTERN, ' ')
```
ima1zumi, line 783: It seems better not to use byte strings because they fail in ASCII-incompatible character encodings. I'll think about how to do it. 🤔 

```
irb(main):001:0> cws ("a".encode(Encoding::UTF_16BE))
irb: warn: can't alias delete from irb_delete.
=> "a"
/Users/mi/ghq/github.com/ruby/irb/lib/irb.rb:783:in `tr': incompatible character encodings: UTF-16BE and UTF-8 (Encoding::CompatibilityError)
```

ima1zumi: I thought it would be better to address this as a separate Issue. Because it won't work even if we fix just this part.

lib/irb.rb
```diff
@@ -463,6 +463,10 @@ class Irb
     # be parsed as :assign and echo will be suppressed, but the latter is
     # parsed as a :method_add_arg and the output won't be suppressed
 
+    PROMPT_MAIN_TRUNCATE_LENGTH = 32
```
ima1zumi, line 466: Why did you choose 32?
(No intention to disagree)

tompng:
About `irb: warn: can't alias delete from irb_delete.`

It reproduces and it's another issue. here is some other examples
```ruby
irb(main):001:0> cws 1
irb: warn: can't alias step from irb_step.
=> 1
irb(1):002:0> cws []
irb: warn: can't alias delete from irb_delete.
=> []
irb([]):003:0> cws Struct.new(:show_cmds).new
irb: warn: can't alias show_cmds from irb_show_cmds.
=> #<struct  show_cmds=nil>                               
```

lib/irb.rb
tompng: No particular reason. I thought 30-40 would be appropriate, just choose 32.

`Object.new.to_s.size => 28` so it should be >= 28

ima1zumi (approved):
LGTM

PR #531, opened 2023-03-01
Add basic release configuration

st0012:
This will help us generate a release note similar to [this one](https://github.com/Shopify/ruby-lsp/releases/tag/v0.4.1).

PR #532, opened 2023-03-01
Specify metadata to provide richer information on rubygems.org

st0012:
Since we're encouraging users to use `irb` as a gem, I feel it'd be worth enriching its rubygems.org page a bit and we can customise the sidebar items with the [`metadata` attribute](https://guides.rubygems.org/specification-reference/#metadata).

### DIFF
--- a/irb.gemspec
+++ b/irb.gemspec
@@ -16,6 +16,11 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://github.com/ruby/irb"
   spec.licenses      = ["Ruby", "BSD-2-Clause"]
 
+  spec.metadata["homepage_uri"] = spec.homepage
+  spec.metadata["source_code_uri"] = spec.homepage
+  spec.metadata["documentation_uri"] = spec.homepage
+  spec.metadata["changelog_uri"] = "#{spec.homepage}/releases"
+
   spec.files         = [
     ".document",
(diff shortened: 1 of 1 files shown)


irb.gemspec
```diff
@@ -16,6 +16,11 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://github.com/ruby/irb"
   spec.licenses      = ["Ruby", "BSD-2-Clause"]
 
+  spec.metadata["homepage_uri"] = spec.homepage
+  spec.metadata["source_code_uri"] = spec.homepage
+  spec.metadata["documentation_uri"] = spec.homepage
```
st0012, line 21: I specifically override the documentation's destination because the readme here has the most up-to-date info. We can modify it later when we improve the documentation on docs.ruby-lang.org.

irb.gemspec
```diff
@@ -16,6 +16,11 @@ Gem::Specification.new do |spec|
...
+  spec.metadata["changelog_uri"] = "#{spec.homepage}/releases"
```
st0012, line 22: I figure we don't want to maintain a `CHANGELOG` file yet so let's just use the release generated by GH for now, which will be enhanced after #531.

hasumikin (approved):
LGTM 👍

PR #534, opened 2023-03-02
Avoid slow symbol completion when completion target is an empty symbol

tompng:
I changed not to complete symbols when you just type `:`. You need one more character to start symbol completion.

If you type `:`, all symbols was shown in the completion dialog and it was slow.
It's because of the completion candidates size.
```
% ruby -e "p Symbol.all_symbols.size"
3737
```

```
% rails c
Loading development environment (Rails 7.0.4.2)
irb(main):001:0> Symbol.all_symbols.size
=> 28738
```

Also, the top completion candidates are non-alphabet symbols. I think it's hard to find a symbol what you want by pressing [tab] in this situation. I always type one more character before looking the completion candidate for symbol. 
```
% ruby -e "p Symbol.all_symbols.take(20)"
[:!, :"\"", :"#", :"$", :%, :&, :"'", :"(", :")", :*, :+, :",", :-, :".", :/, :":", :";", :<, :"=", :>]
```

Changed the regexp `/^(:[^:.]*)$/` → `/^(:[^:.]+)$/`.
It still matches `:a` but does not match to `:` anymore.

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -215,7 +215,7 @@ module IRB
           select_message(receiver, message, proc_candidates | hash_candidates)
         end
 
-      when /^(:[^:.]*)$/
+      when /^(:[^:.]+)$/
         # Symbol
         if doc_namespace
           nil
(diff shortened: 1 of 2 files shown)


st0012 (approved):

ima1zumi (approved):
Good catch!

PR #535, opened 2023-03-02
Fix prompt test not to change STDIO.external_encoding

tompng:
In the test I added in #528, I forgot to pass TestInputMethod as input_method to `IRB::Irb.new`, and it caused changing STDOUT.external_encoding from `nil` to `utf-8` in the test case

### DIFF
--- a/test/irb/test_context.rb
+++ b/test/irb/test_context.rb
@@ -19,11 +19,13 @@ module TestIRB
       def Reline.get_screen_size
         [36, 80]
       end
+      save_encodings
     end
 
     def teardown
       Reline.instance_eval { undef :get_screen_size }
       Reline.define_singleton_method(:get_screen_size, @get_screen_size)
+      restore_encodings
     end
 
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):

ima1zumi:
This problem seems to be resolved with https://github.com/ruby/irb/pull/536.

@tompng 
Do you need to make any additional changing? For example, adding `TestInputMethod.new`.

st0012:
I've hit the similar problem before, so I added [these helpers](https://github.com/ruby/irb/blob/master/test/irb/helper.rb#L45-L57) that we can use in test setup/cleanup.

tompng:
> Do you need to make any additional changing?

I need it to avoid changing external_encoding

I use TestInputMethod.new for not changing the encoding because not specifying the input was a fault,  and also use the helper to ensure it.
Thanks for the information about the helpers

st0012 (approved):

PR #539, opened 2023-03-10
Use symbol.inspect in symbol completion candidate calculation

tompng:
This pull request change to use `Symbol#inspect` instead of `":" + symbol.id2name` in symbol completion candidate calculation.

IRB sometimes shows syntax-invalid symbol ( `:@`, not `:"@"` ) in symbol completion and this pull request will fix it.
スクリーンショット 2023-03-11 2 41 09

### Background
When you type `:"aaaaaa` to IRB, symbol `:"aaa\n"` `:"aaaa\n"` `:"aaaaa\n"`  is created, and this was breaking completion dialog rendering when you type `:aa`.
スクリーンショット 2023-03-11 2 53 15

### About EncodingError

~Symbol#inspect seems to return string with encoding=default_external and does not seem to raise EncodingError.
I removed `encode` and `rescue`.~ 
```ruby
Encoding.default_external = Encoding::EUCJP
sym = "K".force_encoding('UTF-7').to_sym
puts sym.inspect #=> :"\x4B"
sym.inspect.encoding #=> #<Encoding:EUC-JP>
sym.id2name.encode(Encoding.default_external) #=> Encoding::ConverterNotFoundError
```
`Symbol#inspect` raise `Encoding::CompatibilityError` only in truffleruby, so I left `rescue` as is.
```
puts RUBY_ENGINE #=> truffleruby
'K'.force_encoding('UTF-7').to_sym.inspect #=> Encoding::CompatibilityError
```

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -226,7 +226,7 @@ module IRB
         else
           sym = $1
           candidates = Symbol.all_symbols.collect do |s|
-            ":" + s.id2name.encode(Encoding.default_external)
+            s.inspect
           rescue EncodingError
             # ignore
           end
--- a/test/irb/test_completion.rb
+++ b/test/irb/test_completion.rb
@@ -289,12 +289,14 @@ module TestIRB
     end
 
     def test_complete_symbol
-      %w"UTF-16LE UTF-7".each do |enc|
+      symbols = %w"UTF-16LE UTF-7".map do |enc|
         "K".force_encoding(enc).to_sym
       rescue
       end
-      _ = :aiueo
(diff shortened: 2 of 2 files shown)


hasumikin:
My IRB on TMUX on WSL2 works like this (I might be somehow wrong... 🤔 ):

https://user-images.githubusercontent.com/8454208/230240212-dfe022ac-ea14-4cd3-a703-e9c9dc74071b.mp4

ima1zumi:
@hasumikin 
Do you still reproduce it?
I couldn't reproduce the problem with your operation.

hasumikin:
@ima1zumi 
Now it's fixed!

@tompng
I'm closing this, thanks!

PR #541, opened 2023-03-13
Support inspecting BasicObject

### LINKED ISSUE #540: Fix basic object no method error
## Description

Hello. :wave: I'd like to be able to launch IRB and then create a basic object (i.e. `BasicObject.new`) without seeing the following error:

``` 
An error occurred when inspecting the object: #<NoMethodError: undefined method `is_a?' for #<BasicObject:0x000000010dd12a28>>
Result of Kernel#inspect: #<BasicObject:0x000000010dd12a28>
```

Here's a quick screencast as well:

https://user-images.githubusercontent.com/50245/224555000-c059fcdf-0030-4dbd-bed8-9c8bc37b4966.mp4

## Result of irb_info

``` 
Ruby version: 3.2.1
IRB version: irb 1.6.3 (2023-03-06)
InputMethod: RelineInputMethod with Reline 0.3.2 and /Users/bkuhlmann/.inputrc
.irbrc path: /Users/bkuhlmann/.irbrc
RUBY_PLATFORM: arm64-darwin22.3.0
LANG env: en_US.UTF-8
East Asian Ambiguous Width: 1
```

## Terminal Emulator

iTerm (but behavior is the same using macOS Terminal as well).

## Setting Files

I'm using a custom [~/.irbrc](https://github.com/bkuhlmann/dotfiles/blob/main/lib/templates/.irbrc.tt) but behavior is the same whether I use my configuration or not.

st0012:
When inspecting `BasicObject` in irb, the default inspector `pp` would raise an exception because it [calls `is_a?` on the object](https://github.com/ruby/pp/blob/master/lib/pp.rb#L188). We should help `pp` support `BasicObject` (if that's possible), but in the meantime we can also mitigate the problem from irb.

Closes #540

### DIFF
--- a/lib/irb/color_printer.rb
+++ b/lib/irb/color_printer.rb
@@ -4,6 +4,10 @@ require_relative 'color'
 
 module IRB
   class ColorPrinter < ::PP
+    METHOD_IS_A = Object.instance_method(:is_a?)
+    METHOD_RESPOND_TO = Object.instance_method(:respond_to?)
+    METHOD_INSPECT = Object.instance_method(:inspect)
+
     class << self
       def pp(obj, out = $>, width = screen_width)
         q = ColorPrinter.new(out, width)
@@ -22,9 +26,11 @@ module IRB
     end
 
     def pp(obj)
-      if obj.is_a?(String)
+      if METHOD_IS_A.bind(obj).call(String)
         # Avoid calling Ruby 2.4+ String#pretty_print that splits a string by "\n"
         text(obj.inspect)
+      elsif !METHOD_RESPOND_TO.bind(obj).call(:inspect)
+        text(METHOD_INSPECT.bind(obj).call)
       else
         super
       end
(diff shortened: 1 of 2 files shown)


lib/irb/color_printer.rb
```diff
@@ -4,6 +4,10 @@
 
 module IRB
   class ColorPrinter < ::PP
+    METHOD_IS_A = Object.instance_method(:is_a?)
+    METHOD_RESPOND_TO = Object.instance_method(:respond_to?)
+    METHOD_INSPECT = Object.instance_method(:inspect)
```
st0012, line 9: I know we already have store the method object in other places (e.g. `inspector.rb`). But I don't want to create dependencies between files just to reuse this as it's easy to find duplications later on.

tompng (approved):
```
irb(main):001:0> BasicObject.new
=> #<BasicObject:0x0000000103e59788>
irb(main):002:0> [BasicObject.new]
=> [#<BasicObject:0x0000000103eba038>]
```
:tada: Looks good to me :+1:

PR #552, opened 2023-03-30
Don't check RUBY_ENGINE when deciding whether to accept kwargs

st0012:
Ruby implementations like JRuby and TruffleRuby already indicate their compatibility target with RUBY_VERSION. We don't need to exclude them from accepting keyword arguments as long as they target 2.7+.

cc @nirvdrum


tompng (approved):
👍

PR #554, opened 2023-04-01
Remove dead code

st0012:
With @morriar's help, I was able to identify a few methods/modules/constants that haven't be referenced anymore, so I want to remove them in this PR. Each commit message explains when that piece of code stopped being referenced.


tompng (approved):
LGTM :+1:

st0012:
@tompng For PR like this, I think we should not use squash and merge. If there's any issue, reverting individual commits will be easier. Next time I'll document it in the PR description.

PR #555, opened 2023-04-02
Drop Ruby 2.6 support

st0012:
As Ruby 2.7 enters EOL, I think it makes sense to drop the support of 2.6, which lacks proper kwargs support and `Ripper#scan` and makes maintenance more complicated.

### Versioning implication

According to the [semver specificaion](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api), updating a dependency without changing any API does NOT need to bump a major version. And given we don't introduce a feature either, I think this change can safely be included in the next patch or minor release.

> **What should I do if I update my own dependencies without changing the public API?**

> That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts.

**Please let me merge this PR myself in case I found any additional changes to make or undo.**

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -58,19 +58,11 @@ module IRB
 
     BASIC_WORD_BREAK_CHARACTERS = " \t\n`><=;|&{("
 
-    def self.absolute_path?(p) # TODO Remove this method after 2.6 EOL.
-      if File.respond_to?(:absolute_path?)
-        File.absolute_path?(p)
-      else
-        File.absolute_path(p) == p
-      end
-    end
-
     GEM_PATHS =
       if defined?(Gem::Specification)
         Gem::Specification.latest_specs(true).map { |s|
           s.require_paths.map { |p|
-            if absolute_path?(p)
+            if File.absolute_path?(p)
               p
             else
               File.join(s.full_gem_path, p)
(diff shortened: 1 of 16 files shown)


lib/irb/completion.rb
```diff
@@ -58,19 +58,11 @@ def defined? do
...
     BASIC_WORD_BREAK_CHARACTERS = " \t\n`><=;|&{("
 
-    def self.absolute_path?(p) # TODO Remove this method after 2.6 EOL.
-      if File.respond_to?(:absolute_path?)
-        File.absolute_path?(p)
-      else
-        File.absolute_path(p) == p
-      end
-    end
-
```
tompng, line 68: I think we can remove similar code in `lib/irb/ext/loader.rb`

https://github.com/ruby/irb/blob/bf20faa4e6a93edb07923522700deb56f1873618/lib/irb/ext/loader.rb#L27-L48

st0012: Great catch 👍 I've removed it as well.

tompng (approved):
LGTM 👍

ima1zumi (approved):
LGTM ⭐

ima1zumi:
I added the Don’t merge label so that other than st0012 will not merge it.

st0012:
@ima1zumi Labeling is a great idea. Thank you 🙌

hasumikin (approved):

PR #558, opened 2023-04-06
Fix RubyLex's heredoc_with_hembexpr test to avoid ripper tokenizing i…

tompng:
Fix `test_heredoc_with_embexpr` because the expected result was wrong.
We also have to avoid ripper's bug https://bugs.ruby-lang.org/issues/19563 in test cases because I want this test to be executed in all supported ruby version.

```ruby
require 'ripper'
code = [
  %q(<<A+%W[#{<<B),
  %q(#{<<C+%W[),
  %q(a),
  %q(C),
  %q(a), # if this line is an empty string `%(),` then `Ripper.tokenize(code).join == code` returns false
  %q(]}),
  %q(}),
  %q(A),
  %q(B),
  %q(}),
  %q(]),
  %q()
].join("\n")
puts RUBY_VERSION
p Ripper.tokenize(code).join == code #=> true in ruby 2.7~3.2
```

### DIFF
--- a/test/irb/test_ruby_lex.rb
+++ b/test/irb/test_ruby_lex.rb
@@ -229,9 +229,9 @@ module TestIRB
       input_with_prompt = [
         PromptRow.new('001:0:":* ', %q(<<A+%W[#{<<B)),
         PromptRow.new('002:0:":* ', %q(#{<<C+%W[)),
-        PromptRow.new('003:0:":* ', %q()),
-        PromptRow.new('004:0:":* ', %q(C)),
-        PromptRow.new('005:0:]:* ', %q()),
+        PromptRow.new('003:0:":* ', %q(a)),
+        PromptRow.new('004:0:]:* ', %q(C)),
+        PromptRow.new('005:0:]:* ', %q(a)),
         PromptRow.new('006:0:":* ', %q(]})),
         PromptRow.new('007:0:":* ', %q(})),
         PromptRow.new('008:0:":* ', %q(A)),


test/irb/test_ruby_lex.rb
```diff
@@ -229,9 +229,9 @@ def test_heredoc_with_embexpr
       input_with_prompt = [
         PromptRow.new('001:0:":* ', %q(<<A+%W[#{<<B)),
         PromptRow.new('002:0:":* ', %q(#{<<C+%W[)),
-        PromptRow.new('003:0:":* ', %q()),
-        PromptRow.new('004:0:":* ', %q(C)),
```
st0012, line 233: If I understand it correctly, this prompt should be `]` if Ripper works as expected?

tompng: Yes. after `C`, heredoc `<<C` is closed and its inside `%W[` so the prompt should be `]`.

st0012: Thanks for explaining 👍 
Can we add a todo comment to add the this case back as well after the patch is merged?

tompng: If we change it back, I think test passes in ruby head but fails in ruby 2.7~3.2 (unless it's backported) so I want to let this test keep using `%q(a)`.

The reason why I add `%q()` here is to check the prompt between `%W[`, `heredoc_close`, `]`, so changing `%q()` to `%q(a)` still play a role. It does not have to be an empty string.
Test seems to properly fails before this fix #390.

st0012: Oh I thought that patch will be backported. But perhaps not to 2.7 anyway 🤔 . Yeah so ignore my comment then 😄

st0012 (approved):

PR #559, opened 2023-04-06
Simplify command method definition

st0012:
I took incremental steps on this and explained reasons in each commit. I think this simplification should help the command extension API plan and avoid the definition of confusing `#{cmd}_` methods (e.g. `irb_info_` after running `irb_info`).

### DIFF
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -258,15 +258,7 @@ module IRB # :nodoc:
       line = __LINE__; eval %[
         def #{cmd_name}(*opts, **kwargs, &b)
           Kernel.require_relative "#{load_file}"
-          line = __LINE__; eval %[
-            unless singleton_class.class_variable_defined?(:@@#{cmd_name}_)
-              singleton_class.class_variable_set(:@@#{cmd_name}_, true)
-              def self.#{cmd_name}_(*opts, **kwargs, &b)
-                ::IRB::ExtendCommand::#{cmd_class}.execute(irb_context, *opts, **kwargs, &b)
-              end
-            end
-          ], nil, __FILE__, line
-          __send__ :#{cmd_name}_, *opts, **kwargs, &b
+          ::IRB::ExtendCommand::#{cmd_class}.execute(irb_context, *opts, **kwargs, &b)
         end
       ], nil, __FILE__, line
 
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -246,7 +246,7 @@ module IRB # :nodoc:
     #
     # The optional +load_file+ parameter will be required within the method
     # definition.
-    def self.def_extend_command(cmd_name, cmd_class, load_file = nil, *aliases)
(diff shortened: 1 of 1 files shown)


lib/irb/extend-command.rb
```diff
@@ -255,33 +255,12 @@ def self.def_extend_command(cmd_name, cmd_class, load_file = nil, *aliases)
...
-        line = __LINE__; eval %[
-          def #{cmd_name}(*opts, &b)
-            ::IRB::ExtendCommand::#{cmd_class}.execute(irb_context, *opts, &b)
-          end
-        ], nil, __FILE__, line
-      end
+      line = __LINE__; eval %[
+        def #{cmd_name}(...)
+          Kernel.require_relative "#{load_file}"
+          ::IRB::ExtendCommand::#{cmd_class}.execute(irb_context, ...)
```
tompng, line 261: This code does not work in ruby < 2.7.2, (test passed because test runs 2.7.8)
https://bugs.ruby-lang.org/issues/16378#note-15
If this is a bug of ruby2.7.0 and 2.7.1, I think we can ignore it. But since this is something like a new feature, I want irb to work with 2.7.0 and 2.7.1. what do you think?

lib/irb/extend-command.rb
```diff
@@ -255,33 +255,12 @@ def self.def_extend_command(cmd_name, cmd_class, load_file = nil, *aliases)
         cmd_class = cmd_class.name
       end
 
-      if load_file
```
tompng, line 258: Since we assume load_file to be always present, I think we can change optional arg `load_file` to required.
```diff
- def self.def_extend_command(cmd_name, cmd_class, load_file = nil, *aliases)
+ def self.def_extend_command(cmd_name, cmd_class, load_file, *aliases)
```

lib/irb/extend-command.rb
st0012: I don't think we need to support those patch versions because although it's a new feature, there's no blocker for 2.7.1 users to upgrade to 2.7.2+.

lib/irb/extend-command.rb
st0012: Good catch!

lib/irb/extend-command.rb
st0012: But at the same time, it's not a big deal to drop this `...` change. So if you still feel it's better to avoid that, I'm fine with dropping it.

tompng: I agree that upgrading ruby is not a blocker, but I thought about it and I still feel better to avoid it.
We just dropped 2.6, and this will going to drop 2.7.0 and 2.7.1 so soon. (or maybe we can change the required ruby version in irb.gemspec to 2.7.2)

st0012: I see, I've dropped that commit 👍

tompng: Thanks 😄

tompng (approved):

PR #564, opened 2023-04-23
Simplify the help command's implementation

st0012:
The current method-redefining approach brings little benefit, makes it harder to understand the code, and causes warnings like:

> warning: method redefined; discarding old execute

This patch simplifies it while displaying more helpful message when rdoc couldn't be loaded.


tompng (approved):
Looks good 👍

PR #567, opened 2023-04-24
Print deprecation warning for `help` command

st0012:
As proposed in https://github.com/ruby/irb/issues/450, we should repurpose the `help` command for command help to align with other tools, like `ruby/debug`, `pry`, and `byebug`. But before we make the switch, we should start printing deprecation warning for the change.

### Example

```
irb(main):001:0> help String#gsub
[Deprecation] The `help` command will be repurposed to show command help in the future.
For RI document look up, please use `show_doc` command instead.
For command help, please use `show_cmds` for now.
```

### `show_cmds` output

```
Context
  help           [DEPRECATED] Enter the mode to look up RI documents.
  show_doc       Enter the mode to look up RI documents.
```

### DIFF
--- a/lib/irb/cmd/help.rb
+++ b/lib/irb/cmd/help.rb
@@ -5,6 +5,19 @@ require_relative "show_doc"
 module IRB
   module ExtendCommand
     class Help < ShowDoc
+      category "Context"
+      description "[DEPRECATED] Enter the mode to look up RI documents."
+
+      DEPRECATION_MESSAGE = <<~MSG
+        [Deprecation] The `help` command will be repurposed to display command help in the future.
+        For RI document lookup, please use the `show_doc` command instead.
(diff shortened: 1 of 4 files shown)


lib/irb/cmd/help.rb
```diff
@@ -1,55 +1,22 @@
...
-          names.each do |name|
-            begin
-              Ri.display_name(name.to_s)
-            rescue RDoc::RI::Error
-              puts $!.message
-            end
-          end
-        end
+      DEPRECATION_MESSAGE = <<~MSG
+        [Deprecation] The `help` command will be repurposed to show command help in the future.
```
tompng, line 12: Looks good!
Just an idea, I think it would be great if there's a message to indicate that the command `to show command help` is `show_cmds`?

PR #568, opened 2023-04-25
Fix Locale's encoding lookup for ujis/euc encodings

st0012:
In https://github.com/ruby/irb/commit/3ee79e89adb8e21b63d796e53bcc86281685076d, `encoding_aliases.rb` was introduced to return the correct encoding object for `ujis` and `euc` encodings.

However, the return value of `@@legacy_encoding_alias_map[@encoding_name]` is always overridden by a second look up with `Encoding.find(@encoding_name)`. So the logic didn't work as expected.


tompng (approved):
:+1:
It's great that the added test found this :tada:

PR #570, opened 2023-04-26
Add tests for Locale#find and Locale#load

st0012:
These are the only `Locale` APIs that are still being used. I'll do some refactor once the these tests are merged.

### DIFF
--- a/test/irb/test_locale.rb
+++ b/test/irb/test_locale.rb
@@ -79,5 +79,40 @@ module TestIRB
         end
       end
     end
+
+    def test_load
+      # reset Locale's internal cache
+      IRB::Locale.class_variable_set(:@@loaded, [])
+      # Because error.rb files define the same class, loading them causes method redefinition warnings.
+      original_verbose = $VERBOSE
+      $VERBOSE = nil
+
+      jp_local = IRB::Locale.new("ja_JP.UTF-8")
+      jp_local.load("irb/error.rb")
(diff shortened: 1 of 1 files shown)


test/irb/test_locale.rb
```diff
@@ -79,5 +79,40 @@ def test_initialize_with_legacy_ja_encoding_euc
...
+      # reset Locale's internal cache
+      IRB::Locale.class_variable_set(:@@loaded, [])
+
+      en_local = IRB::Locale.new("en_US.UTF-8")
+      en_local.load("irb/error.rb")
+      msg = IRB::CantReturnToNormalMode.new.message
+      assert_equal("Can't return to normal mode.", msg)
+    ensure
+      # before turning warnings back on, load the error.rb file again to avoid warnings in other tests
+      load "irb/lc/error.rb"
```
tompng, line 104: It looks like this is not avoiding wraning in my LOCALE=`ja_JP.UTF-8`

```sh
% LANG='en_US.UTF-8' rake test
pass, no warning
% LANG='ja_JP.UTF-8' rake test
pass, warnings
```

How about loading (and also set as loaded in class_variable) with default locale by
```ruby
IRB::Locale.new.load("irb/error.rb")
```

st0012: I needed to use `IRB_LANG` to reproduce the warning. But yeah using `IRB.Locale` fixes it, great catch 👍

tompng (approved):

PR #571, opened 2023-04-26
Simplify Locale#load

st0012:
Instead of loading file content with `MagicFile` and then evaluating it, we can just use `Kernel.load` to load the file.

I also removed unused `Locale#require` and `Locale#toplevel_load` methods and removed unused `Locale#load` argument.

### DIFF
--- a/lib/irb/locale.rb
+++ b/lib/irb/locale.rb
@@ -109,7 +109,7 @@ module IRB # :nodoc:
 
     alias toplevel_load load
 
-    def load(file, priv=nil)
+    def load(file)
       found = find(file)
       if found
         unless @@loaded.include?(found)
--- a/lib/irb/locale.rb
+++ b/lib/irb/locale.rb
@@ -114,7 +114,7 @@ module IRB # :nodoc:
       if found
         unless @@loaded.include?(found)
           @@loaded << found # cache
-          return real_load(found, priv)
+          Kernel.load(found)
         end
       else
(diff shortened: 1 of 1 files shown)


lib/irb/locale.rb
```diff
@@ -114,7 +114,7 @@ def load(file, priv=nil)
       if found
         unless @@loaded.include?(found)
           @@loaded << found # cache
```
st0012, line 116: I don't think we need to keep this "cache" for performance reason. But removing it would cause a lot of redefinition warning from loading `error.rb`. So I decided to keep it for now.

lib/irb/locale.rb
```diff
@@ -109,12 +109,12 @@ def require(file, priv = nil)
 
     alias toplevel_load load
 
-    def load(file, priv=nil)
+    def load(file)
```
tompng, line 112: It looks good :+1:, but one thing, priv seems to be used in line 92.
```ruby
IRB::Locale.new.require 'irb/error.rb'
# => /Users/tomoya/Documents/github/ruby/irb/lib/irb/locale.rb:112:in `load': wrong number of arguments (given 2, expected 1) (ArgumentError)
```
I think `def require(file, priv = nil)` is not used, so it won't be a problem, but how about removing `priv=nil` from method `require`, or remove `require` method

st0012: Yeah let's remove the `require` too.

tompng (approved):

PR #574, opened 2023-04-27
Retire magic-file.rb

st0012:
`MagicFile` was introduced around v0.9.6, which was like 14~15 years ago. It was needed because back then we needed to read a file's magic comment to determine the encoding of it, and read it with that encoding.

Commit: 3ee79e89

But now we expect files to be encoded in UTF-8 and don't specify encoding through magic comments anymore, `MagicFile` can be retired.


tompng (approved):
LGTM :tada:

eregon:
Nice!

Issue #513, opened 2023-01-31
Support command extension APIs


st0012:
## Description

Since version 1.6, IRB's built-in commands now have `category` and `description` attributes and can be listed in the `show_cmds` command's output. This greatly improved feature discovery and made IRB easier to use.

However, because IRB doesn't have official APIs for custom commands, they are currently defined as context methods. This means those commands won't be listed in the `show_cmds` output.

We should start thinking about command extension APIs to increase custom commands' discoverability as well.

### Details

I want the command definition and related APIs to be similar to Pry's, as I believe its design is intuitive and this could help Pry extension authors get familiar with this new spec more easily.

#### Command Definition

```rb
module IRB
  module Command
    class MyCommand < Command::Base
      category "my_category"
      description "This is my command for xyz"
      aliases ["foo", "bar"]

      def execute(arg_string)
      end
    end
  end
end
```

Some notes:

- `IRB::Command` should be a better namespace than `IRB::ExtendCommand`
- Similarly, `IRB::Command::Base` should be better than `IRB::ExtendCommand::Nop`
- To simplify command handling on IRB part, I prefer directly passing the arg string to the command and let command author do the rest.
- Command name will be inferred from the class's name. In this case, it'll be `my_command`.

#### Command Registration

```rb
IRB::ExtendedCommandBundle.add_command(IRB::Command::MyCommand)
```

- Letting users register with the command class is way easier than asking them to figure out how to load the command lazily. It also simplifies IRB's implementation.
- If we register the command this way, we can also just let users define aliases in the command class.
- While I do worry a bit about loading commands eagerly would slow down IRB boot time, I don't have any actual data for it. Perhaps it's not a valid concern after all.

#### Command Invocation

- `my_command arg1 arg2` will call `MyCommand#execute` with `arg_string = "arg1 arg2"`
    - `foo arg1 arg2` and `bar arg1 arg2` will work this way too
- `my_command(arg1, arg2)` will call the `my_command` method on the current object.
    - **Commands can't be used as methods**

### Possible First Adopters

- [Rails](https://github.com/rails/rails)
    - ~~[Current commands](https://github.com/rails/rails/blob/main/railties/lib/rails/console/app.rb)~~ (they should be considered helpers so #588 should be a better option)
    - We can also port `rails routes` and `rails middlewares` using the new APIs
- [irbtools](https://github.com/janlelis/irbtools)
[...]

PR #575, opened 2023-04-30
Support seamless integration with ruby/debug

### LINKED ISSUE #521: Allow executing debugging commands from irb session
## Description

Currently, `irb` supports 10 debugging commands that are powered by `debug`, which is a huge first-step improvement.

However, when users run those commands, they'll enter a new `debug` session and leave `irb`. This could be inconvenient because:

- `debug` doesn't support multi-line mode
- `debug` doesn't have autocompletion yet
- `irb` and `debug` has different commands or behaviour
- Users will lose local variables defined in the current `irb` session

So it'd be amazing if we can let users run the same debugging commands, while letting them stay inside the current `irb` session.

st0012:
## Summary

This PR enhances the debugging experience by keeping users within the IRB session when executing debugging commands. This ensures users retain multi-line input support and access to all IRB commands, while benefiting from `debug`'s enhanced debugging functionalities.

Simply put: this will be `pry-byebug` equivalent for IRB and `debug`.

## Demonstrations

Current Debugging Experience:

https://github.com/ruby/irb/assets/5079556/05932805-f27a-4d54-adbf-76a36f707c81

Improved Debugging Experience:

https://github.com/ruby/irb/assets/5079556/61b3ce4f-556a-406b-b83f-a9fadac2dc44

## Design & Implementation

- Debugging commands execution is handled by `ExtendCommand::Debug#execute`, which performs several actions:
    1. Checks if the debugging command originates from a `binding.irb` call.
    2. Sets up the debug gem if the above condition is met.
    3. Inserts a debug breakpoint at `Irb#debug_break` with the intended command.
    4. Exits the current `Irb#run` call via `throw :IRB_EXIT`.

    When `Irb#debug_break` is automatically called after step `iv`, the breakpoint triggers, executing the intended command.

- The new `IRB::Debug::UI` class and `Irb#debug_readline` method transform IRB into the debugger session's UI. Depending on the user's input, the behavior varies:
    1. Debugging commands (e.g., `step 3`) are passed directly to the debugger.
    2. IRB commands (e.g., `show_source Foo#bar`) are evaluated within the IRB session.
    3. Other inputs (e.g., `my_object.my_method`) are passed to the debugger.

    After the debugger evaluates the input, it calls `Irb#debug_readline` again, maintaining the impression of a continuous IRB session.

## Known Limitations

- Local variables assigned in the original IRB session are not accessible in the irb:rdbg session.
- The `_` variable is not functional until `debug` supports it.

Resolves #521

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -18,6 +18,7 @@ require_relative "irb/color"
 
 require_relative "irb/version"
 require_relative "irb/easter-egg"
+require_relative "irb/debug"
 
 # IRB stands for "interactive Ruby" and is a tool to interactively execute Ruby
 # expressions read from the standard input.
@@ -373,8 +374,6 @@ module IRB
   class Abort < Exception;end
 
   @CONF = {}
-
-
   # Displays current configuration.
   #
   # Modifying the configuration is achieved by sending a message to IRB.conf.
@@ -441,7 +440,7 @@ module IRB
     # Creates a new irb session
     def initialize(workspace = nil, input_method = nil)
       @context = Context.new(self, workspace, input_method)
-      @context.main.extend ExtendCommandBundle
+      @context.workspace.load_commands_to_main
       @signal_status = :IN_IRB
       @scanner = RubyLex.new(@context)
     end
@@ -455,6 +454,38 @@ module IRB
         # and remove the redundant method
         DEBUGGER__.singleton_class.send(:undef_method, :capture_frames_without_irb)
       end
+    end
+
+    def debug_readline(binding)
+      workspace = IRB::WorkSpace.new(binding)
+      context.workspace = workspace
+      context.workspace.load_commands_to_main
+      scanner.increase_line_no(1)
+
+      # When users run:
+      # 1. Debugging commands, like `step 2`
+      # 2. Any input that's not irb-command, like `foo = 123`
+      #
+      # Irb#eval_input will simply return the input, and we need to pass it to the debugger.
+      input = if IRB.conf[:SAVE_HISTORY] && context.io.support_history_saving?
+        # Previous IRB session's history has been saved when `Irb#run` is exited
+        # We need to make sure the saved history is not saved again by reseting the counter
+        context.io.reset_history_counter
+
+        begin
+          eval_input
+        ensure
+          context.io.save_history
+        end
+      else
+        eval_input
+      end
+
+      if input&.include?("\n")
+        scanner.increase_line_no(input.count("\n") - 1)
+      end
+
+      input
     end
 
     def run(conf = IRB.conf)
@@ -542,6 +573,18 @@ module IRB
       @scanner.each_top_level_statement do |line, line_no, is_assignment|
         signal_status(:IN_EVAL) do
           begin
+            # If the integration with debugger is activated, we need to handle certain input differently
+            if @context.with_debugger
+              command_class = load_command_class(line)
+              # First, let's pass debugging command's input to debugger
+              # Secondly, we need to let debugger evaluate non-command input
+              # Otherwise, the expression will be evaluated in the debugger's main session thread
+              # This is the only way to run the user's program in the expected thread
+              if !command_class || ExtendCommand::DebugCommand > command_class
+                return line
+              end
+            end
+
             evaluate_line(line, line_no)
 
             # Don't echo if the line ends with a semicolon
@@ -579,6 +622,12 @@ module IRB
       end
 
       @context.evaluate(line, line_no)
+    end
+
+    def load_command_class(line)
+      command, _ = line.split(/\s/, 2)
+      command_name = @context.command_aliases[command.to_sym]
+      ExtendCommandBundle.load_command(command_name || command)
     end
 
     def convert_invalid_byte_sequence(str, enc)
@@ -934,12 +983,32 @@ class Binding
   #
   # See IRB@Usage for more information.
   def irb(show_code: true)
+    # Setup IRB with the current file's path and no command line arguments
     IRB.setup(source_location[0], argv: [])
+    # Create a new workspace using the current binding
     workspace = IRB::WorkSpace.new(self)
+    # Print the code around the binding if show_code is true
     STDOUT.print(workspace.code_around_binding) if show_code
-    binding_irb = IRB::Irb.new(workspace)
-    binding_irb.context.irb_path = File.expand_path(source_location[0])
-    binding_irb.run(IRB.conf)
-    binding_irb.debug_break
+    # Get the original IRB instance
+    debugger_irb = IRB.instance_variable_get(:@debugger_irb)
+
+    irb_path = File.expand_path(source_location[0])
+
+    if debugger_irb
+      # If we're already in a debugger session, set the workspace and irb_path for the original IRB instance
+      debugger_irb.context.workspace = workspace
+      debugger_irb.context.irb_path = irb_path
+      # If we've started a debugger session and hit another binding.irb, we don't want to start an IRB session
+      # instead, we want to resume the irb:rdbg session.
+      IRB::Debug.setup(debugger_irb)
+      IRB::Debug.insert_debug_break
+      debugger_irb.debug_break
+    else
+      # If we're not in a debugger session, create a new IRB instance with the current workspace
+      binding_irb = IRB::Irb.new(workspace)
+      binding_irb.context.irb_path = irb_path
+      binding_irb.run(IRB.conf)
+      binding_irb.debug_break
+    end
   end
 end
--- a/lib/irb/cmd/subirb.rb
+++ b/lib/irb/cmd/subirb.rb
@@ -11,8 +11,7 @@ module IRB
 
   module ExtendCommand
     class MultiIRBCommand < Nop
-      def initialize(conf)
-        super
+      def execute(*args)
         extend_irb_context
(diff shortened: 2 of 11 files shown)


lib/irb/debug/ui.rb
```diff
@@ -0,0 +1,106 @@
...
+      end
+
+      def readline _
+        setup_interrupt do
+          tc = DEBUGGER__::SESSION.get_thread_client(@thread)
+          cmd = IRB.debug_readline(binding: tc.current_frame.binding || TOPLEVEL_BINDING, scanner: @scanner, prompt_name: PROMPT_NAME)
+
+          case cmd
+          when nil # when user types C-d
+            "exit!"
```
tompng, line 66: In normal irb, `C-d` and `exit` will exit irb session, not the program.
Maybe `continue` instead of `exit!`?
Also, is it possible to quit debug mode and go back to normal irb mode with `continue`?

st0012: > Maybe continue instead of exit!?

Good point! I'll change it.

> is it possible to quit debug mode and go back to normal irb mode with continue?

Can you describe a use case for that? Since `irb:rdbg` is basically extending IRB session with debugging functionality, why do we want to roll it back?

tompng: > use case

Sorry, I don't have it. I misunderstood that `continue` ends debugging and want irb to go back to normal mode. actually it continues the execution and stops at the next breakpoint.

lib/irb/context.rb
```diff
@@ -490,11 +494,28 @@ def evaluate(line, line_no, exception: nil) # :nodoc:
 
       # Hook command-specific transformation
       command_class = ExtendCommandBundle.load_command(command)
+
+      # If the debugger is activated, we can directly pass debugging command's input
+      if with_debugger && command_class&.ancestors&.include?(ExtendCommand::DebugCommand)
+        throw :IRB_EXIT, original_line
```
tompng, line 500: I think the implementation is a little hacky that `evaluate` does not eval but throw in debug mode, `eval_input`(normally runs IRB's main loop) just return single input.

My idea is doing one of these
- Refactor `each_top_level_statement` and `evaluate`
- Use Fiber to alternately switch to IRB::Debug::UI and IRB's `each_top_level_statement`
- Don't use Debug::UI_Base, store debug session in IRB::Context and implement as ExtendCommand

Refactor like
```ruby
class IRB::Context
  def evaluate(line)
    type, input_or_command = check_input_is_command_or_not(line)
    type == :command ? run_command(input_or_command) : eval(input_or_command)
  end
end
class IRB
  def eval_input
    setup_scanner
    each_top_level_statement{|line|context.evaluate line}
  end
  def debug_read
    setup_scanner
    input = each_top_level_statement{break line}
    context.check_input_is_command_or_not(input)
  end
end
```

Use Fiber like
```ruby
class IRB::Context
  def evaluate(...)
    ...
    if with_debugger
      debugger_ui_fiber.resume original_line
    end
  end
end
class IRB::Debug::UI
  def readline(...)
    debugger_ui_fiber.yield
  end
end
```

st0012: > `type == :command ? run_command(input_or_command) : eval(input_or_command)`

This is what I want to achieve with #513. Once we can easily distinguish command vs Ruby expression, `Context#evaluate` can really just evaluate the latter and don't care about commands, including debugging commands.

FWIW, I think the current `Context#evaluate` has become `Context#process_line` even before this PR. And IMO the right solution is a big refactor along with #513.

To be clear, I'm by no means saying the current implementation isn't a bit messy. It surely is. But I think it's worth thinking it from some slightly different angles:
- Will the current implementation block future command refactoring?
- Can the implementation become much cleaner with the command refactoring?

I personally think the answers are: No and Yes. Therefore it should be ok to ship it first and refactor later, assume everything behaves as expected.

That being said, it's also worth thinking that if we should implement #513 first, and implement this feature on top of it. This has a few benefits:

- The implementation should be simpler, at least around the `Context#evaluate` part
- We can use this big feature as the attraction for upgrading to v2.0.0

The downsides are:

- We won't be taking such integration into account when implementing 513
- This feature will be shipped and tested much later

@hasumikin @ima1zumi what do you think about this?

hasumikin: I feel... regarding this PR, implementing this feature would be good first.

Generally, I think that:
- First, implement a feature if you can make sure it is really necessary and useful
- However, do not leave TOO BIG refactorings in the future
In other words, introducing a feature and refactoring walk in step with each other.

At the same time, @tompng's idea itself looks good to me:

>Refactor each_top_level_statement and evaluate

st0012: Thanks for the feedback 👍 
While we can't clearly separate command/Ruby expression evaluation atm, I think we can move non-evaluation logic out of `Context#evaluate` first to make things cleaner. So I opened #615 for it.

st0012: I won't be able to achieve `type == :command ? run_command(input_or_command) : eval(input_or_command)` yet because we can't separate command and expression evaluation atm. But with #618 I will be able to drop `abort`.

lib/irb.rb
```diff
@@ -560,6 +595,19 @@ def eval_input
...
+              # First, let's pass debugging command's input to debugger
+              # Secondly, we need to let debugger evaluate non-command input
+              # Otherwise, the expression will be evaluated in the debugger's main session thread
+              # This is the only way to run the user's program in the expected thread
+              if !command_class || command_class.ancestors.include?(ExtendCommand::DebugCommand)
+                return line
+              end
+            end
+
             evaluate_line(line, line_no)
```
tompng, line 611: After evaluating line, the context may be changed to debug mode.

With this code
```ruby
class Foo
  def bar()
    binding.irb
  end
  def to_s() = 'foo'
end
binding.irb
```

```ruby
From: a.rb @ line 7 :

    2:   def bar()
    3:     binding.irb
    4:   end
    5:   def to_s() = 'foo'
    6: end
 => 7: binding.irb
irb(main):001:0> Foo.new.bar

From: a.rb @ line 3 :

    1: class Foo
    2:   def bar()
 => 3:     binding.irb
    4:   end
    5:   def to_s() = 'foo'
    6: end
    7: binding.irb
irb(foo):001:0> debug         # Enter debug mode.
irb:rdbg(foo):002:0> continue # Goes back to line 7. Context has changed to debug mode.
=> #<Class:DEBUGGER__>
irb(main):002:0>              # Still in debug mode, but prompt looks not a debug mode.
irb(main):003:0> whereami     # Location is wrong.

From: a.rb @ line 3 :

    1: class Foo
    2:   def bar()
 => 3:     binding.irb
    4:   end
    5:   def to_s() = 'foo'
    6: end
    7: binding.irb
=> nil
irb(main):004:0> Foo.new.bar

From: a.rb @ line 3 :

    1: class Foo
    2:   def bar()
 => 3:     binding.irb
    4:   end
    5:   def to_s() = 'foo'
    6: end
    7: binding.irb
irb:rdbg(foo):003:0>          # Reached to next binding.irb. Prompt shows that it's in debug mode.
```


I think we need something like this after `evaluate_line`.
```ruby
if !@context.with_debugger && defined?(DEBUGGER__::SESSION)
  debug setup and throw :IRB_EXIT
end
```

st0012: Thanks for catching this case 👍 
I think the root problem is that IRB doesn't support nested sessions as described in https://github.com/ruby/irb/issues/650 so I plan to address it separately.

lib/irb.rb
```diff
@@ -560,6 +595,19 @@ def eval_input
...
             is_assignment = assignment_expression?(line)
+
+            # If the integration with debugger is activated, we need to handle certain input differently
+            if @context.with_debugger
+              command_class = load_command_class(line)
+              # First, let's pass debugging command's input to debugger
+              # Secondly, we need to let debugger evaluate non-command input
+              # Otherwise, the expression will be evaluated in the debugger's main session thread
+              # This is the only way to run the user's program in the expected thread
+              if !command_class || command_class.ancestors.include?(ExtendCommand::DebugCommand)
```
vinistock, line 606: Instead of checking the ancestors array, we can check if a class inherits from another directly.

**Note**: normally, you'd do `some_class < OtherClass`. From our learnings maintaining Tapioca, that sometimes fails for certain classes because they override `<`. So here the suggestion is inverted doing `OtherClass > some_class`, which returns the exact same result.

```suggestion
              if !command_class || ExtendCommand::DebugCommand > command_class
```

lib/irb/cmd/subirb.rb
```diff
@@ -29,6 +28,12 @@ class IrbCommand < MultiIRBCommand
       description "Start a child IRB."
 
       def execute(*obj)
+        if irb_context.with_debugger
```
vinistock, line 31: Is it possible to put this in the parent class and invoke super to avoid duplicating the message? Or does that not make sense?

lib/irb.rb
st0012: This is nice, thanks 👍

lib/irb/cmd/subirb.rb
st0012: Updated 👍

st0012:
@tompng would you mind giving it another look when have time? Thx

tompng (approved):
LGTM :+1:
IRB is completely a great debugger now :tada:

PR #576, opened 2023-04-30
Simplify each_top_level_statement

tompng:
This pull request is an extract of refactor from #500 

Refactored each_top_level_statement
- Remove instance variables `@exp_line_no` `@indent` `@continue` `@line` from `ruby-lex.rb`
- Extract reading line part from `def each_top_level_statement` to `def readmultiline`
- Refactor complex flow using `catch(:TERM_INPUT)` and `throw :TERM_INPUT` into simple loop and break
- Delete `def lex` because it is only used from test code now

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -225,7 +225,9 @@ class RubyLex
     line_offset = 0
     loop do
       line = @input.call
-      return code.empty? ? nil : code unless line
+      unless line
+        return code.empty? ? nil : code
+      end
 
       code << line
       # Accept any single-line input for symbol aliases or commands that transform args
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -18,10 +18,7 @@ class RubyLex
 
   def initialize(context)
     @context = context
-    @exp_line_no = @line_no = 1
-    @indent = 0
-    @continue = false
-    @line = ""
+    @line_no = 1
     @prompt = nil
   end
 
@@ -40,6 +37,11 @@ class RubyLex
       result = yield code, line_no
     end
     result
+  end
+
+  def single_line_command?(code)
+    command = code.split(/\s/, 2).first
+    @context.symbol_alias?(command) || @context.transform_args?(command)
   end
 
   # io functions
@@ -65,10 +67,7 @@ class RubyLex
           end
         else
           # Accept any single-line input for symbol aliases or commands that transform args
-          command = code.split(/\s/, 2).first
-          if @context.symbol_alias?(command) || @context.transform_args?(command)
-            next true
-          end
+          next true if single_line_command?(code)
 
           code.gsub!(/\s*\z/, '').concat("\n")
           tokens = self.class.ripper_lex_without_warning(code, context: @context)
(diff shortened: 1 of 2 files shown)


lib/irb/ruby-lex.rb
```diff
@@ -214,67 +213,50 @@ def check_state(code, tokens)
     [ltype, indent, continue, code_block_open]
   end
 
-  def prompt
-    if @prompt
-      @prompt.call(@ltype, @indent, @continue, @line_no)
-    end
+  def save_prompt_to_context_io(ltype, indent, continue, line_num_offset)
+    # Implicitly saves prompt string to `@context.io.prompt`. This will be used in the next `@input.call`.
+    @prompt.call(ltype, indent, continue, @line_no + line_num_offset)
```
st0012, line 218: `@prompt` is not going to be `nil` in actual use cases because `Irb#eval_input` always calls `RubyLex#set_prompt` before calling `RubyLex#each_top_level_statemen`.
But I think this type of implicit dependency is risky and we should probably refactor that later.

lib/irb/ruby-lex.rb
```diff
@@ -214,67 +213,50 @@ def check_state(code, tokens)
...
-          @exp_line_no = @line_no
-
-          @indent = 0
-        rescue TerminateLineInput
-          initialize_input
-          prompt
-        end
-      end
+    # multiline
+    return @input.call if @io.respond_to?(:check_termination)
```
st0012, line 225: I think `check_termination` is a `RelineInputMethod`-specific method? If so, can we just check if `@io.is_a?(RelineInputMethod)`?

lib/irb/ruby-lex.rb
```diff
@@ -214,67 +213,50 @@ def check_state(code, tokens)
...
+
+    # nomultiline
+    code = ''
+    line_offset = 0
+    loop do
+      line = @input.call
+      return code.empty? ? nil : code unless line
```
st0012, line 232: My brain can't handle ternary condition + unless at the same line. So I'd appreciate if we can expand it a bit

```suggestion
      unless line
        return code.empty? ? nil : code
      end
```

lib/irb/ruby-lex.rb
```diff
@@ -214,67 +213,50 @@ def check_state(code, tokens)
...
+
+      code << line
+      # Accept any single-line input for symbol aliases or commands that transform args
+      return code if single_line_command?(code)
+
+      check_target_code = code.gsub(/\s*\z/, '').concat("\n")
+      tokens = self.class.ripper_lex_without_warning(check_target_code, context: @context)
+      ltype, indent, continue, code_block_open = check_state(check_target_code, tokens)
```
st0012, line 240: I feel that we can extract from `check_target_code` to this line into a method like `check_code_state`. And then we can align test and the actual implementation a bit more?

test/irb/test_ruby_lex.rb
```diff
@@ -83,27 +83,24 @@ def assert_row_indenting(lines, row)
...
       ruby_lex = RubyLex.new(context)
-
-      io = proc{ lines.join("\n") }
-      ruby_lex.set_input(io) do
-        lines.join("\n")
-      end
-      ruby_lex.lex
-      ruby_lex
+      ltype, indent, _continue, code_block_open = ruby_lex.check_state(code, tokens)
+      [ltype, indent, code_block_open]
```
st0012, line 103: `ltype` doesn't seem to be used by callers?

lib/irb/ruby-lex.rb
st0012: I think checking by class name will couple IRB's input method logic with RubyLex, which could be used independently by `irb/color`. Let's revisit this later with the changes in #583

lib/irb/ruby-lex.rb
tompng: good catch :+1: I added `def check_code_state`

st0012 (approved):
LGTM 👍 But I think there's a conflict in test file.

st0012 (approved):

PR #582, opened 2023-05-16
Fix Test timedout in test_debug_cmd

tompng:
### Description

In ruby head, `readline-ext` is removed. `Readline.readline` will be `Reline.readline` instead of `readline.so`.

The problem is that Reline prints `"▽\e[6n"`, east-asian-amibuous character and device status report escape sequence.
Reline will wait for cursor position report `\e[<row>;<col>R` forever and cause timeout error.
[Timeout result in #581](https://github.com/ruby/irb/actions/runs/4984571776/jobs/8929008593?pr=581)
To suppress this, I override `Reline::IOGate.cursor_pos` to use `Reline::GeneralIO.cursor_pos` in test_debug_cmd

### Failing test (help wanted)

Still one test is failing in ruby head.
#### Failed assertion
```ruby

assert_match(/\(rdbg\) next/, output)
```
#### Actual output
```
From: /tmp/irb-20230516-1760-qqd7e7.rb @ line 1 :

 => 1: binding.irb
    2: puts "hello"

debug
next
continue
(rdbg) [1, 2] in /tmp/irb-20230516-1760-qqd7e7.rb
     1| binding.irb
=>   2| puts "hello"
=>#0	<main> at /tmp/irb-20230516-1760-qqd7e7.rb:2
(rdbg) hello
```
This is because ruby/debug is directly requiring `readline.so` and fallbacks to `print prompt; gets`.
https://github.com/ruby/debug/blob/4ec9d7ab46df09aa87d64e600df8805a3ee0314c/lib/debug/console.rb#L164-L168

### DIFF
--- a/test/irb/test_debug_cmd.rb
+++ b/test/irb/test_debug_cmd.rb
@@ -211,7 +211,15 @@ module TestIRB
       cmd = [EnvUtil.rubybin, "-I", LIB, @ruby_file.to_path]
       tmp_dir = Dir.mktmpdir
       rc_file = File.open(File.join(tmp_dir, ".irbrc"), "w+")
-      rc_file.write("IRB.conf[:USE_SINGLELINE] = true")
+      rc_file.write(<<~RUBY)
+        IRB.conf[:USE_SINGLELINE] = true
+        require 'readline'
+        if Readline.const_defined?(:IOGate)
+          def (Reline::IOGate).cursor_pos
+            Reline::GeneralIO.cursor_pos
+          end
+        end
+      RUBY
       rc_file.close
 
       @commands = []
--- a/test/irb/test_debug_cmd.rb
+++ b/test/irb/test_debug_cmd.rb
@@ -17,7 +17,7 @@ module TestIRB
 
   class DebugCommandTestCase < TestCase
     IRB_AND_DEBUGGER_OPTIONS = {
-      "RUBY_DEBUG_NO_RELINE" => "true", "NO_COLOR" => "true", "RUBY_DEBUG_HISTORY_FILE" => ''
+      "NO_COLOR" => "true", "RUBY_DEBUG_HISTORY_FILE" => ''
     }
 
     def setup
@@ -213,12 +213,7 @@ module TestIRB
       rc_file = File.open(File.join(tmp_dir, ".irbrc"), "w+")
       rc_file.write(<<~RUBY)
         IRB.conf[:USE_SINGLELINE] = true
-        require 'readline'
-        if Readline.const_defined?(:IOGate)
-          def (Reline::IOGate).cursor_pos
-            Reline::GeneralIO.cursor_pos
-          end
-        end
+        Reline.const_set(:IOGate, Reline::GeneralIO)
       RUBY
       rc_file.close
 


st0012:
@tompng I think if you also apply this change it'll pass:

```diff
diff --git a/test/irb/test_debug_cmd.rb b/test/irb/test_debug_cmd.rb
index 309e241..1cf6066 100644
--- a/test/irb/test_debug_cmd.rb
+++ b/test/irb/test_debug_cmd.rb
@@ -17,7 +17,7 @@ module TestIRB
 
   class DebugCommandTestCase < TestCase
     IRB_AND_DEBUGGER_OPTIONS = {
-      "RUBY_DEBUG_NO_RELINE" => "true", "NO_COLOR" => "true", "RUBY_DEBUG_HISTORY_FILE" => ''
+      "NO_COLOR" => "true", "RUBY_DEBUG_HISTORY_FILE" => ''
     }
```

A few questions:

1. What does this do? 
    > Reline prints "▽\e[6n", east-asian-amibuous character and device status report escape sequence.
    > Reline will wait for cursor position report \e[<row>;<col>R forever and cause timeout error.
2. Should `Reline.readline` behave this way? 
3. I found forcing `Reline::IOGate` to be `GeneralIO` also prevents the issue:

    ```diff
    @@ -214,11 +214,7 @@ module TestIRB
           rc_file.write(<<~RUBY)
             IRB.conf[:USE_SINGLELINE] = true
             require 'readline'
    -        if Readline.const_defined?(:IOGate)
    -          def (Reline::IOGate).cursor_pos
    -            Reline::GeneralIO.cursor_pos
    -          end
    -        end
    +        Reline.const_set(:IOGate, Reline::GeneralIO)
           RUBY
           rc_file.close
    ```
    If the answer of 2) is yes, perhaps we can make `Reline`'s `IOGate` configurable so it works in testing environment too? `ruby/debug` could benefit from this.

tompng:
> What does this do?

It prints east asian ambiguous character, request current cursor position by printing `"\e[6n"`, get the actual cursor position escape sequence from STDIN, and determine the actual east asian ambiguous character width.
User can set east asian ambiguous width to either half-width or full-width in preference panel of terminal emulator app. Reline should know this setting value.

> Should Reline.readline behave this way?

I think yes. By doing it, reline can handle east asian ambiguous chars correctly.
On the other hand, readline does not do this. It assumes ambiguous char is half-width and it fails to render in full-width setting.
Image of readline fails handling full-width

If STDIN and STDOUT is a TTY, reline is assuming it is run inside a terminal that answers to `"\e[6n"`. I think this assumption is correct and what is wrong is that the test is using pty but does not act like a real terminal that reply to `"\e[6n"`.

> 3. `Reline::IOGate to be GeneralIO`

Yes, it is one of the solution. Making configurable seems good.
We can also force reline to use GeneralIO by not using `PTY.spawn` but use `IO.popen`.
Unfortunately, I tried `IO.popen` but there was a problem of buffering.
```ruby
IO.popen ['ruby', '-e', '5.times{sleep 1;puts _1}'], 'w+' do |io|
  5.times { puts io.gets }
  # expect: prints one line per sec for 5 times
  # actual: wait for 5 sec and prints 5 lines at once
  # It means, `line.match?(/binding\.irb/)` in test_debug_cmd will timeout.
end
```

tompng:
Thanks for the diff, test passed!

st0012:
Thanks for the fix and explanation 🙏
Yeah I think ideally we should:
- Either more realisticly simulate users’ activity in tests, which in this case means doing cursor position report (does this make sense though?)
- Or we should switch to popen

But we can do that in a follow up PR and let’s fix the build first 👍

st0012 (approved):

st0012 (approved):

PR #583, opened 2023-05-18
Refactor RubyLex's input/io methods

st0012:
1. Make `RubyLex#set_input` simply assign the input block. This matches the behavior of `RubyLex#set_prompt`.
2. Merge `RubyLex#set_input`'s IO configuration logic with `#set_auto_indent` into `#configure_io`.

I think `RubyLex#configure_io` could use further refactor. But I want to revisit it after #576 is merged.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -537,7 +537,7 @@ module IRB
         @context.io.prompt
       end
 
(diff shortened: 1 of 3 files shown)


tompng (approved):
Looks Good :+1:

PR #584, opened 2023-05-19
Display mod key as `Option` on Darwin platforms

adam12:
Check RUBY_PLATFORM for `darwin` and modify the mod key from `Alt` to `Option`.


st0012 (approved):
Thank you!

PR #585, opened 2023-05-19
Fix dynamic_prompt test not executed, remove unnecessary set_input

tompng:
The assersion `assert_dynamic_prompt` was not working.
`MockIO_DynamicPrompt#dynamic_prompt` is triggered when configure_io is called. it was changed on refactor of `set_input`.

I also removed `set_input { @io.gets }`.
The block passed to set_input is not called and `@io` is always nil because it is an instance variable of `TestIRB::TestRubyLex`.


st0012 (approved):
Great catch!
I wonder if we can move assertions out of the `dynamic_prompt` block so they won't be skipped unnoticed again.

PR #586, opened 2023-05-20
Add assertion for dynamic_prompt's assertion execution

st0012:
Because the assertions for `dynamic_prompt` lives inside a block given to `RubyLex`, they could be skipped unnoticed if the setup is not correct.

This commit adds a simple assertion to check if the block was actually executed.


tompng (approved):
:+1:

PR #589, opened 2023-05-22
Allow `show_source` for private methods

### LINKED ISSUE #577: ShowSource of private method
## Description

When reading source code using show_source, I want to also look at private method definition.

```ruby
class A
  def pub_foobar
    123
  end
  private def pri_foobar
    456
  end
end

a = A.new
binding.irb
```

```
irb(main):001:0> show_source A#pri_foobar
Error: Couldn't locate a definition for A#pri_foobar
=> nil
irb(main):002:0> show_source a.pri_foobar
Error: Couldn't locate a definition for a.pri_foobar
=> nil
irb(main):003:0> show_source A#pub_foobar

From: a.rb:2

  def pub_foobar
    123
  end

=> nil
irb(main):004:0> show_source a.pub_foobar

From: a.rb:2

  def pub_foobar
    123
  end

=> nil
irb(main):005:0> 

```

makenowjust:
Fix #577

### DIFF
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -496,7 +496,7 @@ module TestIRB
           private def show_source_test_method
             unless true
             end
-          end unless defined?(show_source_test_method)
+          end unless private_method_defined?(:show_source_test_method)
         end
       EOS
 
@@ -516,7 +516,7 @@ module TestIRB
           private def show_source_test_method
             unless true
             end
-          end unless defined?(show_source_test_method)
+          end unless private_method_defined?(:show_source_test_method)
         end
 
         Instance = PrivateTest.new unless defined?(Instance)
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -490,6 +490,7 @@ module TestIRB
     end
 
     def test_show_source_private_instance
+      pend if RUBY_ENGINE == 'truffleruby'
       eval(code = <<-EOS, binding, __FILE__, __LINE__ + 1)
         class PrivateInstanceTest
           private def show_source_test_method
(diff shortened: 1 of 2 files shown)


test/irb/test_cmd.rb
```diff
@@ -488,6 +488,47 @@ def show_source_test_method
...
     end
+
+    def test_show_source_private_instance
+      pend if RUBY_ENGINE == 'truffleruby'
+      eval(code = <<-EOS, binding, __FILE__, __LINE__ + 1)
+        class PrivateInstanceTest
+          private def show_source_test_method
+            unless true
+            end
+          end unless defined?(show_source_test_method)
```
tompng, line 499: I think this `defined?` is always nil. How about `private_method_defined?`
```ruby
class A
  def foo;end
  def self.bar;end
  p defined?(foo) #=> nil
  p defined?(bar) #=> 'method'
end
```

makenowjust: Ah, yes. However, I'm not sure why here's `defined?` is necessary. I added "defined" as in the test case above (`test_show_source_end_finder`). Do you know that?

tompng: It is first added in this commit https://github.com/ruby/ruby/commit/ca1aa7afeae43b02d00cf2f2a82acde293e48875 and commited to ruby/irb later.
I think it is meant to avoid method-redefinition warning when running the same test case several times.

In ruby/ruby, there are CI that runs test multiple times like `trunk-repeat20-asserts` in `ci.rvm.jp`.
(I didn't know that. ko1 san told me.)

makenowjust: Thank you. I fixed it.

tompng (approved):
Looks good, thank you :+1:

Issue #579, opened 2023-05-07
When require in .irbrc fails with LoadError, error message will be suppressed


tompng:
## Description

If you write `require 'something'` in irbrc and it fails loading, IRB does not show error message.
I will be hard to know the reason why irbrc is not loaded.

## Reproduce

With this `.irbrc` file
```ruby
puts 'irbrc_loading'
require 'matrix'
puts 'foobarbaz'
foo.bar.baz
```

Normally, error in irbrc will be displayed.
```
[tomoya:irbrctest]% irb
irbrc_loading
foobarbaz
load error: /Users/tomoya/Desktop/irbrctest/.irbrc
NameError: undefined local variable or method `foo' for main:Object
	/Users/tomoya/Desktop/irbrctest/.irbrc:4:in `<top (required)>'
	/Users/tomoya/Documents/github/ruby/irb/lib/irb/init.rb:360:in `load'
	/Users/tomoya/Documents/github/ruby/irb/lib/irb/init.rb:360:in `run_config'
	/Users/tomoya/Documents/github/ruby/irb/lib/irb/init.rb:14:in `setup'
irb(main):001:0> 
```

But LoadError is suppressed (with empty Gemfile and bundle exec, `require 'matrix'` fails)
```
[tomoya:irbrctest]% cat Gemfile
# empty
[tomoya:irbrctest]% bundle exec irb
irbrc_loading
irb(main):001:0> 
```

## Result of irb_info

```
irb(main):001:0> irb_info
=> 
Ruby version: 3.2.0
IRB version: irb 1.6.4 (2023-04-07)
InputMethod: RelineInputMethod with Reline 0.3.3 and /Users/tomoya/.inputrc
.irbrc path: /Users/tomoya/Desktop/irbrctest/.irbrc
RUBY_PLATFORM: x86_64-darwin20
LANG env: ja_JP.UTF-8
East Asian Ambiguous Width: 1
```

## Terminal Emulator

macOS Terminal.app

## Setting Files

Are you using `~/.irbrc` and `~/.inputrc`?
Using `.irbrc`. contents are written above.

st0012:
I think it's [this line](https://github.com/ruby/irb/blob/master/lib/irb/init.rb#L361) that suppresses it. And the rescue of `LoadError` was added from the [very beginning](https://github.com/ruby/irb/commit/f47808999d24865fba1929dea1a7011ff5a517d6) of IRB, so I'm not sure why that's added.
[...]

PR #593, opened 2023-06-01
Fix LoadError rescue when require fails

AI-Mozi:
Issue: #579
Should I add something else (test etc.)?

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -358,7 +358,7 @@ module IRB # :nodoc:
     if @CONF[:RC]
       begin
         load rc_file
-      rescue LoadError, Errno::ENOENT
+      rescue Errno::ENOENT
       rescue # StandardError, ScriptError
         print "load error: #{rc_file}\n"
         print $!.class, ": ", $!, "\n"


st0012 (commented):
Yeah having a test case would be great. Also we should check if the rc file exists before trying to load it.

st0012:
A few points that should help:
- The test should be inside `test_init.rb`
- The gist of the test would be:
    1. Write a rc file that could cause `LoadError`
    2. Make sure `IRB.rc_file` points to that file's path
    3. Call `IRB.run_config`
    4. Make the exception is raised, likely with `assert_raise`

lib/irb/init.rb
```diff
@@ -358,7 +358,7 @@ def IRB.run_config
     if @CONF[:RC]
       begin
         load rc_file
-      rescue LoadError, Errno::ENOENT
+      rescue Errno::ENOENT
       rescue # StandardError, ScriptError
         print "load error: #{rc_file}\n"
```
hasumikin, line 363: I imagine that the commented-out `# StandardError, ScriptError` would originally be meant to rescue a LoadError in addition to other "standard" exceptions.
(Note that LoadError is not a descendant of StandardError but a descendant of ScriptError)
I think we should simply make the comment work. Otherwise, we can not rescue a LoadError.

Besides, as the code will be going to rescue both StandardError and ScriptError, the message print "**load** error:" is not appropriate here. We can also simply remove the word "load".

```suggestion
      rescue StandardError, ScriptError
        print "error: #{rc_file}\n"
```

st0012:
Closing in favor of #817

PR #598, opened 2023-06-04
Refactor ExtendCommand::Nop

st0012:
1. The `conf` name is very confusing because it doesn't mean `IRB.conf` but the current `irb_context` object. So I think it should be renamed.
2. We don't need to expose this method to all command classes, especially
when it's just an alias of `irb_context.irb`.

### DIFF
--- a/lib/irb/cmd/nop.rb
+++ b/lib/irb/cmd/nop.rb
@@ -30,15 +30,15 @@ module IRB
         end
       end
 
-      def self.execute(conf, *opts, **kwargs, &block)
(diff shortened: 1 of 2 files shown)


tompng (approved):
Looks good :+1:
For removing method `irb`, I slightly searched on github and it doesn't seem to be used in custom commands.

st0012:
I personally don't worry about breaking custom commands at the API level as they're never official. I only think we need to be careful on changing constant names.

PR #601, opened 2023-06-10
Use `Kernel.warn` to print command alias warning

st0012:
This aligns with other warnings in IRB and properly channels the message to stderr.

### DIFF
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -289,7 +289,7 @@ module IRB # :nodoc:
           alias_method to, from
         }
       else
-        Kernel.print "irb: warn: can't alias #{to} from #{from}.\n"
+        Kernel.warn "irb: warn: can't alias #{to} from #{from}.\n"
       end
     end
 


lib/irb/extend-command.rb
```diff
@@ -289,7 +289,7 @@ def install_alias_method(to, from, override = NO_OVERRIDE)
           alias_method to, from
         }
       else
-        Kernel.print "irb: warn: can't alias #{to} from #{from}.\n"
+        warn "irb: warn: can't alias #{to} from #{from}.\n"
```
tompng, line 292: Since the previous code is not just `print` but `Kernel.print`, I think it should it be `Kernel.warn` to avoid this kind of behavior
```
irb(main):001:0> obj = Object.new
=> #<Object:0x0000000103f67710>
irb(main):002:0> def obj.warn(a) = puts(a.reverse)
=> :warn
irb(main):003:0> def obj.step = nil
=> :step
irb(main):004:0> cws obj

.pets_bri morf pets saila t'nac :nraw :bri
=> #<Object:0x0000000103f67710>
```

st0012: Good catch 👍 Updated

tompng (approved):

PR #602, opened 2023-06-10
Deprecate measure block usage

st0012:
1. Add custom command definition for `irb_measure`
2. Deprecate `measure` command's block usage

This will officially remove the need to pass block to command's `execute`
method, which will make future command refactoring a lot easier.

### DIFF
--- a/lib/irb/extend-command.rb
+++ b/lib/irb/extend-command.rb
@@ -259,12 +259,34 @@ module IRB # :nodoc:
         cmd_class = cmd_class.name
       end
 
-      line = __LINE__; eval %[
-        def #{cmd_name}(*opts, **kwargs, &b)
-          Kernel.require_relative "#{load_file}"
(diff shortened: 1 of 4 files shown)


lib/irb/extend-command.rb
```diff
@@ -259,12 +259,34 @@ def self.def_extend_command(cmd_name, cmd_class, load_file, *aliases)
...
+
+              IRB.conf[:MEASURE] = true
+              added = IRB.set_measure_callback(&block)
+              puts "\#{added[0]} is added." if added
+            else
+              Kernel.require_relative "cmd/measure"
+              ::IRB::ExtendCommand::Measure.execute(irb_context, type, arg)
+            end
+          end
+        ], nil, __FILE__, line
```
st0012, line 280: I know this is not pretty,  but this is the only way I can think of to make command refactoring easy while not introducing a n immediate breaking change.

st0012:
I think #624 is a better way of handling `measure`, so closing this now.

PR #610, opened 2023-06-21
Omit nesting_level, use indent_level to build prompt string

tompng:
## Description

The document comment says that the part `%i` of the prompt is indent_level.
https://github.com/ruby/irb/blob/1159c13d90c5d2ee708545b6bbfcc18ab826930b/lib/irb.rb#L182-L192

This pull request removes calculation of nesting_level introduced in #500 and #515 and use indent_level for building prompt string.

## Concerns
In `test_ruby_lex.rb`, the word `nesting_level` is used instead of `indent_level`.

But if we want `nesting_level` in prompt string again, we can easily re-implement it like this
```ruby
NESTING_LEVEL_COUNTING_TOKENS = %i[on_kw on_tlambda on_tlambeg on_lparen on_lbracket on_lbrace]
def calc_nesting_level(opens)
  opens.count do |t|
    NESTING_LEVEL_COUNTING_TOKENS.include?(t.event)
  end
end
```

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -184,8 +184,8 @@ class RubyLex
 
   def prompt(opens, continue, line_num_offset)
     ltype = ltype_from_open_tokens(opens)
-    _indent_level, nesting_level = calc_nesting_depth(opens)
-    @prompt&.call(ltype, nesting_level, opens.any? || continue, @line_no + line_num_offset)
+    indent_level = calc_indent_level(opens)
+    @prompt&.call(ltype, indent_level, opens.any? || continue, @line_no + line_num_offset)
   end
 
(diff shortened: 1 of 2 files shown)


st0012:
Looking at the [commit](https://github.com/ruby/irb/commit/f1a775af47c3e9d39b9c917863da391d97f291b7) that introduced `Row`, the assertion used `RubyLex`'s `@indent` ivar but calling it `assert_nesting_level`.

Given that the same test also already had `assert_indenting` to check the output with indentation added, my guess is that it's just `indent` and `nesting` were somehow used interchangeably. And `nesting_level` was picked to avoid having both `assert_indent` and `assert_indenting`, which would be confusing.

st0012 (approved):

PR #611, opened 2023-06-21
Fix process_continue and check_code_block

tompng:
## Rename and refactor process_continue and check_code_block
```ruby
def process_continue(tokens)
  # doing something like maybe_should_continue?
  # check if last token is BEG (half part of `should_continue?`)
end

def check_code_block(code)
  # doing something like code_syntax_is_not_complete_or_maybe_should_continue?
  # check code is syntax_valid recoverable_error unrecoverable_error
  # check if the last token is DOT and checks other duplicated condition with syntax check and open_tokens check (other half part of `should_continue?`)
end
```
↓
```ruby
def should_continue?(tokens) = true or false
def check_code_syntax(code) = :valid or :recoverable_error or :unrecoverable_error or :other_error

def code_terminated?(code, tokens, opens)
  # calculate by should_continue?() and check_code_syntax()
end
```

This will fix prompt bug introduced in #500 or #515. (because process_continue was incomplete, half of the necessary implementation was in check_code_block)
```
# actual
irb(main):001:1> if 1
irb(main):002:0> end.

# expected
irb(main):001:1> if 1
irb(main):002:0* end.
```

## Other fix
Previous implementation of `process_continue` only uses last two tokens.
It ignores only the last token newline tokens
I change it to ignore consecutive trailing comments, spaces, newlines.
It will fix this old bug.
```
actual
irb(main):001:0* 1; # comment
irb(main):002:0* 
irb(main):003:0* 

expected
irb(main):001:0> 1; # comment
=> 1
irb(main):002:0> 
```

### DIFF
--- a/lib/irb/cmd/show_source.rb
+++ b/lib/irb/cmd/show_source.rb
@@ -58,9 +58,9 @@ module IRB
           tokens.chunk { |tok| tok.pos[0] }.each do |lnum, chunk|
             code = lines[0..lnum].join
             prev_tokens.concat chunk
-            continue = lex.process_continue(prev_tokens)
-            code_block_open = lex.check_code_block(code, prev_tokens)
-            if !continue && !code_block_open
+            continue = lex.should_continue?(prev_tokens)
+            syntax = lex.check_code_syntax(code)
+            if !continue && syntax == :valid
               return first_line + lnum
             end
           end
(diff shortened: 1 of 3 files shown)


st0012 (approved):
LGTM 👍 Just need to fix some conflicts

PR #613, opened 2023-06-25
Stop treating history-saving logic as extension

st0012:
Since `IRB.conf[:SAVE_HISTORY]` is [assigned with 1000 by default](https://github.com/ruby/irb/blob/master/lib/irb/init.rb#L54), history-saving is a feature enabled by default.  So it should not be treated as an extension, which adds unnecessary complexity to the code.

(I plan to rename `ext/history` to `ext/eval_history` later as it's a different feature)


tompng (approved):

eregon:
I have this in my ~/.irbrc:
```ruby
require 'irb/completion'
require 'irb/ext/save-history'
IRB.conf[:SAVE_HISTORY] = 1000
```
On Ruby 3.3.0, now `irb` warns like this:
```
irb          
Error loading RC file '/home/eregon/.irbrc':
<internal:/home/eregon/.rubies/ruby-3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require': cannot load such file -- irb/ext/save-history (LoadError)
Did you mean?  irb/ext/eval_history
	from <internal:/home/eregon/.rubies/ruby-3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
	from /home/eregon/.irbrc:2:in `<top (required)>'
	from /home/eregon/.rubies/ruby-3.3.0/lib/ruby/3.3.0/irb/init.rb:399:in `load'
	from /home/eregon/.rubies/ruby-3.3.0/lib/ruby/3.3.0/irb/init.rb:399:in `run_config'
	from /home/eregon/.rubies/ruby-3.3.0/lib/ruby/3.3.0/irb/init.rb:49:in `setup'
	from /home/eregon/.rubies/ruby-3.3.0/lib/ruby/3.3.0/irb.rb:896:in `start'
	from /home/eregon/.rubies/ruby-3.3.0/lib/ruby/gems/3.3.0/gems/irb-1.11.0/exe/irb:9:in `<top (required)>'
	from /home/eregon/.rubies/ruby-3.3.0/bin/irb:25:in `load'
	from /home/eregon/.rubies/ruby-3.3.0/bin/irb:25:in `<main>'
irb(main):001> 
```
There is no issue on Ruby < 3.3, e.g. it works fine on Ruby 3.2.2 (with the irb shipped with that Ruby release).

What is a good solution to ensure history is enabled both on 3.3.0 (where it's default) and on older versions?

st0012:
I think this is similar to #804 

Since as early as IRB 0.9.6, `require 'irb/ext/save-history'` has become obsolete as specifying `IRB.conf[:SAVE_HISTORY] = 1000` would already load it. So you can simply remove that require and it should still work for both Ruby 3.2 and 3.3 (even 3.1 or 3.0 I guess).

> Also is require 'irb/completion' still non-default or I can also remove that?

Yeah you can remove it as it's required by default since Ruby 2.7 (IRB 1.2.6)

PR #615, opened 2023-06-27
Move input line mutation out of `Context#evaluate`

st0012:
This makes sure `Context#evaluate` really just evaluates the input. It will also make #575's implementation cleaner.


tompng (approved):
:+1:

PR #616, opened 2023-06-27
Use Ripper.sexp instead of Regexp for completion

tompng:
## Description

Replaces completion method from Regex based to Ripper.sexp base

## Pros

Can complete these codes that Regexp based implementation provides wrong completion.
```ruby
# Code  | Ripper | Regexp
{}.█   #| Hash   | Hash or Proc
->{}.█ #| Proc   | Hash or Proc
%w{}.█ #| Array  | Hash or Proc
%{}.█  #| String | Hash or Proc
[].█   #| Array  | Array
a[].█  #| ------ | Array
/a/i.█ #| Regexp | ------
%/a/.█ #| String | Regexp
```

## Cons
Cannot complete if there are syntax error before cursor. Regexp based can complete these.
```ruby
@;1.█

1=1;1.█

if a*(b+c)) < d
  1.█
```

## Limitation

Cannot complete these code now. I can support this later (about +40 lines).
```ruby
# Inside ternary operator
a = condition ? true.a█ : false

# Inside block parameter
array.each_with_index do |value, index, other = 1.a█
```

## Why not RubyVM::AbstractSyntaxTree or YARP

Using RubyVM::AbstractSyntaxTree (and I beleave YARP too) is easier than Ripper.sexp but:

- RubyVM::AbstractSynatxTree: TruffleRuby does not have it
- YARP: Not a gem now, ruby >= 3.1 is specified in Gemfile
- Combination of both: Hard to maintain

## Future plan

Use RBS to provide powerful completion

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -24,11 +24,25 @@ module IRB
         end
 
         def eval_global_variables
-          ::Kernel.instance_method(:global_variables).bind(eval("self")).call
+          ::Kernel.global_variables
         end
 
-        def eval_class_constants
-          ::Module.instance_method(:constants).bind(eval("self.class")).call
+        def eval_constants
+          [Object, *eval('::Module.nesting')].flat_map(&:constants).uniq.sort rescue []
+        end
+
+        def eval_class_variables
+          mod = eval('::Module.nesting').first
+          mod&.class_variables || [] rescue []
+        end
+
+        def eval_instance_variable_get(name)
+          ::Kernel.instance_method(:instance_variable_get).bind_call(eval('self'), name) rescue nil
+        end
+
+        def eval_class_variable_get(name)
+          mod = eval('::Module.nesting').first
+          mod&.class_variable_get(name) rescue nil
         end
       end
     }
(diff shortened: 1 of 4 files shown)


st0012:
I think using Ripper already caused some compatibility issues with implementations like TruffleRuby and JRuby? It's mostly why we still have tests pending for TruffleRuby (I paired with @nirvdrum on those).

Given that both TruffleRuby and JRuby are actively adopting YARP, which should support error recovery and has better API, how about:

1. We make the current regexp completion a `RegexpCompleter`
2. We build another completer called `YARPCompleter` and make it the default
3. If the project uses Ruby 3.1-, we fallback to `RegexpCompleter`

tompng:
Thank you for the feedback.

I think ripper compatibility issue is small. I found that pended test in test_ruby_lex and test_cmd passes.
#619

The reason of pend test failure was actually an incompatibility of `SyntaxError#message`, not Ripper.
https://github.com/ruby/irb/blob/62691384f8db5253c1454f37b2c688ab8d2c28b7/lib/irb/ruby-lex.rb#L304-L318
(Before #500, test was failing but If I add some regexp pattern here, test passed)

RegexpCompleter and YARPCompleter idea looks good. I'll try it. I'm looking forward for YARP gem release.

st0012:
@tompng Since Prism (formerly YARP) has been released as a gem and I think the pros generally out-weights the cons, I'm fine with replacing `RegexpCompletor` completely with the completor we build with it (which also means we need to drop 2.7 support though).
But let's finish #707 first and implement the new completor with the specification we make there.

tompng:
Closing this because of #708

PR #617, opened 2023-06-27
Remove unused keyword `exception` from Context#evaluate

tompng:
I think `exception` is always nil. Non-nil value is only passed from test code.

## Background

I think the original code want to do something like this.
```
irb> raise 'a'
in `<main>': a (RuntimeError)
irb> 42
=> 42
irb> [_]
=> [42]
irb> $!
=> #<RuntimeError: a> ← want this
```

But it was breaking the value `_` (https://bugs.ruby-lang.org/issues/14749)
```
irb> raise 'a'
in `<main>': a (RuntimeError)
irb> 42
=> 42
irb> [_]
=> [#<RuntimeError: a>] ← wrong
irb> $!
=> #<RuntimeError: a>
```

And the feature (I think it's never implemented correctly) is dropped
```
irb> raise 'a'
in `<main>': a (RuntimeError)
irb> 42
=> 42
irb> [_]
=> [42]
irb> $!
=> nil ← it's nil now
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -506,8 +506,6 @@ module IRB
 
     # Evaluates input for this session.
     def eval_input
-      exc = nil
-
       @scanner.set_prompt do
         |ltype, indent, continue, line_no|
         if ltype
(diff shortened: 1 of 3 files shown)


st0012 (approved):
Yeah it seems to be the case. Good catch 👍

PR #623, opened 2023-07-03
Refactor eval history

st0012:
This feature was not enabled by default, not tested, and could be easily confused with IRB's history saving feature. So in this PR I did these improvements:

- Rename the file and constant so it's clearer what it is for.
- Add basic test cases for it.
    - I found that it only displays `limit - 1` results, which is probably a bug. But since it's not widely used and we don't have any reports about it, I don't prioritise fixing it atm.


tompng (approved):
Looks good!

> could be easily confused with IRB's history saving feature

renaming :+1:

Issue #588, opened 2023-05-21
Support help method extension API


st0012:
In #513, I proposed extension API for commands, but it doesn't cover all the extension needs. For example, Rails defines [`controller` and `helpers` methods](https://github.com/rails/rails/blob/ef04fbb3b256beececfa44c47c4ec93ac6945e59/railties/lib/rails/console/helpers.rb#L4) for its Rails console, which are meant to be used as methods and be chained with other method calls. So they should not be treated as commands.

However, helper methods do share some common needs with commands, like:

- Feature discovery, e.g. can be listed in the result of `show_cmds` (or `help` in the future)
- Proper registration API and documentation (instead of [manually including a module](https://github.com/rails/rails/blob/main/railties/lib/rails/commands/console/console_command.rb#L72)) 

So I think we should provide (perhaps similar) extension for helper methods to fulfill such needs.

PR #624, opened 2023-07-03
Support helper method registration

st0012:
## Summary

This Pull Request introduces a helper extension API as discussed in issue #588. It aims to provide an API for helper methods in IRB, improving the method-extending approach currently in use. The new approach would make helper methods more discoverable and user-friendly.

## Background

There are two categories of operations in IRB: **Helper methods** and **Commands**.

**Helper methods**
- They are designed to be used as Ruby methods.
- Users primarily use them for their return values, such as calling `app` for a Rails application instance.
- At present, this is achieved by directly extending `ExtendCommandBundle`, like [`controller`](https://github.com/rails/rails/blob/ef04fbb3b256beececfa44c47c4ec93ac6945e59/railties/lib/rails/console/helpers.rb#L15) or [`app`](https://github.com/rails/rails/blob/ef04fbb3b256beececfa44c47c4ec93ac6945e59/railties/lib/rails/console/app.rb#L10-L13) in Rails console.

**Commands**
- They are intended to be used with their own syntax, for example, `show_source Foo#bar`.
- Users utilize them to complete certain tasks (like `edit`) or display information (like `help`). The return value is generally not important.
- They will also have an official registration API, as discussed in #513.

## Motivation

The current method-extending approach has several limitations:
- `ExtendCommandBundle` is an internal component that is currently exposed, which is not ideal.
- IRB cannot effectively detect and aid users in discovering the extended helpers.
- Due to the above reasons, we don't officially support or mention a method to extend IRB with helpers, which has likely limited IRB's potential for personal or library customization.

## Implementation

This PR introduces a new `IRB::HelperMethod::Base` class that provides `description` methods to helper classes. After requiring `irb/helper_method`, users can use `IRB::HelperMethod.register(:my_helper, MyHelper)` to add new helpers. Once registered, a `my_helper` method will be added to IRB sessions.

Here's an example:

```rb
# test.rb

# this doesn't require the entire IRB, just the helper registration API
require "irb/helper_method"

class MyHelperMethod < IRB::HelperMethod::Base
  description "Describe how to use this helper"

  # all kinds of arguments are supported
  def execute(*args, **opts, &block)
    "Hello from MyHelperMethod#execute"
  end
end

IRB::HelperMethod.register(:my_helper, MyHelperMethod)

binding.irb

# irb(main):001:0> puts my_helper
# Hello from MyHelperMethod#execute
# => nil
```

## Changes

- A new `IRB::HelperMethod::Base` class has been added.
- Users can now register new helpers using `IRB::HelperMethod.register(:my_helper, MyHelperMethod)`.
- `conf` is now declared as helpers.
- `help` now also displays registered helpers.

    <details>

    ```
    IRB
      exit           Exit the current irb session.
      irb_load       Load a Ruby file.
      irb_require    Require a Ruby file.
      source         Loads a given file in the current session.
      irb_info       Show information about IRB.
      show_cmds      List all available commands and their description.
      history        Shows the input history. `-g [query]` or `-G [query]` allows you to filter the output.
    
    Workspace
      cwws           Show the current workspace.
      chws           Change the current workspace to an object.
      workspaces     Show workspaces.
      pushws         Push an object to the workspace stack.
      popws          Pop a workspace from the workspace stack.
    
    Multi-irb (DEPRECATED)
    
    Debugging
      debug          Start the debugger of debug.gem.
      break          Start the debugger of debug.gem and run its `break` command.
      catch          Start the debugger of debug.gem and run its `catch` command.
      next           Start the debugger of debug.gem and run its `next` command.
      delete         Start the debugger of debug.gem and run its `delete` command.
      step           Start the debugger of debug.gem and run its `step` command.
      continue       Start the debugger of debug.gem and run its `continue` command.
      finish         Start the debugger of debug.gem and run its `finish` command.
      backtrace      Start the debugger of debug.gem and run its `backtrace` command.
      info           Start the debugger of debug.gem and run its `info` command.
    
    Misc
      edit           Open a file with the editor command defined with `ENV["VISUAL"]` or `ENV["EDITOR"]`.
      measure        `measure` enables the mode to measure processing time. `measure :off` disables it.
    
    Context
      help           [DEPRECATED] Enter the mode to look up RI documents.
      show_doc       Enter the mode to look up RI documents.
    
    Helper methods
      conf           Returns the current context.
    
    Aliases
      $              Alias for `show_source`
      @              Alias for `whereami`
    ```
    </details>

### DIFF
--- a/lib/irb/default_commands.rb
+++ b/lib/irb/default_commands.rb
@@ -98,10 +98,7 @@ module IRB
     end
 
     _register_with_aliases(:irb_context, Command::Context,
-      [
-        [:context, NO_OVERRIDE],
-        [:conf, NO_OVERRIDE],
-      ],
+      [:context, NO_OVERRIDE]
     )
 
     _register_with_aliases(:irb_exit, Command::Exit,
--- a/lib/irb/command/help.rb
+++ b/lib/irb/command/help.rb
@@ -24,7 +24,9 @@ module IRB
 
       def help_message
         commands_info = IRB::Command.all_commands_info
+        helper_methods_info = IRB::HelperMethod.all_helper_methods_info
         commands_grouped_by_categories = commands_info.group_by { |cmd| cmd[:category] }
+        commands_grouped_by_categories["Helper methods"] = helper_methods_info
 
         user_aliases = irb_context.instance_variable_get(:@user_aliases)
 
--- a/lib/irb/helper_method.rb
+++ b/lib/irb/helper_method.rb
@@ -1,0 +1,29 @@
+require_relative "helper_method/base"
+
+module IRB
+  module HelperMethod
+    @helper_methods = {}
+
+    class << self
+      attr_reader :helper_methods
+
+      def register(name, helper_class)
+        @helper_methods[name] = helper_class
+
+        if defined?(HelpersContainer)
+          HelpersContainer.install_helper_methods
+        end
+      end
+
+      def all_helper_methods_info
+        @helper_methods.map do |name, helper_class|
+          { display_name: name, description: helper_class.description }
+        end
+      end
+    end
+
+    # Default helper_methods
+    require_relative "helper_method/conf"
+    register(:conf, HelperMethod::Conf)
+  end
+end
--- a/lib/irb/helper_method/base.rb
+++ b/lib/irb/helper_method/base.rb
@@ -1,0 +1,12 @@
+module IRB
+  module HelperMethod
+    class Base
+      class << self
+        def description(description = nil)
+          @description = description if description
+          @description
+        end
+      end
+    end
+  end
+end
--- a/lib/irb/helper_method/conf.rb
+++ b/lib/irb/helper_method/conf.rb
@@ -1,0 +1,11 @@
+module IRB
+  module HelperMethod
+    class Conf < Base
+      description "Returns the current context."
+
+      def execute
+        IRB.CurrentContext
+      end
+    end
+  end
+end
--- a/lib/irb/workspace.rb
+++ b/lib/irb/workspace.rb
@@ -6,6 +6,8 @@
 
 require "delegate"
 
+require_relative "helper_method"
+
 IRB::TOPLEVEL_BINDING = binding
 module IRB # :nodoc:
   class WorkSpace
@@ -109,9 +111,9 @@ EOF
     attr_reader :main
 
     def load_helper_methods_to_main
-      if !(class<<main;ancestors;end).include?(ExtendCommandBundle)
-        main.extend ExtendCommandBundle
-      end
+      ancestors = class<<main;ancestors;end
+      main.extend ExtendCommandBundle if !ancestors.include?(ExtendCommandBundle)
+      main.extend HelpersContainer if !ancestors.include?(HelpersContainer)
     end
 
     # Evaluate the given +statements+ within the  context of this workspace.
@@ -171,5 +173,17 @@ EOF
 
       "\nFrom: #{file} @ line #{pos + 1} :\n\n#{body}#{Color.clear}\n"
     end
+  end
+
+  module HelpersContainer
+    def self.install_helper_methods
+      HelperMethod.helper_methods.each do |name, helper_method_class|
+        define_method name do |*args, **opts, &block|
+          helper_method_class.new.execute(*args, **opts, &block)
+        end unless method_defined?(name)
+      end
+    end
+
+    install_helper_methods
   end
 end
--- a/test/irb/command/test_help.rb
+++ b/test/irb/command/test_help.rb
@@ -62,5 +62,14 @@ module TestIRB
       assert_match(/\$\s+Alias for `show_source`/, out)
       assert_match(/@\s+Alias for `whereami`/, out)
     end
+
+    def test_help_lists_helper_methods
+      out = run_ruby_file do
+        type "help"
+        type "exit"
+      end
+
+      assert_match(/Helper methods\s+conf\s+Returns the current context/, out)
+    end
   end
 end
--- a/test/irb/test_helper_method.rb
+++ b/test/irb/test_helper_method.rb
@@ -1,0 +1,109 @@
+# frozen_string_literal: true
+require "irb"
+
+require_relative "helper"
+
+module TestIRB
+  class HelperMethodTestCase < TestCase
+    def setup
+      $VERBOSE = nil
+      @verbosity = $VERBOSE
+      save_encodings
+      IRB.instance_variable_get(:@CONF).clear
+    end
+
+    def teardown
+      $VERBOSE = @verbosity
+      restore_encodings
+    end
+
+    def execute_lines(*lines, conf: {}, main: self, irb_path: nil)
+      IRB.init_config(nil)
+      IRB.conf[:VERBOSE] = false
+      IRB.conf[:PROMPT_MODE] = :SIMPLE
+      IRB.conf.merge!(conf)
+      input = TestInputMethod.new(lines)
+      irb = IRB::Irb.new(IRB::WorkSpace.new(main), input)
+      irb.context.return_format = "=> %s\n"
+      irb.context.irb_path = irb_path if irb_path
+      IRB.conf[:MAIN_CONTEXT] = irb.context
+      IRB.conf[:USE_PAGER] = false
+      capture_output do
+        irb.eval_input
+      end
+    end
+  end
+
+  module TestHelperMethod
+    class ConfTest < HelperMethodTestCase
+      def test_conf_returns_the_context_object
+        out, err = execute_lines("conf.ap_name")
+
+        assert_empty err
+        assert_include out, "=> \"irb\""
+      end
+    end
+  end
+
+  class HelperMethodIntegrationTest < IntegrationTestCase
+    def test_arguments_propogation
+      write_ruby <<~RUBY
+        require "irb/helper_method"
+
+        class MyHelper < IRB::HelperMethod::Base
+          description "This is a test helper"
+
+          def execute(
+            required_arg, optional_arg = nil, *splat_arg, required_keyword_arg:,
+            optional_keyword_arg: nil, **double_splat_arg, &block_arg
+          )
+            puts [required_arg, optional_arg, splat_arg, required_keyword_arg, optional_keyword_arg, double_splat_arg, block_arg.call].to_s
+          end
+        end
+
+        IRB::HelperMethod.register(:my_helper, MyHelper)
+
+        binding.irb
+      RUBY
+
+      output = run_ruby_file do
+        type <<~INPUT
+          my_helper(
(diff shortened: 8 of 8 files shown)


lib/irb/helper.rb
```diff
@@ -0,0 +1,25 @@
...
+      def all_helpers_info
+        @helpers.map do |name, helper_class|
+          { display_name: name, description: helper_class.description, category: helper_class.category }
+        end
+      end
+    end
+
+    # Default helpers
+    require_relative "helper/measure"
+    register(:measure, Helper::Measure)
```
st0012, line 23: I think we should add default helpers with `irb/helper` so users/extensions will be able to override them.

lib/irb/cmd/show_cmds.rb
```diff
@@ -28,6 +28,29 @@ def execute(*args)
           output.puts
         end
 
+        helpers_info = IRB::Helper.all_helpers_info
+        output.puts <<~MSG
+        .----------------.
+        | Helper Methods |
+        '----------------'
+
+        (Helpers are Ruby methods that are added to the current IRB session)
```
st0012, line 37: Not sure if we need to explain what helpers are 🤔

lib/irb/helper/measure.rb
```diff
@@ -1,17 +1,9 @@
-require_relative "nop"
-
 module IRB
-  # :stopdoc:
-
-  module ExtendCommand
-    class Measure < Nop
+  module Helper
+    class Measure < Base
```
st0012, line 3: Migration is so simple!

st0012:
@vinistock provided some feedback on this:

- To end users, the distinction between helpers and commands are not very useful. So a separate entry in `show_cmds` may be confusing instead.
- The name `Helper` doesn't explain the concept very well.
    - Maybe `HelperMethod` is better?

lib/irb/cmd/show_cmds.rb
```diff
@@ -28,6 +28,29 @@ def execute(*args)
...
+
+        MSG
+
+        helpers_grouped_by_categories = helpers_info.group_by { |cmd| cmd[:category] }
+        longest_helper_name_length = helpers_info.map { |c| c[:display_name] }.max { |a, b| a.length <=> b.length }.length
```
andyw8, line 42:
```suggestion
        longest_helper_name_length = helpers_info.map { |c| c[:display_name] }.max_by(&:length)
```

st0012: Updated, thank you 👍

tompng:
I like it :+1:

Maybe there  are some helper that should be defined only in main object. For example:

```ruby
irb(main):001> controller
=> <#ApplicationController:0x123>
irb(main):002> cws UsersController.new
irb(UsersController):003> controller
=> # If it returns ApplicationController instead of UsersController, it is misleading.

irb(main):001> cws Article.first
irb(Article):002> update author: controller.current_user
=> true
# User might think that it is possible to write the code above in model method, but it's not.
```

If all of the commands/helpers are `commands` or `helpers only for main`, we could completely avoid helper method polluting workspace object, especially in debug that current workspace object will change frequently by `step`.

(This is just an idea for future, and I don't know if these kind of main-only helpers are really wanted.)

st0012:
@tompng That's a great point! But maybe we can improve it in another way? How about we let helpers access to the current main objects? That way the libraries can make that decision themselves. (I'd reserve those changes until we receive actual requests from library maintainers though)

What I'd also mention is that if we make some commands strictly on the main object, it could be a break change for some people.

bkuhlmann:
Out of curiosity, why the use of `IRB::Helper::Base#execute`? Could `IRB::Helper::Base#call` be the Object API instead? This would allow anyone to use the [command pattern](https://alchemists.io/articles/command_pattern) which would give folks more power to register simple closures if desired. :rocket:

st0012:
@bkuhlmann That's a great point. I picked `execute` because it's the current API on (still private) command objects, but I don't feel strongly about it.

> would give folks more power to register simple closures if desired

Could you elaborate on this?

bkuhlmann:
**Stan**: Yeah, so I was thinking this would be neat in terms of closures:

``` ruby
# Proc
demo = proc { |*positionals, **keywords, &block|  puts "TOOD: Implementation details." }
IRB::Helper.register :demo, demo

# Lamda
demo = -> *positionals, **keywords, &block { "TOOD: Implementation details." }
IRB::Helper.register :demo, demo
```

Then, once the closure is registered, you could send the `#call` message to the registered helper. This could be quite handy for simple helpers. Anything more than that would require using a class which implements the `#call` message. Regardless, any implementation would be interchangeable as long as it adhered to the `#call` message.

I'm making a couple of assumptions, though, which I don't believe matches up with your current implementation:

- Each helper, once registered, would only need to receive the `#call` message. Currently, it looks like you want to send the `.new` message to each helper and then, later, send the `#execute` message.
- The major disadvantage, because a closure wouldn't inherit from `Base`, would mean there would be no description or context. That could be bad but...could work like Rake where if missing a `desc` shows up as hidden. Probably not desirable but food for thought.

Anyway, some thoughts. Given the value of inheriting from the `Base` class, the above might not be desirable. ...but using the `#call` message instead of `#execute` might still be of interest.

st0012:
@bkuhlmann Thanks for the explanation 👍 

Yeah under the current plan just passing a Proc in wouldn't work because of `.new`. We can hide `Helper.new.call` under `Helper::Base.call` to make it work though.

But as you said, not having description would be a huge disadvantage. Since one of the main goals here is to make helpers more discoverable through these extra metadata, supporting this feel like a step backward to me.

I don't plan to support it right away, but I'm open to other people's feedback on this idea. And thank you so much for taking the time to propose it 🙌

st0012:
@tompng I think this can go hand in hand with the command extension API. WDYT?

lib/irb/default_commands.rb
```diff
@@ -236,6 +229,27 @@ def self.load_command(command)
       nil
     end
 
+    def self.install_helper_methods
+      HelperMethod.helper_methods.each do |name, helper_method_class|
+        define_method name do |*args, **opts, &block|
+          verbose, $VERBOSE = $VERBOSE, nil
```
tompng, line 235: What kind of warning need to be suppressed here?
Shouldn't we show a warning occurred inside `helper.execute` implementation?

lib/irb/default_commands.rb
```diff
@@ -236,6 +229,27 @@ def self.load_command(command)
...
+          helper_ivar = "@_helper_method_#{name}".to_sym
+          helper = instance_variable_get(helper_ivar)
+
+          unless helper
+            helper = helper_method_class.new
+            instance_variable_set(helper_ivar, helper)
+          end
+
+          helper.execute(*args, **opts, &block)
```
tompng, line 244: Helper method is an instance method of `workspace.main`. If we pass `self` to execute, perhaps helper_method can do more.
```ruby
helper.execute(self, *args, **opts, &block)
```
We can get a similar value by `IRB.CurrentContext.workspace.main` except for this case: `$a=self` `chws ""` `$a.conf` (`main==""` but `self==$a`), so we don't need to pass it?
What do you think?

lib/irb/default_commands.rb
```diff
@@ -236,6 +229,27 @@ def self.load_command(command)
...
+    def self.install_helper_methods
+      HelperMethod.helper_methods.each do |name, helper_method_class|
+        define_method name do |*args, **opts, &block|
+          verbose, $VERBOSE = $VERBOSE, nil
+          helper_ivar = "@_helper_method_#{name}".to_sym
+          helper = instance_variable_get(helper_ivar)
+
+          unless helper
+            helper = helper_method_class.new
+            instance_variable_set(helper_ivar, helper)
```
tompng, line 241: Command execution uses `new(irb_context).execute(arg)` and does not cache the instance.
I think HelperMethod don't need to cache too, or should cache in another place to avoid adding instance variables to `workspace.main`
```ruby
From: a.rb @ line 2 :

    1: class A
 => 2:   def f = binding.irb
    3: end

irb(#<A:0x000000013019fb00>):001> conf;context;irb_context;
irb(#<A:0x000000013019fb00>):002> self
=> 
#<A:0x000000013019fb00
 @_helper_method_conf=#<IRB::HelperMethod::Conf:0x0000000127985878>,
 @_helper_method_context=#<IRB::HelperMethod::Context:0x0000000127985670>,
 @_helper_method_irb_context=#<IRB::HelperMethod::IrbContext:0x00000001279854b8>>
```

tompng:
Can we leave `context` and `conf` as a command and have no helper method by default?
We can add `IRB.context` as an alias of `IRB.CurrentContext` and use it instead of these helper methods.

I think helper method have a large benefit, and also have problem of polluting object with chws and debug. IRB uses SimpleDelegator for frozen object, so `chws 1` `[self, 1].uniq` returns `[1, 1]`.
For this reason, I'd like to make default IRB clean, and have a stance that no-helper is recommended. User can compare pros and cons, and decide adding helper method or not.

### Other thoughts
Rails's `reload!` can be a command. `app` `helper` `controller` `new_session` does not need to be used from model(`chws User.first`). It can be a method of main object. Perhaps adding to Kernel is better than polluting by helper method. Taking this into account, one possible steps of upgrading commands/helpers might be:

1. Add helper_method API without default helper methods
2. Add a way to show help message for method defined in Kernel or main object
3. Rails can migrate some helper_method to main method with help message added in step 2. Other apps can still use helper_method.

If the demand of helper_method after step 3 is assumed to be low, we can deprecate helper_method after it, or even skip step 1 and implement step 2.

lib/irb/default_commands.rb
st0012: I want to memoize the helper method instance, but then in Ruby 2.7 it'd cause the ivar not initialized warning.

st0012:
> Can we leave context and conf as a command and have no helper method by default?

I think at least `conf` has to be a helper method as the current documents all reference to it for changing configs inside the session ([example](https://docs.ruby-lang.org/en/master/IRB.html#module-IRB-label-IRB+Name))? And commands should NOT return values (or just `nil`) so they're not suitable for this usage.

>  I think helper method have a large benefit, and also have problem of polluting object with chws and debug.

I feel we're talking about 2 related, but not identical problems. While polluting the current context is not ideal, it's already been done in contexts like Rails console to provide helper features for years. And another problem is that those features are hard to discover.

Through this PR, my main goal is to only solve the feature discovery problem because this was a feedback from Rails maintainers. And [this type](https://twitter.com/RubyCademy/status/1778845779932729707) of recurring Rails tips also demonstrates how beneficial it'd be if we can improve it. 

At the same time though, I never heard about people complaining about context pollution, so to me it could be solved much later. And I believe that by introducing official registration APIs first, it'll also make solving the pollution much easier.

tompng:
> at least `conf` has to be a helper method

Ok, I think we can implement it as a helper_method now and make it show a deprecation message later. Let's go ahead!

tompng (approved):
Looks good to me :+1:

PR #628, opened 2023-07-04
Use `max_by` for `longest_cmd_name_length`

andyw8:
A small refactoring for readability.

https://rubyapi.org/3.2/o/enumerable#method-i-max_by

### DIFF
--- a/lib/irb/cmd/show_cmds.rb
+++ b/lib/irb/cmd/show_cmds.rb
@@ -14,7 +14,7 @@ module IRB
       def execute(*args)
         commands_info = IRB::ExtendCommandBundle.all_commands_info
         commands_grouped_by_categories = commands_info.group_by { |cmd| cmd[:category] }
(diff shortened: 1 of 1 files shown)


lib/irb/cmd/show_cmds.rb
```diff
@@ -14,7 +14,7 @@ class ShowCmds < Nop
       def execute(*args)
         commands_info = IRB::ExtendCommandBundle.all_commands_info
         commands_grouped_by_categories = commands_info.group_by { |cmd| cmd[:category] }
-        longest_cmd_name_length = commands_info.map { |c| c[:display_name] }.max { |a, b| a.length <=> b.length }.length
+        longest_cmd_name_length = commands_info.map { |c| c[:display_name] }.max_by(&:length).length
```
tompng, line 17: How about
`map { |c| c[:display_name] }.map(&:length).max`
or
`map { |c| c[:display_name].length }.max`

andyw8: Nice, I've pushed an update.

st0012 (approved):
Thanks 👍

Issue #627, opened 2023-07-04
Multi-line multibyte regular expression raises exception


elfham:
## Description

An exception occurs when attempting to enter a multi-line multi-byte regular expression.

#### Scenario:

Type `/`, <kbd>Enter</kbd>, `あ`, <kbd>Enter</kbd>, `/`, <kbd>Enter</kbd>.

#### Expected:

```
% irb
irb(main):001:0/ /
irb(main):002:0/ あ
irb(main):003:0> /
=>
/
\u3042
/
irb(main):004:0>
```

#### Actual:

Type `/`, <kbd>Enter</kbd>, `あ`, <kbd>Enter</kbd>, then,

```
% irb
irb(main):001:0/ /
irb(main):002:0/ あ/var/work/src/irb_ruby/lib/irb/ruby-lex.rb:97:in `match?': invalid byte sequence in UTF-8 (ArgumentError)
        from /var/work/src/irb_ruby/lib/irb/ruby-lex.rb:97:in `block in configure_io'
        from /var/work/lib/ruby/reline/line_editor.rb:1648:in `process_auto_indent'
        from /var/work/lib/ruby/reline/line_editor.rb:1603:in `input_key'
        from /var/work/lib/ruby/reline.rb:349:in `block (3 levels) in inner_readline'
        from /var/work/lib/ruby/reline.rb:348:in `each'
        from /var/work/lib/ruby/reline.rb:348:in `block (2 levels) in inner_readline'
        from /var/work/lib/ruby/reline.rb:423:in `block in read_io'
        from /var/work/lib/ruby/reline.rb:393:in `loop'
        from /var/work/lib/ruby/reline.rb:393:in `read_io'
        from /var/work/lib/ruby/reline.rb:346:in `block in inner_readline'
        from /var/work/lib/ruby/reline.rb:344:in `loop'
        from /var/work/lib/ruby/reline.rb:344:in `inner_readline'
        from /var/work/lib/ruby/reline.rb:273:in `block in readmultiline'
        from /var/work/lib/ruby/reline/ansi.rb:149:in `block in with_raw_input'
        from /var/work/lib/ruby/reline/ansi.rb:149:in `raw'
        from /var/work/lib/ruby/reline/ansi.rb:149:in `with_raw_input'
        from /var/work/lib/ruby/reline.rb:269:in `readmultiline'
        from /usr/local/lib/ruby/3.1/forwardable.rb:238:in `readmultiline'
[...]

PR #631, opened 2023-07-04
Fix invalid byte sequence in autoindent calculation

tompng:
Workaround for https://github.com/ruby/irb/issues/627

The root cause of this issue is several (more than two) bugs in Reline.
https://github.com/ruby/reline/issues/553

Reline sometimes passes wrong byte_pointer to auto_indent_proc.
If Reline's bug is not going to be fixed soon, IRB needs to validate encoding of `byteslice(0, byte_pointer)` before using it.

### DIFF
--- a/test/irb/test_ruby_lex.rb
+++ b/test/irb/test_ruby_lex.rb
@@ -33,7 +33,7 @@ module TestIRB
     def calculate_indenting(lines, add_new_line)
       lines = lines + [""] if add_new_line
       last_line_index = lines.length - 1
-      byte_pointer = lines.last.length
+      byte_pointer = lines.last.bytesize
 
       context = build_context
       context.auto_indent_mode = true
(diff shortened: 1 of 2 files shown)


test/irb/test_ruby_lex.rb
```diff
@@ -33,7 +33,7 @@ def teardown
     def calculate_indenting(lines, add_new_line)
       lines = lines + [""] if add_new_line
       last_line_index = lines.length - 1
-      byte_pointer = lines.last.length
+      byte_pointer = lines.last.bytesize
```
st0012, line 36: The test seems to capture the issue even without this change?

st0012 (approved):
LGTM, just once question

test/irb/test_ruby_lex.rb
tompng: Yes. This change don't affect the added test. I just fixed it because I found it doing wrong with byte_pointer.

tompng:
I think https://github.com/ruby/reline/pull/562 will also fix this problem.

tompng:
Closing this because this pull request is a workaround, and it's fixed in Reline 0.3.6

PR #633, opened 2023-07-08
Avoid using Reline as a module

st0012:
Using it as a module makes it harder to understand methods' source.


ima1zumi (approved):
Thank you 👍

PR #634, opened 2023-07-08
Drop src_encoding.rb

st0012:
Its method `IRB.default_src_encoding` was only used in `magic-file.rb`, which has been removed.


ima1zumi (approved):
👍

PR #635, opened 2023-07-08
Remove unused `InputMethod#initialize`

st0012:
The constructor takes a `file_name` argument, but it is never used. The only input method that needs a file is `FileInputMethod`, which has its own constructor to take a file object directly.

So the constructor in `InputMethod` is not needed and its child classes don't need to call `super` in their constructors.

I also dropped the attribute accessors for `file_name` as they're not used either.

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -10,16 +10,7 @@ require 'io/console'
 require 'reline'
 
 module IRB
-  STDIN_FILE_NAME = "(line)" # :nodoc:
   class InputMethod
-
-    # Creates a new input method object
-    def initialize(file = STDIN_FILE_NAME)
-      @file_name = file
-    end
-    # The file name of this input method, usually given during initialization.
-    attr_reader :file_name
-
     # The irb prompt associated with this input method
     attr_accessor :prompt
 
@@ -56,7 +47,6 @@ module IRB
   class StdioInputMethod < InputMethod
     # Creates a new input method object
     def initialize
-      super
       @line_no = 0
       @line = []
       @stdin = IO.open(STDIN.to_i, :external_encoding => IRB.conf[:LC_MESSAGES].encoding, :internal_encoding => "-")
@@ -130,7 +120,6 @@ module IRB
 
     # Creates a new input method object
     def initialize(file)
-      super
       @io = file.is_a?(IO) ? file : File.open(file)
       @external_encoding = @io.external_encoding
     end
@@ -183,7 +172,6 @@ module IRB
         if Readline.respond_to?(:encoding_system_needs)
           IRB.__send__(:set_encoding, Readline.encoding_system_needs.name, override: false)
         end
-        super
 
         @line_no = 0
         @line = []
@@ -261,7 +249,6 @@ module IRB
     # Creates a new input method object using Reline
     def initialize
       IRB.__send__(:set_encoding, Reline.encoding_system_needs.name, override: false)
-      super
 
       @line_no = 0
       @line = []
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -123,8 +123,6 @@ module IRB
       @io = file.is_a?(IO) ? file : File.open(file)
       @external_encoding = @io.external_encoding
     end
-    # The file name of this input method, usually given during initialization.
-    attr_reader :file_name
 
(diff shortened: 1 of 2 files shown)


lib/irb/input-method.rb
```diff
@@ -10,16 +10,7 @@
 require 'reline'
 
 module IRB
-  STDIN_FILE_NAME = "(line)" # :nodoc:
   class InputMethod
-
-    # Creates a new input method object
-    def initialize(file = STDIN_FILE_NAME)
```
ima1zumi, line 17: In a search on GitHub, `InputMethod.new` was found to be used in two repositories.
I believe it would be better to show a deprecation warning before removal.

https://github.com/search?q=%22+InputMethod.new%22+language%3ARuby&type=code

st0012: It's thoughtful to think about external dependants when dropping things 👍  But in this case I think we should not be blocked by them because:

1. `irb_remote` hasn't received any update for 5 years and it already passes incorrect number of arguments to `InputMethod.new`. So it's already broken.
2. In `irb-power_assert`, it's used to build an IRB environment for testing incorrectly. I say incorrectly because `InputMethod` is an abstract class. So in practice it should have not been initialised directly. If the author switch to `StdioInputMethod` instead, that may be better and more stable.

But the root problem is that we and previous maintainers don't define what's official APIs, what's not, and ways/principles to achieve what they need without using private APIs.

My current categorisation is:

- Public APIs: 
    - Methods to start an IRB session: `binding.irb` and `IRB.start`
    - IRB commands
    - (Maybe more)
- Should be private but unfortunately got exposed extensively APIs:
    - `IRB::ExtendCommandBundle`
        - I want to fix it with #513 and #624 
    - [`IRB::Irb`](https://github.com/search?q=Irb.new&type=code)
- Private APIs:
    - All the things that are not included in the above categories

So IMO `InputMethod`, or even its child classes like `RelineInputMethod`...etc., should be considered private. Otherwise, our work on refactoring IRB would be seriously affected.

What are your thoughts? @tompng @hasumikin

ima1zumi: Thanks for the explanation. I understand, and it does not seem to be a problem to delete InputMethod.new.
However, I think we should carefully migrate the other InputMethods since they are widely used.
Not all of the code is IRB-related, but there seems to be a lot there. (Not sure if they have been active recently)
https://github.com/search?q=InputMethod.new+language%3ARuby+&type=code

tompng: I think the refactor looks good. InputMethod should not have FileInputMethod specific attribute.

My opinion for public and private are:

Sub classes: IMO should be private forever. There are alternatives.
```ruby
RelineInputMethod # Only used in ruby/irb and it's forked repository
StdioInputMethod # Some repository using this is too old. last update: 12 years ago

# Alternative of IRB::Irb.new(workspace, FileInputMethod.new(file_name))
IRB::Irb.new(workspace, file_name)

# Alternative of IRB::Irb.new(workspace, ReadlineInputMethod.new)
IRB.conf[:USE_MULTILINE] = false
```

InputMethod: Maybe we can make it public in the future. But now, there is no way to set output method. It's hard to implement something like remote IRB using socket because users can only overwrite input. So the use case of custom input method is very limited.

`IRB::Irb#initialize` had ouput_method but it was unused/ and removed in https://github.com/ruby/irb/commit/53f776974d1d84c67a01f69b17fe5ddefa63194f

tompng (approved):

lib/irb/input-method.rb
kachick: Thanks for your consideration, I was facing CI failures with these changes, but this discussion helps me a lot.
[I just replaced `s/InputMethod/IRB::StdioInputMethod/` for now.](https://github.com/kachick/irb-power_assert/commit/7888bde9a62f54e12faf09a454f210e1e6f9b0d9)

Feel free to leave care about irb-power-assert anymore, I will catch up IRB changes in https://github.com/ruby/irb/pull/673 if needed.

st0012: @kachick Thank you for the feedback and sorry for breaking your project's CI. 
After taking a closer look, I think the test setup actually doesn't need to specify an input method at all. So I opened https://github.com/kachick/irb-power_assert/pull/114 and I hope it will simplify its maintenance in the long run 🙂

PR #640, opened 2023-07-12
Unpend RDoc dialog related tests

st0012:
They are our only coverage on the doc dialog of autocompletion. Without them, issues like #638 would slip in unnoticed.

A few notes about these tests:

- To make them work properly, RI data is needed but RDoc HTML is not. Hence `gem rdoc --all --ri --no-rdoc`
- There's an output difference caused by different keyboard shortcut hint on mac and non-mac OS
- For some reason installing RI data still doesn't make it work under Ruby 3.1. But I think coverage on 3.1+ is enough so I just skip them on Ruby 2.7 and 3.0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ on:
   push:
   pull_request:
   schedule:
-    - cron: '30 14 * * *'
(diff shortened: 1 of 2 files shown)


tompng (approved):
:+1:

PR #642, opened 2023-07-14
Fix history-saving feature

### LINKED ISSUE #641: irb stopped saving history since 1.7.2 on macOS M2 Pro Ventura 13.4.1
## Description

1.7.1 saves history for me just fine as part of rails console and standalone, 1.7.2 and 1.7.3 won't save history with or without `~/.irbrc` config

## Result of irb_info

Please paste the result of `irb_info` command in IRB.

Ruby version: 3.1.3
IRB version: irb 1.7.3 (2023-07-12)
.irbrc path: ~/.rvm/rubies/ruby-3.1.3/.irbrc
RUBY_PLATFORM: arm64-darwin21
LANG env: en_US.UTF-8
East Asian Ambiguous Width: 1

## Terminal Emulator

macOS zsh

## Setting Files

No

st0012:
The `HistorySavingAbility` module doesn't do anything if the input method class doesn't define `HISTORY`.

- https://github.com/ruby/irb/blob/3ac96be660bf052902fb4e532c7a46cf294b71eb/lib/irb/history.rb#L10
- https://github.com/ruby/irb/blob/3ac96be660bf052902fb4e532c7a46cf294b71eb/lib/irb/history.rb#L34

Which caused the change in #633 breaks history saving silently.

This patch defines `RelineInputMethod::HISTORY` to avoid this.

Also, instead of checking the existence of `input_method_class::HISTORY`, we should make every input method class declare if it supports history saving or not. With this approach, similar issues would be caught by tests in the future.

Since the default value is `false`, it shouldn't break any custom input method that inherits from `IRB::InputMethod`.

Closes #641


lib/irb/context.rb
```diff
@@ -154,8 +154,12 @@ def initialize(irb, workspace = nil, input_method = nil)
 
     def save_history=(val)
       IRB.conf[:SAVE_HISTORY] = val
+
       if val
-        (IRB.conf[:MAIN_CONTEXT] || self).init_save_history
+        context = (IRB.conf[:MAIN_CONTEXT] || self)
+        if context.io.support_history_saving? && !context.io.singleton_class.include?(HistorySavingAbility)
+          context.io.extend(HistorySavingAbility)
```
st0012, line 161: I decided to put these in the same place to make future refactor easier.

tompng (approved):

modosc:
thanks for the quick fix - is this going to get a release today?

PR #647, opened 2023-07-16
Display `show_cmds`'s output in a pager when in TTY environment

### LINKED ISSUE #646: Display `show_cmds`'s output in pager
The list of commands is already pretty long and often requires scrolling up to see the top ones. This is especially true when using IRB in editors' terminals. 

Given IRB will open to helper/command extension, the list could grow even longer. Therefore, we should consider displaying commands in a pager to provide better experience.

st0012:
This can:

- Make it easier to scroll up and down the commands list
- Avoid pushing up users' previous output
- Allow users to do basic search with `/<word>`

### Demo

https://github.com/ruby/irb/assets/5079556/6c71c514-3c16-4516-b820-64c10b2d0e0d

Closes #646

### DIFF
--- a/lib/irb/cmd/show_cmds.rb
+++ b/lib/irb/cmd/show_cmds.rb
@@ -2,6 +2,7 @@
 
 require "stringio"
 require_relative "nop"
+require_relative "../pager"
 
 module IRB
   # :stopdoc:
@@ -28,9 +29,9 @@ module IRB
           output.puts
         end
 
-        puts output.string
-
-        nil
+        Pager.page do |io|
+          io.puts output.string
+        end
       end
     end
   end
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+module IRB
+  # The implementation of this class is borrowed from RDoc's lib/rdoc/ri/driver.rb.
+  # Please do NOT use this class directly outside of IRB.
+  class Pager
+    PAGE_COMMANDS = [ENV['RI_PAGER'], ENV['PAGER'], 'less', 'more'].compact.uniq
+
+    class << self
+      def page
+        if STDIN.tty? && pager = setup_pager
+          begin
+            pid = pager.pid
+            yield pager
+          ensure
+            pager.close
+          end
+        else
+          yield $stdout
+        end
+      # When user presses Ctrl-C, IRB would raise `IRB::Abort`
+      # But since Pager is implemented by running paging commands like `less` in another process with `IO.popen`,
+      # the `IRB::Abort` exception only interrupts IRB's execution but doesn't affect the pager
+      # So to properly terminate the pager with Ctrl-C, we need to catch `IRB::Abort` and kill the pager process
+      rescue IRB::Abort
+        Process.kill("TERM", pid) if pid
+        nil
+      rescue Errno::EPIPE
+      end
+
+      private
+
+      def setup_pager
+        require 'shellwords'
+
+        PAGE_COMMANDS.each do |pager|
+          pager = Shellwords.split(pager)
+          next if pager.empty?
+
+          if pager.first == 'less' || pager.first == 'more'
+            pager << '-R' unless pager.include?('-R')
+          end
+
+          begin
+            io = IO.popen(pager, 'w')
+          rescue
+            next
+          end
+
+          if $? && $?.pid == io.pid && $?.exited? # pager didn't work
+            next
+          end
+
+          return io
+        end
+
+        nil
+      end
+    end
+  end
+end
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -688,6 +688,16 @@ module TestIRB
 
 
   class ShowCmdsTest < CommandTestCase
+    def setup
+      STDIN.singleton_class.define_method :tty? do
+        false
+      end
+    end
+
+    def teardown
+      STDIN.singleton_class.remove_method :tty?
+    end
+
     def test_show_cmds
       out, err = execute_lines(
         "show_cmds\n"
--- a/test/irb/yamatanooroti/test_rendering.rb
+++ b/test/irb/yamatanooroti/test_rendering.rb
@@ -252,6 +252,24 @@ class IRB::RenderingTest < Yamatanooroti::TestCase
     EOC
   end
 
+  def test_show_cmds_with_pager_can_quit_with_ctrl_c
+    write_irbrc <<~'LINES'
+      puts 'start IRB'
+    LINES
+    start_terminal(40, 80, %W{ruby -I#{@pwd}/lib #{@pwd}/exe/irb}, startup_message: 'start IRB')
+    write("show_cmds\n")
+    write("G") # move to the end of the screen
+    write("\C-c") # quit pager
+    write("'foo' + 'bar'\n") # eval something to make sure IRB resumes
+    close
+
+    screen = result.join("\n").sub(/\n*\z/, "\n")
+    # IRB::Abort should be rescued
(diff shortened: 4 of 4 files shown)


tompng:
Handling Ctrl-C works great :+1:

In my environment, `ESC[1m` is shown.
less_result
It works if I set `ENV['PAGER'] = 'less -r'`.

In `show_doc String`, rdoc is using non-ansi format `= S\bSt\btr\bri\bin\bng\bg \b <\b< \b O\bOb\bbj\bje\bec\bct\bt\n\n` for  string_doc

lib/irb/cmd/show_cmds.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "stringio"
+require "rdoc/ri/driver"
```
tompng, line 4: Do we need to wrap this with `begin rescue` like irb is doing in completion?
Or this file is delayed loaded so we don't need to rescue it? (maybe the loading timing will change in the future?)

tompng: The pager implementation part of `rdic/ri/driver` is about 30 lines of 1500 lines.
Maybe we could implement it in IRB? then we can customize the command line option like
```ruby
# https://github.com/ruby/rdoc/blob/5db4767dda4b4320eeac84aba152ab56ec20b839/lib/rdoc/ri/driver.rb#L1470
pagers = [ENV['RI_PAGER'], ENV['PAGER'], 'pager', 'less', 'more']
↓ 
# IRB's ansi format pager
pagers = [ENV['RI_PAGER'], ENV['PAGER'], 'pager --foobar', 'less --raw-control-chars', 'more --raw-control-chars']
```

st0012: > Do we need to wrap this with begin rescue like irb is doing in completion?

Since we've declared `rdoc` as a dependency in #648, `rdoc` should always be installed along with `irb`. If that didn't happen for some reason, it's something else we need to fix.

Put it in another way: I think we can remove the rescue block in completion and other places.

@hsbt can you help me confirm that if my assumption above is correct? Or when `irb` is used as a default gem, `rdoc` could still be missing? ([example](https://github.com/rails/rails/issues/44143#issuecomment-1013663083))

> Maybe we could implement it in IRB? then we can customize the command line option like

Generally I agree with this direction, but I found a few concerns when implementing it:

- Should `IRB::Pager` also accept `ENV['RI_PAGER']`?
    - If it doesn't, users would get inconsistent behaviour where IRB's pager doesn't pickup their config for RI pager
    - If it does, it sounds like a weird coupling with `rdoc`?
- Should `IRB::Pager` set different pager options (`-r` for example) than `rdoc`'s?
    - This will bring similar concerns the above question

I think if we want to avoid using `rdoc`'s pager, we should do that for all features so we can get rid off of the above issues. For example, IRB should only use `rdoc` to get documentation content and handle the display itself. But at the same time the effort would grow significantly.

st0012:
> In my environment, ESC[1m is shown.

Can you describe your environment setup? I can't reproduce it on my machine with VS Code Terminal, iTerm2, or Terminal.app

> In show_doc String, rdoc is using non-ansi format = `S\bSt\btr\bri\bin\bng\bg \b <\b< \b O\bOb\bbj\bje\bec\bct\bt\n\n for`

Sorry I don't get this part. Do you mean you also need to set `less -r` for RDoc so it displays `show_doc`'s output correctly?

tompng:
I'm using mac pre-installed less
```
% less --version
less 581.2 (POSIX regular expressions)
Copyright (C) 1984-2021  Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Home page: https://greenwoodsoftware.com/less
```
and I can reproduce it with
```
% docker run --rm -ti rubylang/rubyfarm sh
# git clone https://github.com/ruby/irb.git
# cd irb
# git switch implement-#646
# bundle install
# bin/console
irb(main):001:0> show_doc String
=> Works
irb(main):002:0> show_cmds
=> ESC[1mIRBESC[0m
```


`show_doc` works and doesn't need `-r` because `show_doc` don't use ansi escape sequence like `\e[1mBOLD\e[m` but uses another format `B\bBO\bOL\bLD\bD`.
```
echo "B\bBO\bOL\bLD\bD" | less
echo "\e[1mBOLD\e[m" | less
echo "\e[1mBOLD\e[m" | less -r
```

st0012:
That's weird because I'm using the same `less` version:

```
less 581.2 (POSIX regular expressions)
Copyright (C) 1984-2021  Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Home page: https://greenwoodsoftware.com/less
```

And with it, these 3 examples all rendered correctly:

```
echo "B\bBO\bOL\bLD\bD" | less
echo "\e[1mBOLD\e[m" | less
echo "\e[1mBOLD\e[m" | less -r
```

If I combine the 2 versions:

```
echo "B\bBO\bOL\bLD\bD\n\e[1mBOLD\e[m" | less
```

I get:

Screenshot 2023-07-22 at 13 45 00

st0012:
It turns out because I use ohmyzsh, which sets `LESS=-R`, the output is displayed correctly on my machine.

lib/irb/cmd/show_cmds.rb
st0012: Because it's necessary to pass `-r` to `less` in order to correctly display IRB's ANSI escape sequences, I eventually decided to implement `IRB::Pager` with the support of `ENV['RI_PAGER']` for now.

I think we can later replace the use of RDoc pager in `show_doc` and doc dialog with something like this:

```rb
name = "String"
driver = RDoc::RI::Driver.new({})
found, klasses, includes, extends = driver.classes_and_includes_and_extends_for(name)
doc = driver.class_document(name, found, klasses, includes, extends)
formatter = RDoc::Markup::ToAnsi.new
contents = doc.accept(formatter)
IRB::Pager.page do |io|
  io.puts contents
end
IRB::Pager.page do |io|
  io.puts contents
end
```

lib/irb/pager.rb
```diff
@@ -0,0 +1,61 @@
...
+      def setup_pager
+        require 'shellwords'
+
+        PAGE_COMMANDS.each do |pager|
+          pager = Shellwords.split(pager)
+          next if pager.empty?
+
+          if pager.first == 'less' || pager.first == 'more'
+            pager << '-R' unless pager.include?('-R')
+          end
```
st0012, line 42: If the user has `PAGER=less` set, we still want to make sure `-R` or any future flags are appended to the command so the content could be displayed properly without asking users to change things.

lib/irb/pager.rb
```diff
@@ -0,0 +1,61 @@
...
+
+module IRB
+  # The implementation of this class is borrowed from RDoc's lib/rdoc/ri/driver.rb.
+  # Please do NOT use this class directly outside of IRB.
+  class Pager
+    PAGE_COMMANDS = [ENV['RI_PAGER'], ENV['PAGER'], 'less', 'more'].compact.uniq
+
+    class << self
+      def page
+        if STDIN.tty? && pager = setup_pager
```
st0012, line 11: I'm not sure if we should check `STDIN.tty?` or `STDOUT.tty?`. In most cases they should return the same value?

tompng: :+1: It's consistent with IRB shows message `Switch to inspect mode.` when stdin is not a tty.


`irb` → both tty
`irb > output_file` → stdin is a tty
`cat input_file | irb` → stdout is a tty
`cat input_file | irb > output_file` → neither is tty

tompng (approved):

PR #651, opened 2023-07-19
Handle Concurrent Sessions and Saving Readline::HISTORY

### LINKED ISSUE #510: Cannot save to history file in nomultiline mode with concurrent irb session
## Description

When I use `irb --nomultiline` concurrently and enter `exit` in each irb session, irb raises error

```
[tomoya:irb]% bin/console --nomultiline
irb(main):001:0> `bin/console`.size
exit
=> 599
irb(main):002:0> exit
/path/to/irb/lib/irb/ext/save-history.rb:107:in `[]': no implicit conversion of Range into Integer (TypeError)

          history = history[@loaded_history_lines..-1] if @loaded_history_lines
                            ^^^^^^^^^^^^^^^^^^^^^^^^^
	from /path/to/irb/lib/irb/ext/save-history.rb:107:in `save_history'
	from /path/to/irb/lib/irb/ext/save-history.rb:55:in `block in extended'
	from /path/to/irb/lib/irb.rb:499:in `block in run'
	from /path/to/irb/lib/irb.rb:499:in `each'
	from /path/to/irb/lib/irb.rb:499:in `run'
	from /path/to/irb/lib/irb.rb:416:in `start'
	from bin/console:6:in `<main>'
```

## Result of irb_info

```
irb(main):001:0> irb_info
=> 
Ruby version: 3.2.0
IRB version: irb 1.6.2 (2022-12-13)
InputMethod: RelineInputMethod with Reline 0.3.2
LANG env: ja_JP.UTF-8
East Asian Ambiguous Width: 1
```

## Terminal Emulator

Terminal.app

## Setting Files

> Are you using `~/.irbrc` and `~/.inputrc`?

No

chadrschroeder:
Fixes #510.


lib/irb/history.rb
```diff
@@ -51,7 +51,7 @@ def save_history
 
         if File.exist?(history_file) &&
            File.mtime(history_file) != @loaded_history_mtime
-          history = history[@loaded_history_lines..-1] if @loaded_history_lines
+          @loaded_history_lines&.times { history.delete_at(0) }
```
chadrschroeder, line 54: Maybe there's a better way to do this but `Readline::HISTORY` is pretty limited in how it can be manipulated.

The original issue here was a `no implicit conversion of Range into Integer` error because `Readline::HISTORY` doesn't take a range in `[]`.

test/irb/test_history.rb
```diff
@@ -14,12 +14,22 @@ def teardown
       IRB.conf[:RC_NAME_GENERATOR] = nil
     end
 
-    class TestInputMethodWithHistory < TestInputMethod
-      HISTORY = Array.new
+    class TestInputMethodWithRelineHistory < TestInputMethod
+      # When IRB.conf[:USE_MULTILINE] is true, IRB::ReidlineInputMethod uses Reline::History
+      HISTORY = Reline::History.new(Reline.core.config)
```
chadrschroeder, line 19: Changed from an Array to the history class that is used by default when `:USE_MULTILINE` is on.

test/irb/test_history.rb
```diff
@@ -14,12 +14,22 @@ def teardown
...
 
       include IRB::HistorySavingAbility
     end
 
+    class TestInputMethodWithReadlineHistory < TestInputMethod
+      # When IRB.conf[:USE_MULTILINE] is false, IRB::ReadlineInputMethod uses Readline::HISTORY
+      HISTORY = Readline::HISTORY
```
chadrschroeder, line 26: I created an additional input method test class that uses `Readline::HISTORY`.  At this point, I've only made use of it in the `test_history_concurrent_use` test method to illustrate the bug.  I wasn't sure if we want to test both input methods for all tests.

test/irb/test_history.rb
```diff
@@ -138,10 +150,11 @@ def test_history_concurrent_use_not_present
       IRB.conf[:SAVE_HISTORY] = 1
       Dir.mktmpdir("test_irb_history_") do |tmpdir|
         ENV["HOME"] = tmpdir
-        io = TestInputMethodWithHistory.new
+        io = TestInputMethodWithRelineHistory.new
         io.class::HISTORY.clear
         io.load_history
-        io.class::HISTORY.concat(%w"line1 line2")
+        io.class::HISTORY << 'line1'
+        io.class::HISTORY << 'line2'
```
chadrschroeder, line 157: `concat` isn't available for `Readline::HISTORY`.  I updated this logic so that this test could use `TestInputMethodWithReadlineHistory` if we wanted.

test/irb/test_history.rb
```diff
@@ -169,15 +182,17 @@ def assert_history(expected_history, initial_irb_history, input)
...
-        io = TestInputMethodWithHistory.new
+        io = input_method.new
         io.class::HISTORY.clear
         io.load_history
         if block_given?
-          history = io.class::HISTORY.dup
+          previous_history = []
+          io.class::HISTORY.each { |line| previous_history << line }
           yield IRB.rc_file("_history")
-          io.class::HISTORY.replace(history)
```
chadrschroeder, line 178: `Readline::HISTORY` doesn't respond to `replace`.  Updated so that this works for both input methods.

test/irb/test_history.rb
```diff
@@ -169,15 +182,17 @@ def assert_history(expected_history, initial_irb_history, input)
...
+          io.class::HISTORY.clear
+          previous_history.each { |line| io.class::HISTORY << line }
         end
-        io.class::HISTORY.concat(input.split)
+        input.split.each { |line| io.class::HISTORY << line }
```
chadrschroeder, line 195: `Readline::HISTORY` doesn't respond to `concat`.  Updated how this adds the new lines.

lib/irb/history.rb
tompng: Unlike the previous code, `Reline::HISTORY` will be changed in this operation. Maybe it works, but it's safer and easy to maintain if it doesn't change HISTORY object itself.

chadrschroeder: Good point.  Updated.

lib/irb/history.rb
```diff
@@ -51,14 +52,19 @@ def save_history
...
            File.mtime(history_file) != @loaded_history_mtime
-          history = history[@loaded_history_lines..-1] if @loaded_history_lines
+          from = @loaded_history_lines
           append_history = true
         end
 
+        hist = []
+        from.upto(self.class::HISTORY.size - 1) do |index|
+          hist << self.class::HISTORY[index].split("\n").join("\\\n")
+        end
```
tompng, line 65: This code does not look rubyish compared to the original `history = history[@loaded_history_lines..-1] if @loaded_history_lines` that does not work.
How about changing `history` to an array first? then it will work fine for `Readline::HISTORY` too.

```ruby
def save_history
  history = self.class::HISTORY.to_a
```

test/irb/test_history.rb
```diff
@@ -17,8 +17,16 @@ def teardown
       IRB.conf[:RC_NAME_GENERATOR] = nil
     end
 
-    class TestInputMethodWithHistory < TestInputMethod
-      HISTORY = Array.new
+    class TestInputMethodWithRelineHistory < TestInputMethod
+      # When IRB.conf[:USE_MULTILINE] is true, IRB::ReidlineInputMethod uses Reline::History
```
tompng, line 21: `IRB::ReidlineInputMethod` → `IRB::RelineInputMethod`
`IRB::ReidlineInputMethod` is not used anymore.

Other part of the change in `test_history.rb` looks great :+1:

chadrschroeder: Thanks.  Updated.

lib/irb/history.rb
chadrschroeder: Done.

chadrschroeder (commented):
Updated.  Locally I'm getting a lot of test errors from `TestIRB::DebugCommandTest`, probably related to the changes from `master` that were merged into this branch.  I get the errors in `master` as well.  I haven't dug into why.  The CI testing appears to be running successfully.

tompng (approved):
Looks good :+1:
Thank you

PR #652, opened 2023-07-20
Fix nested sessions' history saving

### LINKED ISSUE #666: Running irb on debugger on irb dose not save the irb history
## Description

Running irb on debugger on irb dose not save the irb history.

### To Reproduce

1. Prepare a Ruby script `~/path_to_test_rb/test.rb` like this:

   ```ruby
   require "debug"

   class Test
       def run
           debugger
       end
   end
   ```

2. Remove `~/.irb_history` and `~/.rdbg_history` files to clean up
3. Run irb (`irb#1`)
4. Require `~/path_to_test_rb/test.rb` on irb
5. Execute `Test.new.run` on irb
6. Debugger start
7. Execute the irb command on debug
8. Start the second irb (`irb#2`)
9. Quit `irb#2`
10. Execute the continue command on debug
11. Quit `irb#1`
12. Execute `ls ~/.irb_history  ~/.rdbg_history`

For example, my case is:

```ruby
$ irb
irb(main):001:0> require "./test"
=> true
irb(main):002:0> Test.new.run

[1, 7] in ~/path_to_test_rb/test.rb
     1| require "debug"
     2|
     3| class Test
     4|     def run
=>   5|         debugger
     6|     end
     7| end
=>#0    Test#run at ~/path_to_test_rb/test.rb:5
  #1    <top (required)> at (irb):2
  # and 18 frames (use `bt' command for all frames)
(ruby) puts :debugger
debugger
nil
(rdbg) irb    # command
irb(#<Test:0x000000010d472dc8>):001:0> puts :irb
irb
=> nil
irb(#<Test:0x000000010d472dc8>):002:0> quit
(rdbg) continue    # continue command
=> #<Test:0x000000010d472dc8>
$ ls ~/.irb_history  ~/.rdbg_history
```

### Actual Result

```sh
$ ls ~/.irb_history  ~/.rdbg_history
ls: /Users/username/.irb_history: No such file or directory
/Users/username/.rdbg_history
```

### Expected behavior

```sh
$ ls ~/.irb_history  ~/.rdbg_history
/Users/username/.irb_history
/Users/username/.rdbg_history
```

## Additional Info

I have tested the following cases:

**1. Running only irb**

- I got the expected result
  - `irb_history` was saved

**2. Running irb on debugger executed by `ruby -e 'require "debug"; debugger'`**

- I got the expected result
  - `irb_history` and `.rdbg_history` were saved

**3. Running debug on irb**

- I got the expected result
  - `irb_history` and `.rdbg_history` were saved

**4. Running irb on irb**

- I got the expected result
  - `irb_history` was saved

## Result of irb_info, and debug/rdbg version

```sh
$ irb
irb(main):001:0> irb_info
Ruby version: 3.3.0
IRB version: irb 1.7.4 (2023-07-15)
RUBY_PLATFORM: x86_64-darwin22
LANG env: ja_JP.UTF-8
East Asian Ambiguous Width: 1

$ gem list | grep debug
debug (1.8.0)

$ rdbg -v
rdbg 1.8.0
```

## Terminal Emulator

```
iTerm2
Build 3.4.19
```

## Setting Files

No specific settings were applied.

## Additional Note

~This report is also planned to be submitted to [ruby/debug](https://github.com/ruby/debug).~
I have also reported this problem on ruby/debug#1009

And then,
I realized that this problem might be similar to #659 after making this report.

st0012:
Fix nested IRB sessions' history saving:
    
1. Dynamically including `HistorySavingAbility` makes things unnecessarily complicated and should be avoided.
2. Because both `Reline` and `Readline` use a single `HISTORY` constant to store history data. When nesting IRB sessions, only the first IRB session should handle history loading and saving so we can avoid duplicating history.
3. History saving callback should NOT be stored in `IRB.conf` as it's recreated every time `IRB.setup` is called, which would happen when nesting IRB sessions.

Fixes #666 #659

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -482,8 +482,15 @@ module IRB
     end
 
     def run(conf = IRB.conf)
+      in_nested_session = !!conf[:MAIN_CONTEXT]
       conf[:IRB_RC].call(context) if conf[:IRB_RC]
       conf[:MAIN_CONTEXT] = context
+
+      save_history = !in_nested_session && conf[:SAVE_HISTORY] && context.io.support_history_saving?
+
+      if save_history
+        context.io.load_history
+      end
 
       prev_trap = trap("SIGINT") do
         signal_handle
@@ -496,6 +503,7 @@ module IRB
       ensure
         trap("SIGINT", prev_trap)
         conf[:AT_EXIT].each{|hook| hook.call}
+        context.io.save_history if save_history
       end
     end
 
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -8,7 +8,6 @@ require_relative "workspace"
 require_relative "inspector"
 require_relative "input-method"
 require_relative "output-method"
-require_relative "history"
 
 module IRB
   # A class that wraps the current state of the irb session, including the
@@ -130,8 +129,6 @@ module IRB
       else
         @io = input_method
       end
-      self.save_history = IRB.conf[:SAVE_HISTORY] if IRB.conf[:SAVE_HISTORY]
-
       @extra_doc_dirs = IRB.conf[:EXTRA_DOC_DIRS]
 
       @echo = IRB.conf[:ECHO]
@@ -154,13 +151,6 @@ module IRB
 
     def save_history=(val)
       IRB.conf[:SAVE_HISTORY] = val
-
-      if val
-        context = (IRB.conf[:MAIN_CONTEXT] || self)
-        if context.io.support_history_saving? && !context.io.singleton_class.include?(HistorySavingAbility)
-          context.io.extend(HistorySavingAbility)
-        end
-      end
     end
 
     def save_history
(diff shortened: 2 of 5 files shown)


lib/irb/context.rb
```diff
@@ -154,13 +151,6 @@ def initialize(irb, workspace = nil, input_method = nil)
 
     def save_history=(val)
```
st0012, line 152: I don't think we need to have this `save_history` accessors on individual context objects, but it's better to keep them for compatibility.

lib/irb.rb
```diff
@@ -485,6 +485,11 @@ def run(conf = IRB.conf)
       conf[:IRB_RC].call(context) if conf[:IRB_RC]
       conf[:MAIN_CONTEXT] = context
 
+      if conf[:SAVE_HISTORY] && context.io.support_history_saving?
+        context.io.load_history
+        conf[:AT_EXIT].push proc{ context.io.save_history }
```
st0012, line 490: Note: `conf[:AT_EXIT]` will be called when `Irb#run` exits and is not associated with Ruby's `at_exit` callback.

tompng (approved):

Issue #653, opened 2023-07-22
Deprecate multi-irb feature


st0012:
## Plan

- Print deprecation warnings on these commands: `irb`, `jobs`, `fg`, and `kill` in the next minor release
- Completely remove those commands in `v2.0.0`

## Reasons

- It's concept highly overlaps with workspace commands like `chws`, `pushws`...etc.
- It spawns a new thread for every new new session, which conflicts with the [upcoming irb-debug integration](https://github.com/ruby/irb/pull/575)
- But most importantly, we can't find any practical use for this feature and the cost of maintaining its compatibility with other features is high
    - For example, it will require the solution of #650 to support multi-thread irb sessions too

## Feedback wanted

If you use the multi-irb feature and don't want to see it deprecated from IRB, please leave a comment below to explain your use case of it. It will help us assess this plan.

zenspider:
Please consider reusing the old command names. `chws` is just... not good.

kml:
Hi. I find `irb` command to be useful as an equivalent of [pry cd](https://github.com/pry/pry/blob/master/lib/pry/commands/cd.rb).
For example when debugging session starts in irb console (like rails console - testing on staging case), not by putting `binding.irb` in the tested file.

```ruby
bin/rails console

# assume it's part of the application
class Greeter
  def hello = "Hello, World!"
end

g = Greeter.new
irb g # <-----------------
Multi-irb commands are deprecated and will be removed in IRB 2.0.0. Please use workspace commands instead.
If you have any use case for multi-irb, please leave a comment at https://github.com/ruby/irb/issues/653
irb#1(#<Greeter:0x00007faa7b850188>):001> hello
=> "Hello, World!"
```

st0012:
@kml Thanks for the feedback 🙂 I'm adding a simpler `cd` command in #971

jwkoelewijn:
I love how this got implemented.... Now whenever I try to assign to an instance variable with the name `jobs` it now prints a message and doesn't actually create the var... brilliant.

ozydingo:
This implementation is a bit overly aggressive. As @jwkoelewijn points out, it breaks any code we might type or copy-paste that uses a `jobs` instance variable:

```
worker (local) > jobs = Job.all
Multi-irb commands are deprecated and will be removed in IRB 2.0.0. Please use workspace commands instead.
If you have any use case for multi-irb, please leave a comment at https://github.com/ruby/irb/issues/653
[...]

PR #654, opened 2023-07-22
Deprecate multi-irb commands

st0012:
This is part of #653 

- Print deprecated message when any of the commands are used

    ```
    irb(main):002:0> jobs
    Multi-irb commands are deprecated and will be removed in IRB 2.0.0.
    If you have any use case for multi-irb, please leave a comment at https://github.com/ruby/irb/issues/653
    => #0->irb on main (#<Thread:0x0000000104dab128 run>: running)
    ```

- Put related commands under `Multi-irb` category with a deprecated label
    
    ```
    Multi-irb (DEPRECATED)
    ```

### DIFF
--- a/lib/irb/cmd/subirb.rb
+++ b/lib/irb/cmd/subirb.rb
@@ -18,6 +18,13 @@ module IRB
 
       private
 
+      def print_deprecated_warning
+        warn <<~MSG
+          Multi-irb commands are deprecated and will be removed in IRB 2.0.0. Please use workspace commands instead.
+          If you have any use case for multi-irb, please leave a comment at https://github.com/ruby/irb/issues/653
+        MSG
+      end
+
       def extend_irb_context
         # this extension patches IRB context like IRB.CurrentContext
         require_relative "../ext/multi-irb"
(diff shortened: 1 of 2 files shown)


lib/irb/cmd/subirb.rb
```diff
@@ -18,45 +18,56 @@ def initialize(conf)
 
       private
 
+      def print_deprecated_warning
+        warn <<~MSG
+          Multi-irb commands are deprecated and will be removed in IRB 2.0.0.
+          If you have any use case for multi-irb, please leave a comment at https://github.com/ruby/irb/issues/653
```
tompng, line 24: LGTM :+1:
I think it will be more helpful if there is a navigation to workspace commands

st0012: Do you mean something like:

```
If you use the `irb` command to create a new workspace, you can use `chws` instead.
```

tompng: Yes. or like
```
Multi-irb commands are deprecated and will be removed in IRB 2.0.0. Please use workspace commands instead.
```

PR #657, opened 2023-07-25
Page `ls`'s output

st0012:
This PR:

- Introduces `Pager#page_content` that decides whether to page the content based on if it'd take too much of user's screen space.
    - For example, `"a"` will not be paged. But `"a" * 1000000` probably will be.
    - This matches Pry's paging behaviour.
- Use `Pager#page_content` in `ls` and `show_cmds` commands.

## Demo

https://github.com/ruby/irb/assets/5079556/074c152a-b0e6-40e2-ae00-1ca1699fd807

### DIFF
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -7,6 +7,16 @@ module IRB
     PAGE_COMMANDS = [ENV['RI_PAGER'], ENV['PAGER'], 'less', 'more'].compact.uniq
 
     class << self
+      def page_content(content)
+        if content_exceeds_screen_height?(content)
+          page do |io|
+            io.puts content
+          end
+        else
+          $stdout.puts content
+        end
+      end
+
       def page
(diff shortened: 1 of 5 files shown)


lib/irb/pager.rb
```diff
@@ -30,6 +40,21 @@ def page
 
       private
 
+      def content_exceeds_screen_height?(content)
+        screen_height, screen_width = begin
+          Reline.get_screen_size
+        rescue Errno::EINVAL
```
tompng, line 46: Maybe we should also fix Reline to rescue EINVAL? Reline only rescue ENOTTY.

I found a comment for EINVAL reproducible environment here
https://github.com/ruby/ruby/blob/26aef1c73639718bc00c271ce7176bf3ee565c12/test/io/console/test_io_console.rb#L310
and here
https://bugs.ruby-lang.org/issues/10494

tompng (approved):

PR #658, opened 2023-07-27
Decouple `edit` and `show_source` commands

st0012:
Ideally all commands should be independent from each other. But currently `edit` is implemented based on the `show_source` command for its source-finding mechanism.

I think the better approach is to extract the source-finding part to a standalone utility class and use it in both commands instead.


tompng (approved):
:+1:
The changed part in `find_source` `find_end` (local var name, when to check File.exist?) also looks good :+1:

Issue #659, opened 2023-07-31
History is not saved with nested IRB sessions


st0012:
```rb
# test.rb
class Foo
  def self.bar
    binding.irb
  end
end
binding.irb
```

```
$ cat ~/.irb_history
cat: /Users/hung-wulo/.irb_history: No such file or directory
$ be ruby test.rb

From: test.rb @ line 7 :

    2: class Foo
    3:   def self.bar
    4:     binding.irb
    5:   end
    6: end
 => 7: binding.irb

irb(main):001:0> 111
=> 111
irb(main):002:0> Foo.bar

From: test.rb @ line 4 :

    1: # test.rb
    2: class Foo
    3:   def self.bar
 => 4:     binding.irb
    5:   end
    6: end
    7: binding.irb

irb(Foo):001:0> 222
=> 222
irb(Foo):002:0> exit
=> nil
irb(main):003:0> 333
=> 333
irb(main):004:0> exit
$ cat ~/.irb_history
cat: /Users/hung-wulo/.irb_history: No such file or directory
```

PR #660, opened 2023-07-31
Extract integration testing helpers out of debug command tests

st0012:
The ability to run a test case in a subprocess is useful for testing many other features, like nested IRB sessions. So I think it's worth extracting them into a new test case class.

For example, I'll need it for #659

### DIFF
--- a/test/irb/helper.rb
+++ b/test/irb/helper.rb
@@ -7,6 +7,11 @@ begin
 rescue LoadError # ruby/ruby defines helpers differently
 end
 
+begin
+  require "pty"
+rescue LoadError # some platforms don't support PTY
+end
+
 module IRB
   class InputMethod; end
 end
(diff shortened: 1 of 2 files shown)


test/irb/helper.rb
```diff
@@ -73,4 +78,109 @@ def without_rdoc(&block)
...
+
+    def type(command)
+      @commands << command
+    end
+
+    def write_ruby(program)
+      @ruby_file = Tempfile.create(%w{irb- .rb})
+      @ruby_file.write(program)
+      @ruby_file.close
+    end
```
tompng, line 180: Just an idea

I know that this code is not changed from `test_debug_cmd.rb`, but could it be like this?
```ruby
ruby_code = <<~RUBY
  foo
  binding.irb
  bar
RUBY
run_ruby_file ruby_code do end
```

This helper will going to be used in several test, so it's a good opportunity to change it.

`write_ruby` was introduced in https://github.com/ruby/irb/commit/976100c1c26808b9d8c58f72442dbc01c13a790f.
In that point, defining `write_ruby` like another existing method `write_irbrc` makes sense because it's optional. Not always used.
Now, write_ruby is always used before run_ruby_file. I think passing as an argument is more simple.


Although, my idea has a weak point. Must think how to do something like
```
assert_match(/Object#foo at #{@ruby_file.to_path}/, output)
```

test/irb/helper.rb
```diff
@@ -73,4 +78,109 @@ def without_rdoc(&block)
...
+      #{'=' * 27} END OF OUTPUT #{'=' * 27}
+      MSG
+      assert_block(message) { false }
+    ensure
+      File.unlink(@ruby_file) if @ruby_file
+      FileUtils.remove_entry tmp_dir
+    end
+
+    # read.gets could raise exceptions on some platforms
+    # https://github.com/ruby/ruby/blob/master/ext/pty/pty.c#L729-L736
```
tompng, line 136: Line number is changed now. https://github.com/ruby/ruby/blob/master/ext/pty/pty.c#L721-L728

test/irb/helper.rb
st0012: I agree with the direction, but I feel we can design the API better after seeing other use cases. Can we revisit this after we applied integration for #659's fix?

test/irb/helper.rb
st0012: Updated 👍

tompng (approved):

PR #661, opened 2023-08-01
Add workspace category

st0012:


st0012:
I think we can later refactor their implementation too 🤔

tompng (approved):

PR #669, opened 2023-08-05
Don't echo an expression's result when it ends with a semicolon

### LINKED ISSUE #656: Skip inspect when line ends with a semicolon `;`
## Description

One small feature from Pry that I'm missing in IRB is the ability to skip inspecting the return value by adding a semicolon at the end of a line:

```
$ pry
[1] pry(main)> 42
=> 42
[2] pry(main)> 42;
[3] pry(main)> exit
```

vs

```sh-session
$ irb
irb(main):001:0> 42
=> 42
irb(main):002:0> 42;
=> 42
irb(main):003:0> exit
```

## Workaround

I can just append `; nil` to skip inspect for my value and inspect `nil` instead.

## Result of irb_info

```
irb(main):001:0> irb_info
Ruby version: 3.2.2
IRB version: irb 1.7.4 (2023-07-15)
RUBY_PLATFORM: arm64-darwin22
LANG env: fr_FR.UTF-8
LC_ALL env: en_US.UTF-8
East Asian Ambiguous Width: 1
```

## Terminal Emulator

Terminal.app

## Setting Files

Yes, but I removed them just to make sure it wasn't impacting this issue.

st0012:
Closes #656

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -562,7 +562,8 @@ module IRB
             is_assignment = assignment_expression?(line)
             evaluate_line(line, line_no)
 
-            if @context.echo?
+            # Don't echo if the line ends with a semicolon
+            if @context.echo? && !line.match?(/;\s*\z/)
               if is_assignment
                 if @context.echo_on_assignment?
                   output_value(@context.echo_on_assignment? == :truncate)
--- a/test/irb/test_evaluation.rb
+++ b/test/irb/test_evaluation.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require "tempfile"
+
+require_relative "helper"
+
+module TestIRB
+  class EchoingTest < IntegrationTestCase
+    def test_irb_echos_by_default
+      write_ruby <<~'RUBY'
+        binding.irb
+      RUBY
+
+      output = run_ruby_file do
+        type "123123"
+        type "exit"
+      end
+
+      assert_include(output, "=> 123123")
+    end
+
+    def test_irb_doesnt_echo_line_with_semicolon
+      write_ruby <<~'RUBY'
+        binding.irb
+      RUBY
+
+      output = run_ruby_file do
+        type "123123;"
+        type "123123   ;"
+        type "123123;   "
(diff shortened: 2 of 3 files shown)


test/irb/test_evaluation.rb
```diff
@@ -0,0 +1,44 @@
...
+
+require "tempfile"
+
+require_relative "helper"
+
+module TestIRB
+  class EchoingTest < IntegrationTestCase
+    def test_irb_echos_by_default
+      write_ruby <<~'RUBY'
+        binding.irb
```
st0012, line 11: For some reason, opening IRB with `IRB.start` never works in integration tests.

```rb
# this hangs and prints nothing
write_ruby <<~'RUBY'
  require "irb"
  IRB.start(__FILE__)
RUBY
```

Similarly, introducing a helper like `run_irb` to make it run `exe/irb` doesn't work either.

So I decided to go with `binding.irb` for now and investigate this further later.

lib/irb.rb
```diff
@@ -562,7 +562,8 @@ def eval_input
             is_assignment = assignment_expression?(line)
             evaluate_line(line, line_no)
 
-            if @context.echo?
+            # Don't echo if the line ends with a semicolon
+            if @context.echo? && !line.match?(/;\s*\z/)
```
tompng, line 566: I think this simple implementation is good enough.

However, I found a false positive. Using`tokens.last` will be more strict. What do you think?
```ruby
puts %;string;
```
```ruby
puts +% ; 
```
```ruby
puts <<';'
;
```
```ruby
puts <<' '
heredoc;
 
```

And maybe a false negative, maybe not. (Checking this will be very hard)
```ruby
puts(<<A);
A
```

st0012: Great catch! And yeah I agree it's better to make this check with `Ripper`. But I'd like to avoid invoking it outside of `RubyLex` and want to make this property an yield argument of `each_top_level_statement` instead (like #670). WDYT?

mykhal:
Nice. Thanks.

Maybe it should be also mentioned in docs somewhere.

PR #670, opened 2023-08-06
Move assignment check to RubyLex

st0012:
Since assignment check relies on tokenization with `Ripper`, it feels like the responsibility of `RubyLex`. `Irb#eval_input` should simply get the result when calling `each_top_level_statement` on `RubyLex`.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -429,30 +429,6 @@ module IRB
   end
 
   class Irb
-    ASSIGNMENT_NODE_TYPES = [
-      # Local, instance, global, class, constant, instance, and index assignment:
-      #   "foo = bar",
-      #   "@foo = bar",
-      #   "$foo = bar",
-      #   "@@foo = bar",
-      #   "::Foo = bar",
-      #   "a::Foo = bar",
-      #   "Foo = bar"
-      #   "foo.bar = 1"
-      #   "foo[1] = bar"
-      :assign,
-
-      # Operation assignment:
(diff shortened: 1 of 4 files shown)


lib/irb.rb
```diff
@@ -555,11 +531,9 @@ def eval_input
 
       @scanner.configure_io(@context.io)
 
-      @scanner.each_top_level_statement do |line, line_no|
+      @scanner.each_top_level_statement do |line, line_no, is_assignment|
```
tompng, line 534: There is a plan to add end_with_semicolon_check and number of parameter will increase. Then it will look like this.
```ruby
@scanner.each_top_level_statement do |line, line_no, is_a, is_b, is_c|

end
```

How about this?
```ruby
@scanner.each_top_level_statement do |line, line_no| # line_no and line is always needed.
  # IMO optional information of `line` should be explicitly calculated here
  is_a = @scanner.check_a?(line)
  is_c = @scanner.check_c?(line)
  is_d = @scanner.check_d?(line)
end
```

st0012: For the assignment check, I think it's better to be handled from `each_top_level_statement` so it's always calculated before line evaluation. I'm fine using the suggested pattern for semicolon check though.

I was also thinking instead of returning individual information as the yielding argument, perhaps we can encapsulate them inside a `Statement` object? But I will explore it after this and #669 is merged.

tompng (approved):

PR #671, opened 2023-08-07
Make Reline/ReadlineInputMethod inherit StdioInputMethod

st0012:
They are both built on top of stdio and are basically extended version of `StdioInputMethod`. They also share several attributes and methods with `StdioInputMethod`.


tompng (approved):
:+1:

Issue #665, opened 2023-08-03
Public API to calculate indent


DmitryPogrebnoy:
## Description

Hi! IRB in multistring expressions adds appropriate indents to the next line. However, there is no API to get the indent value for the next line.

As far as I know, right now you can only get the indentation in some hacky way via an internal API. The internal API changes frequently and so the stability of such a hacky implementation suffers. It would be great to have a public API for this. Thank you!

st0012:
Would you mind sharing your use case of it? That'll help us decide how the API could look like to better fit your needs.

DmitryPogrebnoy:
This API is mainly needed to be used in the `ruby-denbug-ide` gem. [Here is an example of how it is implemented now](https://github.com/ruby-debug/ruby-debug-ide/blob/master/lib/ruby-debug-ide/commands/expression_info.rb#L11C9-L11C16). This method evaluates three values for an input expression in Ruby. The method calculates whether the expression is complete, calculates the indentation size for the next line of the expression, and finds the last prompt.

Currently, the indentation calculation uses an internal API that changes from version to version. It is inconvenient to maintain it. That's why there is a need for a public API.

tompng:
For now, here are some idea how to calculate indentation.

### Use a formatter or a language server
I think Indent calculation is similar to code formatting. Maybe there's an API in formatter or language server.

### Use a parser/tokenizer library
One way to calculate indent is to use error tolerant parser like YARP or RubyVM::AbstractSyntaxTree(ruby>=3.3).
Another way is to tokenize the code and scan unclosed tokens by yourself. You can use Ripper(IRB uses it), CodeRay(Pry uses it), YARP, Parser.
Here's an example https://gist.github.com/tompng/959d374826153dd3bef59232c456773a

### Fork and make an indent calculation library
IRB's current indent calculation is [here](
https://github.com/ruby/irb/blob/3847532e54743888b432f7a4cd06578da0bd7b27/lib/irb/ruby-lex.rb#L399-L422).
It uses `IRB::NestingParser` which is ported from https://github.com/tompng/katakata_irb/blob/ea461ae8ffa88a9a968365cfff9b7aaa51426cc3/lib/katakata_irb/nesting_parser.rb
I created a demo gem that provides indent calculation
https://github.com/tompng/indent_calculator_demo

PR #673, opened 2023-08-07
Add a new readme section for IRB extension

st0012:
Since we have concerns on breaking private API usages during refactoring (like #635) and we also received requests like #665, I think it's worth acknowledging that we're aware of the risk as well as explain the direction we work toward to.

In addition to the new section, I think the readme now also has enough content that could use a ToC. (I'm aware of the GH generated one, but I think it's usually overlooked so I don't want to count on it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,23 @@ IRB stands for "interactive Ruby" and is a tool to interactively execute Ruby ex
 
 The `irb` command from your shell will start the interpreter.
 
+- [Installation](#installation)
+- [Usage](#usage)
+  - [The `irb` Executable](#the-irb-executable)
+  - [The `binding.irb` Breakpoint](#the-bindingirb-breakpoint)
+- [Commands](#commands)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,21 @@ IRB stands for "interactive Ruby" and is a tool to interactively execute Ruby ex
 
 The `irb` command from your shell will start the interpreter.
 
+- [Installation](#installation)
+- [Usage](#usage)
+  - [The `irb` Executable](#the-irb-executable)
+  - [The `binding.irb` Breakpoint](#the-bindingirb-breakpoint)
+- [Commands](#commands)
```
tompng, line 14: Looks good.
Adding navigation for `Debugging with IRB` section added in #689 will be better 😄

st0012: Yeah I've rebased and updated it 👍

tompng (approved):

PR #679, opened 2023-08-12
Drop showing indent level number in DEFAULT prompt and INF_RUBY prompt

tompng:
Removes indent level number in `:DEFAULT` and `:INF_RUBY` prompt mode.

```
irb(main):001:1> if true
irb(main):002:2>   puts(
irb(main):003:3>     1,
irb(main):004:3>     2,
↓
irb(main):001> if true
irb(main):002>   puts(
irb(main):003>     1,
irb(main):004>     2,
``````

Until irb 1.0.0, auto indent was not enabled by default and this indent level number was useful.
```
irb(main):001:0> IRB::VERSION
=> "1.0.0"
irb(main):002:0> def f
irb(main):003:1> if true
irb(main):004:2> puts(
irb(main):005:3* 1,
irb(main):006:3* 2,
irb(main):007:3* )
irb(main):008:2> end
irb(main):009:1> end
=> :f
```
But now, auto indent is enabled by default. Indent level is identical to next-line's indentation spaces. You can easily know how deep the code is nested without seeing this number. It is just making prompt long.

### Other prompt modes
For `:INF_RUBY` prompt mode, only `PROMPT_I` is present. In this case, indent level number (always zero) is completely useless.
Indent level number still remains in `:CLASSIC` prompt mode.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -194,10 +194,10 @@ require_relative "irb/easter-egg"
 # For instance, the default prompt mode is defined as follows:
 #
 #     IRB.conf[:PROMPT_MODE][:DEFAULT] = {
-#       :PROMPT_I => "%N(%m):%03n:%i> ",
-#       :PROMPT_N => "%N(%m):%03n:%i> ",
-#       :PROMPT_S => "%N(%m):%03n:%i%l ",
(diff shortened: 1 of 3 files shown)


st0012 (approved):
Love it 😍

PR #684, opened 2023-08-17
Clear all context usages in RubyLex

st0012:
After this change, `RubyLex` will not interact with `Context` directly in any way. This decoupling has a few benefits:

- It makes `RubyLex` easier to test as it no longer has a dependency on `Context`. We can see this from the removal of `build_context` from `test_ruby_lex.rb`.
- It will make `RubyLex` easier to understand as it will not be affected by state changes in `Context` objects.
- It allows `RubyLex` to be used in places where `Context` is not available.

(This is based on #683 and needs it to be merged first)

### DIFF
--- a/lib/irb/source_finder.rb
+++ b/lib/irb/source_finder.rb
@@ -43,7 +43,7 @@ module IRB
     private
 
     def find_end(file, first_line)
-      lex = RubyLex.new(@irb_context)
+      lex = RubyLex.new
       lines = File.read(file).lines[(first_line - 1)..-1]
       tokens = RubyLex.ripper_lex_without_warning(lines.join)
       prev_tokens = []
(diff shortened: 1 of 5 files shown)


lib/irb/source_finder.rb
```diff
@@ -43,17 +43,18 @@ def find_source(signature)
     private
 
     def find_end(file, first_line)
-      lex = RubyLex.new(@irb_context)
+      local_variables = @irb_context.local_variables
```
tompng, line 46: I think local_variables should be `[]` here.
Parsing a file does not need to depend on binding's local variable.


```ruby
# a.rb
a /%#/ if false; def foo;end #.size; def foo;"

binding.irb
__END__
"
end
```
```
$ ruby -Ilib a.rb
irb(main):001> $ foo
# shows source of foo
irb(main):002> a = 1
=> 1
irb(main):003> $ foo
# shows wrong source range of foo
```

st0012: Good point. Updated 👍

tompng (approved):

PR #685, opened 2023-08-20
Remove unused `PROMPT_N`

smmr0:
Since #500 ([revision `364a6d56`](https://bugs.ruby-lang.org/projects/ruby-master/repository/git/revisions/364a6d56d776270da09604816d623047c66c5e32)), the `PROMPT_N` prompt format is no longer used (explanation follows). This change removes it.

---

The only usage of `PROMPT_N`'s value is in `IRB::Irb#eval_input`:

<https://github.com/ruby/irb/blob/65e8e6869075c47c95591f1d12175eb6377b8a9f/lib/irb.rb#L522-L534>

In order for `@context.prompt_n` to be used here, `continue` must be falsey and `indent` must be positive.

That block given to `set_prompt` is called only by`RubyLex#prompt`:

<https://github.com/ruby/irb/blob/65e8e6869075c47c95591f1d12175eb6377b8a9f/lib/irb/ruby-lex.rb#L157-L161>

In order to satisfy the conditions above, `opens` must be empty (technically [it could also be populated with only falsey values](https://thoughtbot.com/blog/any-empty), but that's not the case here) and `calc_indent_level(opens)` must be positive.

`RubyLex#calc_indent_level` is defined as:

<https://github.com/ruby/irb/blob/65e8e6869075c47c95591f1d12175eb6377b8a9f/lib/irb/ruby-lex.rb#L365-L388>

If `opens` is empty, then the block given to `opens.each_with_index` will never be evaluated, and `calc_indent_level` will return the initial value of `indent_level`: `0`. Therefore `calc_indent_level(opens)` cannot be positive while `opens` is empty.

Since these conditions cannot be satisfied, `PROMPT_N` will never be used.

---

However, maybe it's unintended that `PROMPT_N` is unused now, in which case I can close this PR.

### DIFF
--- a/doc/irb/irb.rd.ja
+++ b/doc/irb/irb.rd.ja
@@ -125,7 +125,6 @@ irb起動時に``~/.irbrc''を読み込みます. もし存在しない場合は
 
    IRB.conf[:PROMPT][:MY_PROMPT] = { # プロンプトモードの名前
      :PROMPT_I => nil,		  # 通常のプロンプト
-     :PROMPT_N => nil,		  # 継続行のプロンプト
      :PROMPT_S => nil,		  # 文字列などの継続行のプロンプト
      :PROMPT_C => nil,		  # 式が継続している時のプロンプト
      :RETURN => "    ==>%s\n"	  # リターン時のプロンプト
@@ -140,7 +139,7 @@ OKです.
 
   IRB.conf[:PROMPT_MODE] = :MY_PROMPT
 
-PROMPT_I, PROMPT_N, PROMPT_S, PROMPT_Cは, フォーマットを指定します.
+PROMPT_I, PROMPT_S, PROMPT_Cは, フォーマットを指定します.
 
   %N	起動しているコマンド名が出力される.
   %m	mainオブジェクト(self)がto_sで出力される.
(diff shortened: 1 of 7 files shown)


tompng:
Thanks for the investigation. The change looks good :+1:

I've checked git log to find what PROMPT_N was doing before.
PROMPT_N and PROMPT_C is almost combined to PROMPT_C in IRB 1.1.0.
IRB 1.0.0
```
prompt_i> if 1
prompt_n> 2.
prompt_c> to_s
prompt_n> end
=> "2"
```

IRB 1.1.0, the first version of IRB integrated with Reline.
```
prompt_i> i█
↓
prompt_c> if 1
prompt_c> 2.
prompt_c> to_s
prompt_c> en█
↓
prompt_c> if 1
prompt_c> 2.
prompt_c> to_s
prompt_i> end
=> "2"
```

Since IRB 1.1.0 and before #500, PROMPT_N seems to be used only in a weird condition `(indent > 0) and (continue || code_block_open) == false`
https://github.com/ruby/irb/blob/359cb28defcfa416b0633e9c0b6a494e834fa41f/lib/irb/ruby-lex.rb#L99
`code_block_open == false` and `indent > 0` conflicts. I think it was a bug that this condition is accidentally satisfied in some input before #500.
```
# example of prompt_n accidentally used
prompt_n> ((end
```

So I think PROMPT_N should be removed in IRB 1.1.0. Thank you for finding this :smile:

tompng (approved):

PR #686, opened 2023-08-20
Support `VISUAL` env var, and prefer it over `EDITOR`

smmr0:
The [`edit` command](https://github.com/ruby/irb/blob/66e69fa0dc91b38aedf17f897c70f4da6cca3052/lib/irb/cmd/edit.rb) currently only uses the `EDITOR` environment variable to know which command to run to open that editor. It should also use the `VISUAL` environment variable, and it should prefer it over `EDITOR`.

See [this StackExchange answer](https://unix.stackexchange.com/a/4861) for some context:

> The `EDITOR` editor should be able to work without use of "advanced" terminal functionality (like old `ed` or `ex` mode of `vi`). It was used on teletype terminals.
> 
> A `VISUAL` editor could be a full screen editor as `vi` or `emacs`.
> 
> E.g. if you invoke an editor through bash (using `C-x C-e`), bash will try first `VISUAL` editor and then, if `VISUAL` fails (because terminal does not support a full-screen editor), it tries `EDITOR`.
> 
> Nowadays, you can leave `EDITOR` unset or set it to `vi -e`.

This pull request changes the `edit` command to first look for a value for `VISUAL`, then if `VISUAL` is unset, look for a value for `EDITOR`.

---

I've previously made [this same change in Rails](https://github.com/rails/rails/pull/48374).

[Pry also uses `VISUAL`](https://github.com/pry/pry/blob/52799c4902cca764f052ae7ded5361bc50feaf57/lib/pry/editor.rb#L8-L14) (<https://github.com/pry/pry/pull/211>).

### DIFF
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -914,12 +914,15 @@ module TestIRB
 
   class EditTest < CommandTestCase
     def setup
+      @original_visual = ENV["VISUAL"]
       @original_editor = ENV["EDITOR"]
       # noop the command so nothing gets executed
-      ENV["EDITOR"] = ": code"
+      ENV["VISUAL"] = ": code"
+      ENV["EDITOR"] = ": code2"
     end
 
     def teardown
+      ENV["VISUAL"] = @original_visual
       ENV["EDITOR"] = @original_editor
     end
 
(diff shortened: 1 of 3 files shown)


test/irb/test_cmd.rb
```diff
@@ -982,5 +985,20 @@ def test_edit_with_instance_method
...
+      out, err = execute_lines(
+        "edit",
+        irb_path: __FILE__
+      )
+
+      assert_empty err
+      assert_match("path: #{__FILE__}", out)
+      assert_match("command: ': code2'", out)
+
+      ENV["VISUAL"] = original_visual
```
st0012, line 1001: Nit: I think recovering `VISUAL` here is not necessary as it'd be done by the teardown callback anyway.

```suggestion
      ENV.delete("VISUAL")

      out, err = execute_lines(
        "edit",
        irb_path: __FILE__
      )

      assert_empty err
      assert_match("path: #{__FILE__}", out)
      assert_match("command: ': code2'", out)
```

st0012 (approved):
Thanks 👍

PR #692, opened 2023-08-23
Deprecate RubyLex and warn about referencing to it

st0012:
`RubyLex` has always been a private component of IRB, so we should explicitly discourage usages of it.
Also, it should be placed under the `IRB` module like other components.

(It'd be easier to review by ignoring whitespaces with [this link](https://github.com/ruby/irb/pull/692/files?diff=split&w=1))

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -8,489 +8,494 @@ require "ripper"
 require "jruby" if RUBY_ENGINE == "jruby"
 require_relative "nesting_parser"
 
-# :stopdoc:
-class RubyLex
-  ASSIGNMENT_NODE_TYPES = [
-    # Local, instance, global, class, constant, instance, and index assignment:
-    #   "foo = bar",
-    #   "@foo = bar",
-    #   "$foo = bar",
-    #   "@@foo = bar",
-    #   "::Foo = bar",
-    #   "a::Foo = bar",
-    #   "Foo = bar"
-    #   "foo.bar = 1"
-    #   "foo[1] = bar"
-    :assign,
+module IRB
+  # :stopdoc:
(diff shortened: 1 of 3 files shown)


tompng:
`test_nesting_parser.rb` uses `RubyLex.ripper_lex_without_warning` and it needs to be `IRB::RubyLex.ripper_lex_without_warning`

lib/irb/ruby-lex.rb
```diff
@@ -8,489 +8,500 @@
...
-    false
+  end
+  # :startdoc:
+end
+
+class RubyLex < IRB::RubyLex
+  def initialize(_)
+    warn <<~MSG
+      RubyLex is a private component of IRB and referencing to it from outside of IRB is deprecated.
+    MSG
```
tompng, line 504: What do you think about deprecate_constant?
```ruby
RubyLex = IRB::RubyLex
Object.deprecate_constant :RubyLex
```
It only warns with `ruby -w`. 
It can also warn use of `RubyLex::Const` and `RubyLex.class_method` but can't customize deprecation message.

st0012: Hmm I did try that but when I referenced it an exception was raised. But I guess it's probably something wrong with my testing script. Have updated it now 👍

tompng (approved):

k0kubun:
The test has been failing on this CI: http://ci.rvm.jp/results/trunk-repeat50@ruby-sp2-docker/4695983

peterzhu2118:
@k0kubun I have a fix in #695

st0012:
@k0kubun sorry that I didn't notice it. I've merged @peterzhu2118's fix.

k0kubun:
Thanks both!

PR #693, opened 2023-08-23
Improve help/show_cmds message during debugger integration

st0012:
1. When in the `irb:rdbg` session, IRB should pass `help` command to the debugger, instead of invoking ri.
2. `show_cmds` should update its content when in the `irb:rdbg` session:
    - It should remove the old `Debugging` session as they're not useful in that context.
    - It should also remove `help` due to the change of 1)
    - Finally, I think it's better to append `debug`'s help message at the end of its output, so users don't need to learn when to use `show_cmds` and when to use `help`.

### DIFF
--- a/lib/irb/statement.rb
+++ b/lib/irb/statement.rb
@@ -60,7 +60,9 @@ module IRB
       end
 
       def should_be_handled_by_debugger?
-        IRB::ExtendCommand::DebugCommand > @command_class
+        require_relative 'cmd/help'
+        require_relative 'cmd/debug'
+        IRB::ExtendCommand::DebugCommand > @command_class || IRB::ExtendCommand::Help == @command_class
       end
 
       def evaluable_code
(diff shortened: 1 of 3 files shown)


lib/irb/statement.rb
```diff
@@ -60,7 +60,8 @@ def suppresses_echo?
       end
 
       def should_be_handled_by_debugger?
-        IRB::ExtendCommand::DebugCommand > @command_class
+        require_relative 'cmd/help'
```
tompng, line 63: I think we need `require_relative 'cmd/debug'` here too.
This method is used as
```ruby
if @context.with_debugger && statement.should_be_handled_by_debugger?
```
and it depends on calling order. The code below does not work now.
```ruby
if statement.should_be_handled_by_debugger? && @context.with_debugger
```

tompng (approved):

PR #701, opened 2023-08-30
Decouple RubyLex from prompt and line_no

tompng:
Removes the last RubyLex's instance variable `@prompt` and `@line_no`.

Reason:
`@line_no` is completely controlled by IRB::Irb with `@scanner.increase_line_no`. It should be IRB::Irb's instance variable.
`@scanner.set_prompt do end` and `@scanner.prompt(args)` seems roundabout.
I wanted to remove `save_prompt_to_context_io` because the flow is complex and hard to understand.

After deleting `@prompt` and `@line_no` from RubyLex, RubyLex will have no state at all. RubyLex does not have to be a class anymore.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -437,6 +437,7 @@ module IRB
       @context.workspace.load_commands_to_main
       @signal_status = :IN_IRB
       @scanner = RubyLex.new
+      @line_no = 1
     end
 
     # A hook point for `debug` command's breakpoint after :IRB_EXIT as well as its clean-up
@@ -454,7 +455,7 @@ module IRB
       workspace = IRB::WorkSpace.new(binding)
       context.workspace = workspace
       context.workspace.load_commands_to_main
-      scanner.increase_line_no(1)
+      @line_no += 1
 
       # When users run:
       # 1. Debugging commands, like `step 2`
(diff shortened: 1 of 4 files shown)


lib/irb.rb
```diff
@@ -513,34 +514,38 @@ def run(conf = IRB.conf)
...
-          @context.io.prompt = p = ""
-        end
-        if @context.auto_indent_mode and !@context.io.respond_to?(:auto_indent)
-          unless ltype
-            prompt_i = @context.prompt_i.nil? ? "" : @context.prompt_i
-            ind = prompt(prompt_i, ltype, indent, line_no)[/.*\z/].size +
-              indent * 2 - p.size
-            @context.io.prompt = p + " " * ind if ind > 0
-          end
+    def prompt(opens, continue, line_offset)
```
st0012, line 517: Can we call it `generate_prompt` or `generate_prompt_string` to make its action more clear?

st0012 (commented):
Currently we don't declare private methods under `IRB::Irb`, but I think these 2 prompt methods can be the first 2. WDYT?
While we expect `Irb` to be a private component, it's still relatively easy to retrieve (e.g. with `irb_context.irb`) and could be used by us in other gems too (e.g. https://github.com/ruby/debug/pull/1024). So making some methods private could ideally make people think twice before calling them.

st0012 (approved):

PR #704, opened 2023-08-31
Drop rdoc's version requirement

### LINKED ISSUE #703: Minor Version of IRB introducing undisclosed psych breaking changes
## Description

For any of us using Ruby 3.0 or 2.7.8, both of which have psych 3 as a default gem, and having a working environment using irb 1.7.1, upgrading to irb 1.8.0 [https://github.com/ruby/irb/releases/tag/v1.8.0](https://github.com/ruby/irb/releases/tag/v1.8.0)  causes an undocumented breaking change.

This is because irb 1.8.0 now depends on rdoc because of [https://github.com/ruby/irb/pull/648](https://github.com/ruby/irb/pull/648) .

Rdoc has a dependency on psych 4 (see [https://github.com/ruby/rdoc/blob/e3bd8698df51b413b0b203247edc160b876ac79a/rdoc.gemspec#L236](https://github.com/ruby/rdoc/blob/e3bd8698df51b413b0b203247edc160b876ac79a/rdoc.gemspec#L236) , which has significant breaking changes from the Ruby defaults of psych 3.

For more on the breaking changes, see the following:
* [https://github.com/ruby/psych/discussions/571](https://github.com/ruby/psych/discussions/571) 
* [https://bugs.ruby-lang.org/issues/17866](https://bugs.ruby-lang.org/issues/17866) 

Amongst other articles (Rails and Ruby have both had to make changes because of the safe_load change)

We are aware that psych 4 will be/is the default in Ruby 3.1 and Rails 7.1, but for those of us not there, this is an unexpected breaking change caused by a minor update to irb. 

While I would love the rdoc dependency removed in lieu of Ruby defaults or per-application choice, if nothing else, the transitive upgrade to psych should be disclosed or documented as a breaking change in my opinion, with a major version bump for irb.

st0012:
1. The newer versions of rdoc requires pysch 4.0+, which could break apps using Ruby 3.0 or 2.7. #703 has more detailed explanation on this.
2. We actually don't use any version-specific rdoc APIs. So having a version requirement is not necessary atm.

Closes #703

### DIFF
--- a/irb.gemspec
+++ b/irb.gemspec
@@ -42,5 +42,5 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = Gem::Requirement.new(">= 2.7")
 
   spec.add_dependency "reline", ">= 0.3.8"
-  spec.add_dependency "rdoc", "~> 6.5"
+  spec.add_dependency "rdoc"
 end


tompng (approved):

tnir:
Thanks! 💞

PR #708, opened 2023-09-08
Type based completion using Prism and RBS

tompng:
Implement type based autocompletion #734 using Prism and RBS

Default completor is still `:regexp`. To enable `:type` completor, write `IRB.conf[:COMPLETOR] = :type` to irbrc.

When Prism is available
without_rbs

When Prism and RBS is available
with_rbs

When Prism is not available, it fallbacks to RegexpCompletor

This pull request will port gem katakata_irb's completion(https://github.com/tompng/katakata_irb/) to IRB.

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -230,13 +230,13 @@ module IRB
     HISTORY = Reline::HISTORY
     include HistorySavingAbility
     # Creates a new input method object using Reline
-    def initialize
+    def initialize(completor)
       IRB.__send__(:set_encoding, Reline.encoding_system_needs.name, override: false)
 
-      super
+      super()
 
       @eof = false
-      @completor = RegexpCompletor.new
+      @completor = completor
 
       Reline.basic_word_break_characters = BASIC_WORD_BREAK_CHARACTERS
       Reline.completion_append_character = nil
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -275,7 +275,7 @@ module IRB
     end
 
     def completion_info
-      autocomplete_message = IRB.conf[:USE_AUTOCOMPLETE] ? 'Autocomplete' : 'Tab Complete'
+      autocomplete_message = Reline.autocompletion ? 'Autocomplete' : 'Tab Complete'
       "#{autocomplete_message}, #{@completor.inspect}"
     end
 
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -193,6 +193,10 @@ module IRB
       }
     end
 
+    def completion_info
+      'RegexpCompletor'
+    end
+
     # Reads the next line from this input method.
     #
     # See IO#gets for more information.
@@ -268,6 +272,11 @@ module IRB
         rescue LoadError
         end
       end
+    end
+
+    def completion_info
+      autocomplete_message = IRB.conf[:USE_AUTOCOMPLETE] ? 'Autocomplete' : 'Tab Complete'
+      "#{autocomplete_message}, #{@completor.inspect}"
     end
 
     def check_termination(&block)
--- a/lib/irb/type_completion/completor.rb
+++ b/lib/irb/type_completion/completor.rb
@@ -1,0 +1,230 @@
+require 'prism'
+require 'irb/completion'
+require_relative 'type_analyzer'
+module IRB
+  module TypeCompletion
+    class Completor < BaseCompletor # :nodoc:
+      HIDDEN_METHODS = %w[Namespace TypeName] # defined by rbs, should be hidden
+
+      def inspect
+        name = 'TypeCompletion::Completor'
+        prism_info = "Prism: #{Prism::VERSION}"
+        if Types.rbs_builder
+          "#{name}(#{prism_info}, RBS: #{RBS::VERSION})"
+        elsif Types.rbs_load_error
+          "#{name}(#{prism_info}, RBS: #{Types.rbs_load_error.inspect})"
+        else
+          "#{name}(#{prism_info}, RBS: loading)"
+        end
+      end
+
+      def completion_candidates(preposing, target, _postposing, bind:)
+        @preposing = preposing
+        verbose, $VERBOSE = $VERBOSE, nil
+        code = "#{preposing}#{target}"
+        @result = analyze code, bind
+        name, candidates = candidates_from_result(@result)
+
+        all_symbols_pattern = /\A[ -\/:-@\[-`\{-~]*\z/
+        candidates.map(&:to_s).select { !_1.match?(all_symbols_pattern) && _1.start_with?(name) }.uniq.sort.map do
+          target + _1[name.size..]
+        end
+      rescue SyntaxError, StandardError => e
+        handle_error(e)
+        []
+      ensure
+        $VERBOSE = verbose
+      end
+
+      def doc_namespace(preposing, matched, postposing, bind:)
+        name = matched[/[a-zA-Z_0-9]*[!?=]?\z/]
+        method_doc = -> type do
+          type = type.types.find { _1.all_methods.include? name.to_sym }
+          if type.is_a? Types::SingletonType
+            "#{Types.class_name_of(type.module_or_class)}.#{name}"
+          elsif type.is_a? Types::InstanceType
+            "#{Types.class_name_of(type.klass)}##{name}"
+          end
+        end
+        call_or_const_doc = -> type do
+          if name =~ /\A[A-Z]/
+            type = type.types.grep(Types::SingletonType).find { _1.module_or_class.const_defined?(name) }
+            type.module_or_class == Object ? name : "#{Types.class_name_of(type.module_or_class)}::#{name}" if type
+          else
+            method_doc.call(type)
+          end
+        end
+
+        value_doc = -> type do
+          return unless type
+          type.types.each do |t|
+            case t
+            when Types::SingletonType
+              return Types.class_name_of(t.module_or_class)
+            when Types::InstanceType
+              return Types.class_name_of(t.klass)
+            end
+          end
+          nil
+        end
+
+        case @result
+        in [:call_or_const, type, _name, _self_call]
+          call_or_const_doc.call type
+        in [:const, type, _name, scope]
+          if type
+            call_or_const_doc.call type
+          else
+            value_doc.call scope[name]
+          end
+        in [:gvar, _name, scope]
+          value_doc.call scope["$#{name}"]
+        in [:ivar, _name, scope]
+          value_doc.call scope["@#{name}"]
+        in [:cvar, _name, scope]
+          value_doc.call scope["@@#{name}"]
+        in [:call, type, _name, _self_call]
+          method_doc.call type
+        in [:lvar_or_method, _name, scope]
+          if scope.local_variables.include?(name)
+            value_doc.call scope[name]
+          else
+            method_doc.call scope.self_type
+          end
+        else
+        end
+      end
+
+      def candidates_from_result(result)
+        candidates = case result
+        in [:require, name]
+          retrieve_files_to_require_from_load_path
+        in [:require_relative, name]
+          retrieve_files_to_require_relative_from_current_dir
+        in [:call_or_const, type, name, self_call]
+          ((self_call ? type.all_methods : type.methods).map(&:to_s) - HIDDEN_METHODS) | type.constants
+        in [:const, type, name, scope]
+          if type
+            scope_constants = type.types.flat_map do |t|
+              scope.table_module_constants(t.module_or_class) if t.is_a?(Types::SingletonType)
+            end
+            (scope_constants.compact | type.constants.map(&:to_s)).sort
+          else
+            scope.constants.sort
+          end
+        in [:ivar, name, scope]
+          ivars = scope.instance_variables.sort
+          name == '@' ? ivars + scope.class_variables.sort : ivars
+        in [:cvar, name, scope]
+          scope.class_variables
+        in [:gvar, name, scope]
+          scope.global_variables
+        in [:symbol, name]
+          Symbol.all_symbols.map { _1.inspect[1..] }
+        in [:call, type, name, self_call]
+          (self_call ? type.all_methods : type.methods).map(&:to_s) - HIDDEN_METHODS
+        in [:lvar_or_method, name, scope]
+          scope.self_type.all_methods.map(&:to_s) | scope.local_variables
+        else
+          []
+        end
+        [name || '', candidates]
+      end
+
+      def analyze(code, binding = Object::TOPLEVEL_BINDING)
+        # Workaround for https://github.com/ruby/prism/issues/1592
+        return if code.match?(/%[qQ]\z/)
+
+        lvars_code = binding.local_variables.map do |name|
+          "#{name}="
+        end.join + "nil;\n"
+        code = lvars_code + code
+        ast = Prism.parse(code).value
+        name = code[/(@@|@|\$)?\w*[!?=]?\z/]
+        *parents, target_node = find_target ast, code.bytesize - name.bytesize
+        return unless target_node
+
+        calculate_scope = -> { TypeAnalyzer.calculate_target_type_scope(binding, parents, target_node).last }
+        calculate_type_scope = ->(node) { TypeAnalyzer.calculate_target_type_scope binding, [*parents, target_node], node }
+
+        case target_node
+        when Prism::StringNode, Prism::InterpolatedStringNode
+          call_node, args_node = parents.last(2)
+          return unless call_node.is_a?(Prism::CallNode) && call_node.receiver.nil?
+          return unless args_node.is_a?(Prism::ArgumentsNode) && args_node.arguments.size == 1
+
+          case call_node.name
+          when :require
+            [:require, name.rstrip]
+          when :require_relative
+            [:require_relative, name.rstrip]
+          end
+        when Prism::SymbolNode
+          if parents.last.is_a? Prism::BlockArgumentNode # method(&:target)
+            receiver_type, _scope = calculate_type_scope.call target_node
+            [:call, receiver_type, name, false]
+          else
+            [:symbol, name] unless name.empty?
+          end
+        when Prism::CallNode
+          return [:lvar_or_method, name, calculate_scope.call] if target_node.receiver.nil?
+
+          self_call = target_node.receiver.is_a? Prism::SelfNode
+          op = target_node.call_operator
+          receiver_type, _scope = calculate_type_scope.call target_node.receiver
+          receiver_type = receiver_type.nonnillable if op == '&.'
+          [op == '::' ? :call_or_const : :call, receiver_type, name, self_call]
+        when Prism::LocalVariableReadNode, Prism::LocalVariableTargetNode
+          [:lvar_or_method, name, calculate_scope.call]
+        when Prism::ConstantReadNode, Prism::ConstantTargetNode
+          if parents.last.is_a? Prism::ConstantPathNode
+            path_node = parents.last
+            if path_node.parent # A::B
+              receiver, scope = calculate_type_scope.call(path_node.parent)
+              [:const, receiver, name, scope]
+            else # ::A
+              scope = calculate_scope.call
+              [:const, Types::SingletonType.new(Object), name, scope]
+            end
+          else
+            [:const, nil, name, calculate_scope.call]
+          end
+        when Prism::GlobalVariableReadNode, Prism::GlobalVariableTargetNode
+          [:gvar, name, calculate_scope.call]
+        when Prism::InstanceVariableReadNode, Prism::InstanceVariableTargetNode
+          [:ivar, name, calculate_scope.call]
+        when Prism::ClassVariableReadNode, Prism::ClassVariableTargetNode
+          [:cvar, name, calculate_scope.call]
+        end
+      end
+
+      def find_target(node, position)
+        location = (
+          case node
+          when Prism::CallNode
+            node.message_loc
+          when Prism::SymbolNode
+            node.value_loc
+          when Prism::StringNode
+            node.content_loc
+          when Prism::InterpolatedStringNode
+            node.closing_loc if node.parts.empty?
+          end
+        )
+        return [node] if location&.start_offset == position
+
+        node.compact_child_nodes.each do |n|
+          match = find_target(n, position)
+          next unless match
+          match.unshift node
+          return match
+        end
+
+        [node] if node.location.start_offset == position
+      end
+
+      def handle_error(e)
+      end
+    end
+  end
+end
--- a/lib/irb/type_completion/completor.rb
+++ b/lib/irb/type_completion/completor.rb
@@ -9,6 +9,10 @@ module IRB
     class Completor < BaseCompletor # :nodoc:
       HIDDEN_METHODS = %w[Namespace TypeName] # defined by rbs, should be hidden
 
+      class << self
+        attr_accessor :last_completion_error
+      end
+
       def inspect
         name = 'TypeCompletion::Completor'
         prism_info = "Prism: #{Prism::VERSION}"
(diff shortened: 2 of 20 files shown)


Morriar:
Using static analysis to provide auto-completion is a surprising direction for IRB which already has access to a lot of dynamic information.

It seems like you are reimplementing features that belong more to a static type checker such as Steep or Sorbet or IDE features already covered by the Ruby LSP. I wonder if IRB is the right place for this?

tompng:
> It seems like you are reimplementing features that belong more to a static type checker

Yes, it's a reimplemention but has some difference from other static type checker.

- Few static information and many runtime information
  In IRB, most of the thing is dynamically defined. It doesn't analyze the whole project code but only analyze code in current input buffer.
  It is tightly coupled with `binding` by design to use many runtime information.
- Focuses on loose completion, not strict type check

So I think it is worth making a static + dynamic type analyzer specialized for REPL.

st0012:
> So I think it is worth making a static + dynamic type analyzer specialized for REPL.

Since we're going to need a full analyzer with dependency on `rbs`, could it be a standalone gem instead, and IRB just has an completion adapter for it? Similar to how Rails define adapters for background job libraries like `resque`, `sidekiq`....etc. but not directly including them.

lib/irb/type_completion/completor.rb
```diff
@@ -0,0 +1,238 @@
...
+            method_doc.call scope.self_type
+          end
+        else
+        end
+      end
+
+      def candidates_from_result(result)
+        candidates = case result
+        in [:require | :require_relative => method, name]
+          if IRB.const_defined? :RegexpCompletor # IRB::VERSION >= 1.8.2
```
st0012, line 102: If we don't plan to backport this to older versions, shouldn't `RegexpCompletor` always be present?

lib/irb/type_completion/completor.rb
```diff
@@ -0,0 +1,238 @@
...
+            path_completor = IRB::RegexpCompletor.new
+          elsif IRB.const_defined? :InputCompletor # IRB::VERSION <= 1.8.1
+            path_completor = IRB::InputCompletor
+          end
+          if !path_completor
+            []
+          elsif method == :require
+            path_completor.retrieve_files_to_require_from_load_path
+          else
+            path_completor.retrieve_files_to_require_relative_from_current_dir
```
st0012, line 112: These could simply be `retrieve_files_to_require_from_load_path` as it inherits `BaseCompletor`?

lib/irb/type_completion/completor.rb
```diff
@@ -0,0 +1,238 @@
...
+        code = lvars_code + code
+        ast = Prism.parse(code).value
+        name = code[/(@@|@|\$)?\w*[!?=]?\z/]
+        *parents, target_node = find_target ast, code.bytesize - name.bytesize
+        return unless target_node
+
+        calculate_scope = -> { TypeAnalyzer.calculate_target_type_scope(binding, parents, target_node).last }
+        calculate_type_scope = ->(node) { TypeAnalyzer.calculate_target_type_scope binding, [*parents, target_node], node }
+
+        if target_node.is_a?(Prism::StringNode) || target_node.is_a?(Prism::InterpolatedStringNode)
```
st0012, line 160: We should be able to utilise `Prism::Dispatcher` and avoid these `is_a?(NodeClass)` conditions and the case statement below. You can use this [test](https://github.com/ruby/prism/blob/ac693764ae14c293f800611ce4e23e60dfcc64d7/test/prism/dispatcher_test.rb#L6) as an example.

lib/irb/type_completion/type_analyzer.rb
```diff
@@ -0,0 +1,1168 @@
...
+      }
+
+      def initialize(dig_targets)
+        @dig_targets = dig_targets
+      end
+
+      def evaluate(node, scope)
+        method = "evaluate_#{node.type}"
+        if respond_to? method
+          result = send method, node, scope
```
st0012, line 46: This looks very much like a visitor pattern. Can we use `Prism::Visitor` instead?

lib/irb/type_completion/completor.rb
tompng: Combined this `is_a?` into `case target_node when`.

lib/irb/type_completion/type_analyzer.rb
tompng: Yes, it's like a visitor pattern, similar to Prism::Compiler that visit returns a value.
But analyzing also needs different setup and different argument for each child visit, needs to control visit order.

Implementing arguments to visitor will look like this
```ruby
def visit_with_args(node, args)
  @args_stack << args
  visit(node)
  @args_stack.pop
end

def visit_foobar_node(node)
  args = @args_stack.last
  v0 = calculate_something(node.child_nodes[0].name) # sometimes it needs to skip recursive visit for some node
  setup_for_2
  v1 = visit_with_arg(node.child_nodes[2], args_for_2)
  setup_for_1
  v2 = visit_with_arg(node.child_nodes[1], args_for_1)
  return calculate_something(v0, v1, v2, args)
end
```

Needs setup and arguments:
for IfNode, it needs setup, teardown and different varaible-scope-branch arguments for predicate, statements and consequent.

Needs to control visit order:
for MultiWrite node, it should visit receiver-of-write-target first, then visit value part, then visit write-targets with values to assign.

In this case, I think using visitor with stack is not as intuitive as a simple recursive visitor-like implementation.

kddnewton: Hopefully this will help: https://github.com/ruby/prism/pull/1740

kddnewton: Actually I'm closing that PR because @eregon had a much better suggestion, see here: https://github.com/ruby/prism/pull/1740#issuecomment-1781426941

eregon: Additionally using Prism::Compiler should be quite a bit faster because it avoids the `respond_to?` and `send` with a dynamic method name.

> Needs to control visit order:

A visitor/Prism::Compiler can do that, you can choose the order in which you call `visit` on the children (or not call `visit` at all and handle the children in the same method). There is no need to use `visit_all`/`visit_child_nodes`.

tompng: Thanks for the information. I like visit_with. It looks possible to implement with recreating visitor.

However, I think some of the pros of visitor is lost in this case.
I think simple dispatch with `send if respond_to?` is enough, and for this case, using visitor might introduce complexity, so I want to let this implementation as is.

st0012: > However, I think some of the pros of visitor is lost in this case.

> using visitor might introduce complexity

Could you explain with examples to help me get the context?

tompng: Benefit is that the implementation is simple.

These pros of using Visitor are lost in this case.
- Only need to implement few method focusing on: in this case, we need to implement every single one.
- Does not need to know how to traverse: in this case, we need to specify traverse order for every node.
- Fast: Recreating visitor will make it slightly slow (and make code complex)
- Double dispatch mechanism: Dispatch is easy in ruby compared to other language.

If the pros is relatively small, using visitor will just make the code complicated.

This is the current implementation
```ruby
def evaluate(node, scope); dispatch using send; end

def evaluate_foo_node(node, scope)
  scope2 = scope.new_child_scope
  a = evaluate(node.a, scope2)
  b = evaluate(node.b, scope2)
  scope.update(scope2)
  type_of_foo_calculation(a,b)
end
```

Using visitor will look like this
```ruby
def visit_foo_node(node)
  scope2 = @scope.new_child_scope
  a = self.class.new(@bar, scope2).visit(node.a) # boilerplate to visit with args
  b = self.class.new(@bar, scope2).visit(node.b)
  @scope.update(scope2)
  type_of_foo_calculation(a,b)
end
```

The boilerplate can be cleared by this
```ruby
def evaluate(node, scope)
  self.class.new(@bar, scope).visit(node)
end

def visit_foo_node(node)
  scope2 = @scope.new_child_scope
  a = evaluate(node.a, scope2)
  b = evaluate(node.b, scope2)
  @scope.update(scope2)
  type_of_foo_calculation(a,b)
end
```

But as you see, it's almost same to current implementation. It's essentially same but with some complexity overhead.
This is just using visitor as method-dispatching that can be also obtained by `send`. Using Visitor class but not using its traverse mechanism is confusing.
`self` might be changed while visiting nodes. It means that setting instance variable has no effect to the root visitor instance and the fact should be recognized while maintaining.
`node` and `@scope` are calculation parameter of `visit_foo_node(node)` but `@scope` is not in the method args. In the current implementation, `evaluate_foo_node(node, scope)` explicitly represents it.

lib/irb/type_completion/completor.rb
```diff
@@ -0,0 +1,232 @@
...
+            node.value_loc
+          when Prism::StringNode
+            node.content_loc
+          when Prism::InterpolatedStringNode
+            node.closing_loc if node.parts.empty?
+          end
+        )
+        return [node] if location&.start_offset == position
+
+        node.child_nodes.each do |n|
```
kddnewton, line 217: You can use `compact_child_nodes` here to make this a little faster

kddnewton (approved):
Honestly this looks great. It's super impressive how much work you've put in.

Please let me know if there's anything in the prism API that could be added to help you.

lib/irb/input-method.rb
```diff
@@ -270,6 +274,11 @@ def initialize
       end
     end
 
+    def completion_info
+      autocomplete_message = IRB.conf[:USE_AUTOCOMPLETE] ? 'Autocomplete' : 'Tab Complete'
```
st0012, line 278: Let's use `Reline.autocompletion` as the source of truth here? Otherwise this message would change with `IRB.conf[:USE_AUTOCOMPLETE]` while the actual autocompletion setting will not.

lib/irb/type_completion/completor.rb
```diff
@@ -0,0 +1,230 @@
+require 'prism'
+require 'irb/completion'
+require_relative 'type_analyzer'
+module IRB
```
st0012, line 4: nit: let's add an empty line here.

lib/irb/type_completion/completor.rb
```diff
@@ -0,0 +1,230 @@
+require 'prism'
```
st0012, line 1: Let's add `# frozen_string_literal: true`
```suggestion
# frozen_string_literal: true

require 'prism'
```

lib/irb/type_completion/completor.rb
```diff
@@ -0,0 +1,230 @@
...
+        []
+      ensure
+        $VERBOSE = verbose
+      end
+
+      def doc_namespace(preposing, matched, postposing, bind:)
+        name = matched[/[a-zA-Z_0-9]*[!?=]?\z/]
+        method_doc = -> type do
+          type = type.types.find { _1.all_methods.include? name.to_sym }
+          if type.is_a? Types::SingletonType
```
st0012, line 43: At line 60 you used case statement for the exact conditions. Let's be consistent and do the same here too.

lib/irb/type_completion/types.rb
```diff
@@ -0,0 +1,443 @@
...
+        aa, bb = [atypes, btypes].map {|types| (types[InstanceType] || []).map(&:klass) }
+        (aa.flat_map(&:ancestors) & bb).any?
+      end
+
+      def self.type_from_object(object)
+        case object
+        when Array, Hash, Module
+          type_from_object_recursive(object, max_level: 4)
+        else
+          klass = object.singleton_class rescue object.class
```
st0012, line 137: If the object inherits directly from `BasicObject`, this will cause error because it likely won't have neither `singleton_class` nor `class` method. Please add a few cases for `BasicObject` because they actually appear more often than we thought.
Also, perhaps it's better to implement the fallback mechanism at this level?

lib/irb/type_completion/types.rb
```diff
@@ -0,0 +1,443 @@
...
+          InstanceType.new klass
+        end
+      end
+
+      def self.type_from_object_recursive(object, max_level:)
```
st0012, line 142: This method is rather complicated. Can we have some unit tests for it?

I'm also a bit worry about the performance impact of resolving types recursively. If say we have a tree structure that's built with a hash, users would probably only need completion on it for 1 or 2 levels? But then we'd need to resolve the entire tree for it, which could be super expensive.

Do you think we can hold off on this for now, and wait for users' request on it?

lib/irb/type_completion/types.rb
```diff
@@ -0,0 +1,443 @@
...
+          else
+            key_types = keys.map(&:class).uniq.map { InstanceType.new _1 }
+            value_types = values.map(&:class).uniq.map { InstanceType.new _1 }
+            InstanceType.new Hash, { K: UnionType[*key_types], V: UnionType[*value_types] }
+          end
+        when Module
+          SingletonType.new object
+        else
+          InstanceType.new object.class
+        end
```
st0012, line 169: I wonder if we should have a rescue here as a circuit breaker and log the exception to help issue reporting?

lib/irb/type_completion/types.rb
```diff
@@ -0,0 +1,443 @@
...
+      def self.type_from_object_recursive(object, max_level:)
+        max_level -= 1
+        sample_size = 1000
+        case object
+        when Array
+          values = object.size > sample_size ? object.sample(sample_size) : object
+          if max_level > 0
+            InstanceType.new Array, { Elem: UnionType[*values.map { type_from_object_recursive(_1, max_level: max_level) }] }
+          else
+            InstanceType.new Array, { Elem: UnionType[*values.map(&:class).uniq.map { InstanceType.new _1 }] }
```
st0012, line 151: `BasicObject` would break this.

lib/irb/type_completion/types.rb
```diff
@@ -0,0 +1,443 @@
...
+          end
+        when Hash
+          keys = object.size > sample_size ? object.keys.sample(sample_size) : object.keys
+          values = object.size > sample_size ? object.values.sample(sample_size) : object.values
+          if max_level > 0
+            key_types = keys.map { type_from_object_recursive(_1, max_level: max_level) }
+            value_types = values.map { type_from_object_recursive(_1, max_level: max_level) }
+            InstanceType.new Hash, { K: UnionType[*key_types], V: UnionType[*value_types] }
+          else
+            key_types = keys.map(&:class).uniq.map { InstanceType.new _1 }
```
st0012, line 161: `BasicObject` would break this.

lib/irb/type_completion/types.rb
```diff
@@ -0,0 +1,443 @@
...
+            value_types = values.map(&:class).uniq.map { InstanceType.new _1 }
```
st0012, line 162: `BasicObject` would break this.

lib/irb/type_completion/types.rb
```diff
@@ -0,0 +1,443 @@
...
+            InstanceType.new Hash, { K: UnionType[*key_types], V: UnionType[*value_types] }
+          end
+        when Module
+          SingletonType.new object
+        else
+          InstanceType.new object.class
```
st0012, line 168: `BasicObject` would break this.

lib/irb/type_completion/scope.rb
```diff
@@ -0,0 +1,408 @@
...
+      BREAK_RESULT = '%break'
+      NEXT_RESULT = '%next'
+      RETURN_RESULT = '%return'
+      PATTERNMATCH_BREAK = '%match'
+
+      attr_reader :module_nesting, :self_object
+
+      def initialize(binding, self_object, local_variables)
+        @binding = binding
+        @self_object = self_object
```
st0012, line 17: This is just a reminder: `self_object` is `BasicObject`, many methods here will break. But IRB doesn't really work inside a `BasicObject` anyway (other things would break first), we can probably get away with it for now.

lib/irb/type_completion/scope.rb
```diff
@@ -0,0 +1,408 @@
+# frozen_string_literal: true
+
+require 'set'
+require_relative 'types'
+module IRB
+  module TypeCompletion
+    class BaseScope
```
st0012, line 7: Why do we need to extract `BaseScope` out of `Scope`?

test/irb/type_completion/test_scope.rb
```diff
@@ -0,0 +1,113 @@
+# frozen_string_literal: true
+
+return unless RUBY_VERSION >= '3.0.0'
+return if RUBY_ENGINE == 'truffleruby' # needs endless method definition
+
+require 'irb/type_completion/scope'
+require 'irb/type_completion/types'
```
st0012, line 7: I think `scope` already requires `types`?

lib/irb/type_completion/types.rb
tompng: Nice catch :+1:

Currently, the level is restricted to 4, and max sampling size for each hash/array is 1000, but for very complicated structure (large or recursive), the performance was terrible.
```ruby
irb> a=[];50.times{a<<a}
irb> a. # takes about one second
```

> Do you think we can hold off on this for now, and wait for users' request on it?

Yes, we should hold off. 1 level is enough for the first version.

Possible future choices:
- carefully increase level
- restrict by total count of type object, not by depth
- lazy evaluation

lib/irb/type_completion/scope.rb
tompng: Scope represents mutable virtual scope that has table of what variable will be changed during type analysis.
BaseScope represents immutable existing scope in current binding. It does not have table of variable changes. It just retrieves the actual value in current binding.

`base_scope['lvar']` will execute
```ruby
binding.eval('lvar') if binding.local_variables.include?(:lvar)
```

scope['lvar'] will execute
```ruby
if @table.has? 'lvar'
  # When lvar is changed in the input, (example: `lvar = lvar.foo; lvar.`)
  # Type of lvar.foo is stored in @table['lvar']
  @table['lvar']
else
  parent_scope['lvar'] # parent_scope is base_scope or another scope. root scope is always base_scope
end
```

st0012: It seems that `Scope` actually doesn't share many ivars with `BaseScope`. It also doesn't utilise `BaseScope`'s implementation much (e.g. calling `super`). Furthermore, a `Scope` always has a parent while `BaseScope` doesn't. The things they mostly share seem to be constants like `BREAK_RESULT`?

These make me wonder if using inheritance here is the best option. They seem to be very different structure that just respond to the same interfaces. If that's the case, I think `Scope` shouldn't inherit `BaseScope` as it'd cause a wrong impression to readers.

Perhaps renaming them to `RootScope` and `LeafScope` will also help clarify their relationships.

lib/irb/type_completion/types.rb
```diff
@@ -0,0 +1,448 @@
+# frozen_string_literal: true
+
+module IRB
+  module TypeCompletion
+    module Types
+      singleton_class.attr_reader :rbs_builder, :rbs_load_error
+
+      def self.preload_in_thread
+        @loader_thread ||= Thread.new do
```
st0012, line 9: In "production" environment, this will only be called once. Do we need to memoize it?

lib/irb/type_completion/types.rb
tompng: I completely forgot about BasicObject. fixed in 63693d3268f3ef1fa559badc0e2f2159b181cb10

Fallback mechanism like `BaseScope.type_of(fallback: Types::NIL){b.local_variable_get name}` is actually a fallback of `local_variable_get(name)` causing NameError.

I think type_from_object normally does not raise error after BasicObject fix.
I changed to explicitly rescue the expected error and remove fallback mechanism, because current fallback will suppress and might hide important error.
```
BaseScope.type_of(fallback: fallback) { @binding.local_variable_get(name) }
# ↓
value = begin
  @binding.local_variable_get(name)
rescue NameError
end
Types.type_from_object(value)
```
8a0fb2e4743a9b97590272123ee2c6289393383f

lib/irb/type_completion/types.rb
tompng: I think it has some benefits.

The original pull-request came to katakata_irb is an option of async loading because in large project, it takes more than seconds.
The default is async, but users can change to sync, always complete loading before IRB starts, by writing `IRB::TypeCompletion::Types.preload_in_thread.join` to irbrc.
I think it will help when we enable switching completors with irb commands like `irb(main):001> conf.completor = :type`.

I also think it's safer. (for multi-irb, executing `binding.irb` inside irb, `binding.irb; binding.irb` or other case)

lib/irb/type_completion/scope.rb
tompng: I see. Yes, it's not a inheritance. RootScope name looks good.
I think LeafScope is not accurate because it can be a parent of another LeafScope, so I'll keep the name Scope.

lib/irb/type_completion/types.rb
st0012: IMO `IRB::TypeCompletion::Types.preload_in_thread` should be considered private. If we want to let users control async/sync loading, we should encapsulate it under our config design.

> I think it will help when we enable switching completors with irb commands

I don't think it's very useful to switch completor mid-session other than testing/debugging IRB itself?

tompng: Agree that it should be private.

Actually, in production running this ruby file, preload_in_thread is ran twice.
```ruby
# run with `ruby a.rb`
2.times do
  binding.irb
end
```
Typing `exit` fast with slow_loading environment, (many entry in rbs_collection.yaml and sig directory), preload_in_thread is ran twice before `@rbs_builder` is assigned.

To avoid preloading twice, we need to cache something outside of thread because thread creation can take a while. Cannot use `@rbs_builder` for this because assigning to it will delay. Other way is to see the return value of require and run preload_in_thread only if it is true.


In test, directly calling load_rbs_builder looks good. I'm thinking to change it to
```ruby
def self.preload_in_thread
  return if @preload_started
  @preload_started = true
  Thread.new do
    load_rbs_builder
  end
end
```

lib/irb/type_completion/types.rb
tompng: done in 12efb40d0edb0078ae2d0a532dd225b54ef6ba25

lib/irb/type_completion/completor.rb
```diff
@@ -0,0 +1,239 @@
...
+
+      def analyze(code, binding = Object::TOPLEVEL_BINDING)
+        # Workaround for https://github.com/ruby/prism/issues/1592
+        return if code.match?(/%[qQ]\z/)
+
+        lvars_code = binding.local_variables.map do |name|
+          "#{name}="
+        end.join + "nil;\n"
+        code = lvars_code + code
+        ast = Prism.parse(code).value
```
kddnewton, line 151: In the latest version of prism, you can pass in the surrounding scopes to the parse which should help with this. It would look something like:

```ruby
ast = Prism.parse(code, scopes: [binding.local_variables])
```

tompng: Thank you for adding this option :+1:
updated

lib/irb/type_completion/types.rb
tompng: Added test for BasicObject. `Types.type_from_object` is not wrapped with begin-rescue with fallback type anymore. So now exception will be caught and stored to `Completor.last_completion_error` for issue reporting.

lib/irb/type_completion/completor.rb
st0012: Since we’re using a new API, let’s also add a version check to `build_completor`?

st0012: Since we’re using a new API, let’s also add a version check to `build_completor`?

tompng: Added. also add warn message why using `:type` completor failed.

st0012 (approved):
Amazing work 👍

hasumikin (approved):
I'd agree with shipping this feature with Ruby 3.3

Functions I would like to add gradually:
- Freely change the completer within the IRB session
- To display a kind of a "backtrace" of how the completion candidates are created with one simple command

ima1zumi (approved):
✨

PR #710, opened 2023-09-12
Mention `irb_debug` alternative in the docs

janko:
When starting an IRB session via `binding.irb` inside a Rails view template, the `debug` method cannot be used because it's already defined in Action View.

I only found out about `irb_debug` from the IRB warning `can't alias debug from irb_debug`, so I thought it would be useful to mention it in the documentation.


st0012 (approved):
That's a nice addition, thanks 👍

PR #720, opened 2023-09-23
Disable pager in show-source test

tompng:
Test was failing with `rake test | ruby -e "loop{puts gets || exit}"`


st0012 (approved):
Perhaps we can make it a general setup for `CommandTestCase`?
Also, I'm curious about the reason to run `rake test | ruby -e "loop{puts gets || exit}"`

tompng:
> general setup for CommandTestCase

done :+1:

I found a blog post https://jinroq.hatenablog.jp/entry/2023/09/23/191254 that says ShowSourceTest is failing.

I don't know how the test is executed, but I found that `rake test` succeed and `rake test | any_command` failed.
`rake test > test_output.txt` also fails.

PR #726, opened 2023-10-11
Fix require path completion disturbing string method completion

### LINKED ISSUE #715: Completion does not work for Strings end with some symbolic characters
## Description

Completion does not seem to work for Strings end with some symbolic characters.

### Good

Type `"hoge+".len`.

```
irb(main):001* "hoge+".len
               "hoge+".length
```

Completion candidates are displayed.

Type <kbd>TAB</kbd>.

```
irb(main):001> "hoge+".length
               "hoge+".lengthPress Alt+d to read the full document
                             String.length

                             (from ruby core)
                             ----------------------------------------
                             :
                             :
```

The method is completed.

### Bad

Type `"hoge|".len`.

```
irb(main):001* "hoge|".len
```

No completion candidates are displayed.

Type <kbd>TAB</kbd>.

```
irb(main):001> "hoge|".len
```

The method is not completed.

### Other characters

Good: `!` `"` `#` `$` `%` `'` `)` `*` `+` `,` `-` `.` `/` `:` `?` `@` `[` `\` `]` `^` `_` `}` `~`

Bad: ` ` `&` `(` `;` `<` `=` `>` `` ` `` `{` `|`

## Result of irb_info

```
irb(main):001> irb_info
Ruby version: 3.1.4
IRB version: irb 1.8.1 (2023-09-04)
InputMethod: RelineInputMethod with Reline 0.3.8
RUBY_PLATFORM: amd64-freebsd13
LANG env: ja_JP.UTF-8
East Asian Ambiguous Width: 2
=> nil
irb(main):002>
```

## Terminal Emulator

PuTTYrv 0.79-ranvis (CJK)

## Setting Files

No `~/.irbrc` or `~/.inputrc`.

tompng:
Fixes #715 

`complete_require_path` should return `nil` if the input is not `require "path` or `require_relative "path`.
If it returns `[]` instead of `nil` in non require pattern, other completion calculation is skipped.


lib/irb/completion.rb
```diff
@@ -132,24 +132,22 @@ def complete_require_path(target, preposing, postposing)
...
-            path.start_with?(actual_target)
-          }.map { |path|
-            quote + path
-          }
-        end
+      return unless tok&.event == :on_ident && tok.state == Ripper::EXPR_CMDARG
+
+      case tok.tok
+      when 'require'
+        result = retrieve_files_to_require_from_load_path.select { |path|
```
tompng, line 139: warning: assigned but unused variable - result

st0012 (approved):

PR #730, opened 2023-10-14
Restore IRB::InputCompletor for compatibility

tompng:
`CompletionProc` and `retrieve_completion_data` seems to be used
`ReservedWords` `Operators` is used in one repository
Other constants and methods does not seem to be used


st0012 (approved):
I can confirm this fixes `debug`'s test failure 👍

PR #736, opened 2023-10-21
Minor refactors around irb.rb

st0012:
With recent changes around `RubyLex`, `irb.rb`'s size has grown quite a bit. So I think it's worth doing some cleanup and put things at or conventional places to make it easier to navigate.


tompng (approved):
:+1:

PR #737, opened 2023-10-21
Easter egg for autocomplete mode

tompng:
When you enter `RubyVM[TAB]`, IRB shows an easter egg instead of document in `irb --noautocomplete` mode if `ENV['RUBY_YES_I_AM_NOT_A_NORMAL_USER']` is not set.
I beleave this is to avoid normal user easily use `RubyVM`, but this feature(?) was not working in autocomplete mode.

This pull request adds easter egg for autocomplete mode. Shows easter egg ruby logo inside document dialog, shows more easter egg (`:dancing` and `:logo`) when alt+d is pressed.

dialog_easteregg


ima1zumi (approved):
LGTM 💎 

image

tompng:
data is generated by https://gist.github.com/tompng/696e978363535ce7289d105a2200b226

PR #738, opened 2023-10-22
[DOC] RDoc for module IRB

BurdetteLamar:
I think everything from the beginning of the module doc down through section "Color Highlighting" is review-worthy.  Below that, not, so I'll mark this as a draft.

I'm striving organize this by features, and not by command-line options and IRB.conf entries. But to make sure they're available as two coherent groups I've added an index for each.

It's taken me weeks to do this much, and will likely take weeks to complete and get reviewed/approved.

It may be good to review/revise piecemeal.


lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# If the search fails, there is no configuration file.
+#
+# If the search succeeds, the configuration file is read as Ruby code,
+# and so can contain any Ruby programming you like.
+#
+# \Method <tt>conf.rc?</tt> returns +true+ if a configuration file was read,
+# +false+ otherwise.
+# \Hash entry <tt>IRB.conf[:RC]</tt> also contains that value.
+#
+# [REVIEWERS] What should be said about IRB.conf[:RC_NAME_GENERATOR]?
```
peterzhu2118, line 155: I think`IRB.conf[:RC_NAME_GENERATOR]` is an internal thing, so maybe we don't need to mention it?

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# == Input
+#
+# This section describes the features that allow you to change
+# the way \IRB input works;
+# see also {Input and Output}[rdoc-ref:IRB@Input+and+Output].
+#
+# === Input Command History
+#
+# By default, \IRB stores a history of up to 1000 input commands
+# in file <tt>./.irb_history</tt>.
```
peterzhu2118, line 217: I'm pretty sure it stores it in the home directory rather than the current directory.

```suggestion
# in file <tt>~/.irb_history</tt>.
```

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# to the configuration file.
+#
+# During the session, method <tt>conf.ignore_eof?</tt> returns the setting,
+# and method <tt>conf.ignore_eof = _boolean_</tt> sets it.
+#
+# === SIGINT
+#
+# By default, <tt>IRB.conf[:IGNORE_SIGINT]</tt> is +true+,
+# which means that typing the interrupt character <tt>Ctrl-C</tt>
+# does not cause the session to exit.
```
peterzhu2118, line 296: Ditto.

```suggestion
# which means that typing the interrupt character <tt>Ctrl-C</tt>
# causes the session to exit.
```

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# === Automatic Completion
+#
+# By default, \IRB enables
+# {automatic completion}[https://en.wikipedia.org/wiki/Autocomplete#In_command-line_interpreters]:
+#
+# You can disable it by either of these:
+#
+# - Adding <tt>IRB.conf[:USE_AUTOCOMPLETE] = false</tt> to the configuration file.
+# - Giving command-line option <tt>--noautocomplete</tt>
+#   ()<tt>--autocomplete</tt> is the default).
```
peterzhu2118, line 313:
```suggestion
#   (<tt>--autocomplete</tt> is the default).
```

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# - Giving command-line options:
+#
+#   - <tt>--singleline</tt>
+#     or <tt>--nomultiline</tt>
+#     sets the input method to IRB::ReadlineInputMethod.
+#   - <tt>--nosingleline</tt>
+#     or <tt>--multiline/tt>
+#     sets the input method to IRB::RelineInputMethod.
+#
+# See {Notes on Initialization Precedence}[rdoc-ref:IRB@Notes+on+Initialization+Precedence].
```
peterzhu2118, line 371: I'm not a fan of putting redirection to `Initialization Precedence` all over the place. IMO the user should be aware of the different ways of setting configuration.

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# If the search succeeds, the configuration file is read as Ruby code,
+# and so can contain any Ruby programming you like.
+#
+# \Method <tt>conf.rc?</tt> returns +true+ if a configuration file was read,
+# +false+ otherwise.
+# \Hash entry <tt>IRB.conf[:RC]</tt> also contains that value.
+#
+# [REVIEWERS] What should be said about IRB.conf[:RC_NAME_GENERATOR]?
+#
+# === Session \Context
```
peterzhu2118, line 157: I think we should move most of the documentation after this related to `IRB::Context` into the documentation for `IRB::Context`.

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# By default, \IRB saves no history of evaluations (returned values),
+# and the related methods <tt>conf.eval_history</tt>, <tt>_</tt>,
+# and <tt>__</tt> are undefined.
+#
+# You can turn on that history, and set the maximum number of evaluations to be stored:
+#
+# - In the configuration file: add <tt>IRB.conf[:EVAL_HISTORY] = _n_</tt>.
+#   (Examples below assume that we've added <tt>IRB.conf[:EVAL_HISTORY] = 5</tt>.)
+# - In the session (at any time): <tt>conf.eval_history = _n_</tt>.
+#   [BUG] Raises exception "undefined method `push' for nil:NilClass (NoMethodError)."
```
peterzhu2118, line 502: This doesn't bug out for me (IRB 1.8.3 on macOS):

```
irb(main):003> conf.eval_history = 10
=> 10
irb(main):004> "hi"
=> "hi"
irb(main):005> _
=> "hi"
irb(main):006> __
=> 
3 10
4 "hi"
5 "hi"
```

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# You can enable verbosity by:
+#
+# - Adding to the configuration file: <tt>IRB.conf[:VERBOSE] = true</tt>
+#   (the default is +nil+).
+# - Giving command-line options <tt>--verbose</tt>
+#   (the default is <tt>--noverbose</tt>).
+#
+# See {Notes on Initialization Precedence}[rdoc-ref:IRB@Notes+on+Initialization+Precedence].
+#
+# During a session, you can retrieve or set verbosith with methods
```
peterzhu2118, line 721:
```suggestion
# During a session, you can retrieve or set verbosity with methods
```

lib/irb.rb
BurdetteLamar: It's user-visible in IRB.conf.  Is that a bug?

lib/irb.rb
BurdetteLamar: I was not happy with either, but the option/config-file precedence is the opposite of what one expects in a modern app.  I'll remove the links if you think it best.

lib/irb.rb
peterzhu2118: I think there's a lot of things user-visible, but are probably not significant enough to be documented. If you want to document it, then you can. I don't have strong feelings either way.

lib/irb.rb
peterzhu2118: I think it's distracting to the reader to have this line all over the place.

.document
```diff
@@ -1,4 +1,5 @@
 LICENSE.txt
 README.md
-doc
+doc/irb/indexes.rdoc
 lib/**/*.rb
+lib/irb/lc/help-message
```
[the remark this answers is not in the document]
BurdetteLamar: I have changed the link to point to the raw file on GH, which displays better anyway.

.gitignore
```diff
@@ -2,7 +2,7 @@
 /.yardoc
 /_yardoc/
 /coverage/
-/doc/
+/doc/irb/*.ja
```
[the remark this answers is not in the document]
BurdetteLamar: I don't want to ignore the indexes.rdoc page, so I thought this was needed.

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
 
-# IRB stands for "interactive Ruby" and is a tool to interactively execute Ruby
-# expressions read from the standard input.
+# == \IRB
+# 
+# \Module \IRB ("Interactive Ruby") provides a shell-like interface
+# that supports user interaction with the Ruby interpreter.
 #
-# The +irb+ command from your shell will start the interpreter.
+# Its operates as a <i>read-eval-print loop</i>
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+#
+# ==== Command-Line Options
+#
+# Many command-line options affect entries in hash <tt>IRB.conf</tt>,
+# which in turn affect the initial configuration of the \IRB session.
+#
+# Details of the options are described in the relevant subsections below.
+#
+# A cursory list of the \IRB command-line options
+# may be seen in the {help message}[rdoc-ref:lib/irb/lc/help-message],
```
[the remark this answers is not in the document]
BurdetteLamar: No longer necessary?

lib/irb.rb
BurdetteLamar: Removed.

lib/irb.rb
BurdetteLamar:
```$ ls -l .irb_history
-rw------- 1 burdette burdette 1205 Oct 24 11:12 .irb_history
$ irb
irb(main):001> 1 + 1
=> 2
irb(main):002> exit
$ ls -l .irb_history
-rw------- 1 burdette burdette 1216 Oct 24 11:31 .irb_history
```

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+#   irb(main):006:0> conf.command_aliases[:foo] = :show_source
+#   => :show_source
+#   irb(main):007:0> conf.command_aliases[:bar] = :whereami
+#   => :whereami
+#   irb(main):008:0> conf.command_aliases
+#   => {:foo=>:show_source, :bar=>:whereami}
+#   irb(main):009:0> conf.command_aliases.delete(:foo)
+#   => nil
+#   irb(main):010:0> conf.command_aliases
+#   => {:bar=>:whereami}
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+#
+# === End-of-File
+#
+# By default, <tt>IRB.conf[:IGNORE_EOF]</tt> is +false+,
+# which means that typing the end-of-file character <tt>Ctrl-D</tt>
+# does not cause the session to exit.
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

lib/irb.rb
BurdetteLamar: Fixed.

lib/irb.rb
BurdetteLamar: Fixed.

lib/irb.rb
BurdetteLamar: Removed.

lib/irb.rb
BurdetteLamar: Removed.  (Don't what I did to make me think this.)

lib/irb.rb
BurdetteLamar: Fixed.

lib/irb.rb
BurdetteLamar: I'm not clear on what exactly you think should be moved.  As I said, I'm trying to document as features, rather than implementation (options, IRB.conf entries -- and also classes).  For each feature I've tried to show how it's state is initialized, and what if anything can modify it downstream.  I'd like to leave all here, at least for now.

st0012: I think we should remove the `Context` section as I consider it a private component and I don't want to maintain backward-compatibility for it.

lib/irb.rb
```diff
@@ -21,36 +21,883 @@
...
+#   - <tt>--singleline</tt>
+#     or <tt>--nomultiline</tt>
+#     sets the input method to IRB::ReadlineInputMethod.
+#   - <tt>--nosingleline</tt>
+#     or <tt>--multiline/tt>
+#     sets the input method to IRB::RelineInputMethod.
+#
+# See {Notes on Initialization Precedence}[rdoc-ref:IRB@Notes+on+Initialization+Precedence].
+#
+# You can retrieve the current value using method <tt>conf.io</tt>:
```
st0012, line 352: Let's avoid exposing `conf.io` to users. If they need to see the input method being used, they should use the `irb_info` command.

st0012:
@BurdetteLamar Thank you so much for this update. I will review them in several batches 🙏

lib/irb.rb
BurdetteLamar: Removed.

lib/irb.rb
BurdetteLamar: Reference removed.  I did not mention `irb_into` b/c I want to handle all the commands at once (not looking forward to that!), and not in this PR.

lib/irb.rb
peterzhu2118: I still see two of these, including this one.

.gitignore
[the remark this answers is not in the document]
peterzhu2118: I think we can just completely delete this from the gitignore.

lib/irb.rb
BurdetteLamar: Removed.

.gitignore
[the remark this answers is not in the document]
BurdetteLamar: Removed.

BurdetteLamar:
@peterzhu2118, two fixes done.  (My re-review thingy did not come back on.)

lib/irb.rb
```diff
@@ -21,36 +21,917 @@
...
+# which in turn affect the initial configuration of the \IRB session.
+#
+# Details of the options are described in the relevant subsections below.
+#
+# A cursory list of the \IRB command-line options
+# may be seen in the {help message}[rdoc-ref:lib/irb/lc/help-message],
+# which is also displayed if you use command-line option <tt>--help</tt>.
+#
+# If you are interested in a specific option, consult the
+# {index}[./doc/irb/indexes_rdoc.html#label-Index+of+Command-Line+Options].
```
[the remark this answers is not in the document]
peterzhu2118: `{index}[rdoc-ref:doc/irb/indexes.rdoc@Index+of+Command-Line+Options]` works for me.

lib/irb.rb
BurdetteLamar: /home/burdette/Documents/GitHub/irb

lib/irb.rb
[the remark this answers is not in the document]
BurdetteLamar: Fixed (two places, slightly different).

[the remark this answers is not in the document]
BurdetteLamar: Fixed (two places, slightly different).

lib/irb.rb
```diff
@@ -21,36 +21,852 @@
...
+#
+# - The value of variable <tt>$IRBRC</tt>, if defined.
+# - The value of variable <tt>$XDG_CONFIG_HOME/irb/irbrc</tt>, if defined.
+# - File +$HOME/.irbrc+, if it exists.
+# - File +$HOME/.config/irb/irbrc+, if it exists.
+# - File +.config/irb/irbrc+ in the current directory, if it exists.
+# - File +.irbrc+ in the current directory, if it exists.
+# - File +irb.rc+ in the current directory, if it exists.
+# - File +_irbrc+ in the current directory, if it exists.
+# - File +$irbrc+ in the current directory, if defined.
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

lib/irb.rb
peterzhu2118: That's very odd. What if you deleted the `.irb_history` file in the currently directory? Does it create that file and append to it?

BurdetteLamar:
```
$ rm .irb_history
$ irb
irb(main):001> 1 + 1
=> 2
irb(main):002> exit
$ ls -l .irb_history
-rw------- 1 burdette burdette 11 Nov  4 12:53 .irb_history
$ cat .irbrc

$ 
```

tompng: IRB will save `.irb_history` to the same directory of `.irbrc`.
```
$ mkdir tmpdir
$ touch tmpdir/.irbrc
$ IRBRC=tmpdir/.irbrc irb
irb(main):001> irb_info
......
.irbrc path: tmpdir/.irbrc
......
irb(main):002> exit
$ ls -a tmpdir
./              ../             .irbrc          .irbrc_history
```

Normally, irb will save to `~/.irb_history`. If `.irbrc` exists in current directory, then history will saved to current directory.

BurdetteLamar: Ah, thanks much!

BurdetteLamar: @tompng, would it be correct to say that if any IRB resource file is found, the history file will be in that same directory?

BurdetteLamar: @tompng, I have changed the text at l. 202.

BurdetteLamar: Changed accordingly.

BurdetteLamar:
@st0012, @peterzhu2118: ready for review.

BurdetteLamar:
@st0012, @peterzhu2118:  Do you consider this still active?  If not, what sort of different approach can we take?

st0012:
@BurdetteLamar sorry that I had to prioritise feature/bug fixes for Ruby 3.3's release 🙏  I will review this PR in several gos soon.

I do have a question though: is it possible we can rewrite `irb.rb`'s documents in Markdown? I'm not familiar with RDoc format so I don't know if we'd lose anything by doing so.
I'd also prefer having `indexes.md` instead of `indexes.rdoc` so we align with `ruby/ruby`'s trend to embrace Markdown format. WDYT?

doc/irb/indexes.rdoc
```diff
@@ -0,0 +1,193 @@
...
+  initial value: +16+.
+- <tt>:COMMAND_ALIASES</tt>: Defines input {command aliases}[rdoc-ref:IRB@Command+Aliases];
+  initial value:
+
+    {
+      "$": :show_source,
+      "@": :whereami,
+      break: :irb_break,
+      catch: :irb_catch,
+      next: :irb_next
```
st0012, line 95: I recently hide these from users so we can remove them.

doc/irb/indexes.rdoc
```diff
@@ -0,0 +1,193 @@
...
+  See {Single-IRB Mode}[rdoc-ref:IRB@Single-IRB+Mode].
+- <tt>:USE_AUTOCOMPLETE</tt>: Whether to use
+  {automatic completion}[rdoc-ref:IRB@Automatic+Completion];
+  initial value: +true+.
+- <tt>:USE_COLORIZE</tt>: Whether to use
+  {color highlighting}[rdoc-ref:IRB@Color+Highlighting];
+  initial value: +true+.
+- <tt>:USE_LOADER</tt>: Whether to use the
+  {IRB loader}[rdoc-ref:IRB@IRB+Loader] for +require+ and +load+;
+  initial value: +false+.
```
st0012, line 186: I recently added `USE_PAGER` as well. It controls if pager is enabled for commands and return value output and defaults to `true`.

BurdetteLamar:
> @BurdetteLamar sorry that I had to prioritise feature/bug fixes for Ruby 3.3's release 🙏 I will review this PR in several gos soon.
> 
> I do have a question though: is it possible we can rewrite `irb.rb`'s documents in Markdown? I'm not familiar with RDoc format so I don't know if we'd lose anything by doing so. I'd also prefer having `indexes.md` instead of `indexes.rdoc` so we align with `ruby/ruby`'s trend to embrace Markdown format. WDYT?

Ok, will look at switching to Markdown.  Will mark this as a draft (once more).

lib/irb.rb
```diff
@@ -21,36 +21,863 @@
...
+# You can disable it by either of these:
+#
+# - Adding <tt>IRB.conf[:USE_AUTOCOMPLETE] = false</tt> to the configuration file.
+# - Giving command-line option <tt>--noautocomplete</tt>
+#   (<tt>--autocomplete</tt> is the default).
+#
+# \Method <tt>conf.use_autocomplete?</tt> returns +true+
+# if automatic completion is enabled, +false+ otherwise.
+#
+# The setting may not be changed during the session.
```
st0012, line 284: @tompng I think we can mention the completor config in the next section. Could you add some info for it?

lib/irb.rb
```diff
@@ -21,36 +21,863 @@
...
+#
+#   IRB.conf[:NEWLINE_BEFORE_MULTILINE_OUTPUT] = false
+#
+# During a session, you can retrieve or set the value using
+# methods <tt>conf.newline_before_multiline_output?</tt>
+# and <tt>conf.newline_before_multiline_output=</tt>.
+#
+# Examples:
+#
+#   rb(main):001:0> conf.inspect_mode = false
```
st0012, line 420: I think this `rb` misses an `i`?

In newer IRB versions we removed the quote indicator, so prompts would not have the last `:0`. Nowadays they look like this instead:

```
irb(main):001>
```

lib/irb.rb
```diff
@@ -21,36 +21,863 @@
...
+#   it also highlights errors such as mismatched parentheses.
+# - Output: \IRB highlights syntactical elements.
+#
+# You can disable color highlighting by:
+#
+# - Adding to the configuration file: <tt>IRB.conf[:USE_COLORIZE] = false</tt>
+#   (the default value is +true+).
+# - Giving command-line option <tt>--nocolorize</tt>
+#
+# == Debugging
```
st0012, line 687: I want to use the debugging section to introduce the [`irb:rdbg` session](https://github.com/ruby/irb#debugging-with-irb) instead.

And I don't think we need to mention the `-d` flag here. I personally have never used it for debug IRB.

lib/irb.rb
```diff
@@ -21,36 +21,863 @@
...
+#
+# Command-line option <tt>-d</tt> sets variables <tt>$VERBOSE</tt>
+# and <tt>$DEBUG</tt> to +true+;
+# these have no effect on \IRB output.
+#
+# === Tracer
```
st0012, line 693: Let's remove anything related to tracer for now as the current integration needs [an upgrade](https://github.com/ruby/irb/issues/752) before being usable again.

lib/irb.rb
```diff
@@ -21,36 +21,863 @@
...
+# inin the same directory as that file).
+#
+# A new \IRB session creates the history file if it does not exist,
+# and appends to the file if it does exist.
+#
+# You can change the filepath by adding to your configuration file:
+# <tt>IRB.conf[:HISTORY_FILE] = _filepath_</tt>,
+# where _filepath_ is a string filepath.
+#
+# During the session, method <tt>conf.history_file</tt> returns the filepath,
```
st0012, line 213: IMO, we should start discouraging using `conf` as it's actually `IRB::Context` instead of the `IRB.conf` hash, which is kinda misleading.
Exposing `Context`'s methods like this also makes future refactoring harder while providing little value since most of its values should be similar to what `IRB.conf` already holds.

BurdetteLamar:
If we're looking to switch to Markdown, let's hold off on reviewing this here.

BurdetteLamar:
I think things are starting to get out of hand here.  My aim in this PR is to revise the stated sections (top through "Encodings"), adding value and avoiding regressions or new errors.  When we've reached that state (which we may have already), I'd like this to be approved and merged.  Otherwise, we can keep proposing further improvements and refinements indefinitely.

I'd like this to be completed in my lifetime (I'll be 81 soon), so let's figure out how to pinch this off and  put it to bed.

Also, I think changing to Markdown is out-of-scope for this PR, as are revisions to the previously-existing sections "Commands" through "Example using IRB Sessions."

st0012:
> My aim in this PR is to revise the stated sections (top through "Encodings"), adding value and avoiding regressions or new errors. 

Sorry that I wasn't aware of this. If that's the case, yeah we can split this into multiple PRs. 

I will go through this again and unless there's any incorrect information, I'll merge this first. Does that sound good to you?
@ruby/irb-reline WDYT?

BurdetteLamar:
> > My aim in this PR is to revise the stated sections (top through "Encodings"), adding value and avoiding regressions or new errors.
> 
> Sorry that I wasn't aware of this. If that's the case, yeah we can split this into multiple PRs.
> 
> I will go through this again and unless there's any incorrect information, I'll merge this first. Does that sound good to you? @ruby/irb-reline WDYT?

I'd have loved to do this much as multiple PRs, but I don't think that works, b/c all of the new here needs to replace all of the target old all at once (cannot be replaced as 1-1 fragments, I didn't think).

That having been done, I'll look forward to reworking the remaining sections (the trailing ones, except the last), and will also be happy to work further on whatever anyone thinks useful.

st0012:
> I'd have loved to do this much as multiple PRs, but I don't think that works, b/c all of the new here needs to replace all of the target old all at once (cannot be replaced as 1-1 fragments, I didn't think).

Sorry that I wasn't clear. By splitting I meant the changes I suggested and the markdown rewrite. I'm aware that this PR's changes are not easy to split 😄

BurdetteLamar:
@st0012, I will endeavor to capture some of your comments elsewhere.  There's no ~~cattle~~ documentor prod per se, but we can put up some Issues.  Please feel free to add doc issues as you like and assign to me.

st0012:
I think some of my comments should still addressed in this PR, like outdated values and example prompts...etc. I've resolved the ones that could be be splitted out. Once they're addressed, I think we can merge this first 👍

lib/irb.rb
st0012: Since IRB can be installed as a gem, you can have it with any supported Ruby version (2.7+ atm).

If you pull this repo directly, runnin `bundle exec irb` will automatically use the latest version of it.
You can also do `gem install irb` to pull the latest release version of it.

BurdetteLamar:
> I think some of my comments should still addressed in this PR, like outdated values and example prompts...etc. I've resolved the ones that could be be splitted out. Once they're addressed, I think we can merge this first 👍

@st0012:  Example prompts updated (can you spell tedious) in new doc, but not in old (sessions stuff, etc.).

Which are the 'outdated values'?

doc/irb/indexes.rdoc
st0012: @BurdetteLamar These should be removed

BurdetteLamar: Removed.

st0012 (approved):
Thanks for the rewrite 🙌 
I'll make a few small changes to hide/hint about deprecated features after merging this.

BurdetteLamar:
> Thanks for the rewrite 🙌 I'll make a few small changes to hide/hint about deprecated features after merging this.

Thanks for your help in reviewing and commenting.

Going forward:
- I'll work on changing the indexes to markdown (good learning for me).
- If we are going to change the main page to markdown, its text *must* be stable across that change;  some agreed-upon time in the future.
- Two things still to look at:  sessions and commands;  I'll see to those.
- Anything else you think needs attention can be put up as documentation Issues?

PR #741, opened 2023-10-30
Use dummy document for document dialog testing

tompng:
Use a dummy document by monkey patching RDoc in irbrc.


tompng:
#743 is much better

PR #744, opened 2023-11-01
Update SourceFinder's find_end algorithm

tompng:
For a very large module, `show_source` was slow.

```ruby
# master branch
IRB::SourceFinder.new(context).find_source('Reline::LineEditor')
processing time: 4.850789s
=> #<struct IRB::SourceFinder::Source file="...", first_line=6, last_line=3300>

# this pull request
IRB::SourceFinder.new(binding).find_source('Reline::LineEditor')
processing time: 0.080596s
=> #<struct IRB::SourceFinder::Source file="...", first_line=6, last_line=3300>
```

## How to find end
```ruby
module M
  class A
    B = C.new([ # Find new open-tokens found in start_line. It's `(` and `[`.
      f(),
      g(),
      h()
    ]
    ) # The line that closes all tokens opened in start_line is the end_line
  end
end
```

## Improvement
Can show `show_source A#foo` of this.
```ruby
class A
  method_name = 'foo'
  module_eval(<<~RUBY, __FILE__, __LINE__ + 1)
    def #{method_name}(args)
      42
    end
  RUBY
end
```
It cannot handled in previous implementation because `"def \#{method_name}(args)\n42"` is an invalid code.

```ruby
# master branch
irb> show_source A#foo
def #{method_name}(args)

# this pull request
irb> show_source A#foo
def #{method_name}(args)
  42
end
```

## Degration
For this code,
```ruby
A.module_eval(<<~METHOD, __FILE__, __LINE__ + 1)
  def f
  end
  def g
  end
METHOD

```

`show_source A#f` will show both f and g
```ruby
# master branch
irb> show_source A#f
def f
end

# this pull request
irb> show_source A#f
def f
end
def g
end
```

### DIFF
--- a/lib/irb/cmd/edit.rb
+++ b/lib/irb/cmd/edit.rb
@@ -31,7 +31,7 @@ module IRB
         if !File.exist?(path)
           source =
             begin
-              SourceFinder.new(@irb_context).find_source(path)
+              SourceFinder.new(@irb_context.workspace.binding).find_source(path)
             rescue NameError
               # if user enters a path that doesn't exist, it'll cause NameError when passed here because find_source would try to evaluate it as well
               # in this case, we should just ignore the error
--- a/lib/irb/cmd/show_source.rb
+++ b/lib/irb/cmd/show_source.rb
@@ -28,7 +28,7 @@ module IRB
           return
         end
 
-        source = SourceFinder.new(@irb_context).find_source(str)
+        source = SourceFinder.new(@irb_context.workspace.binding).find_source(str)
 
         if source
           show_source(source)
--- a/lib/irb/source_finder.rb
+++ b/lib/irb/source_finder.rb
@@ -12,53 +12,52 @@ module IRB
     )
     private_constant :Source
 
-    def initialize(irb_context)
-      @irb_context = irb_context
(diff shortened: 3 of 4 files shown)


st0012:
I like the extracted tests for `SourceFinder` 👍 

> Fails to show multi-line endless method definition

This is a bit concerning because I think it's gradually gaining popularity. 

And it seems that Pry, which uses the [method_source gem](https://github.com/banister/method_source), handle all mentioned cases correctly. I assume its performance won't be great though as it basically [evals the code until syntax error disappears](https://github.com/banister/method_source/blob/master/lib/method_source/code_helpers.rb#L66-L72).

Is it possible we can first cover this case:

```
class A
  method_name = 'foo'
  module_eval(<<~RUBY, __FILE__, __LINE__ + 1)
    def #{method_name}(args)
      42
    end
  RUBY
end
```

And then find an approach to improve performance without regession? 

Personally I'd put accuracy before performance because if I can't use `show_source` to find a method correctly, any alternative is likely take longer than 5 secs.

st0012:
Yeah I was thinking about `RubyVM::AbstractSyntaxTree.parse` too. Thanks for researching on this 👍

PR #754, opened 2023-11-09
Add command line option to select which completor to use

tompng:
Command line option is more easy to try than writing to `.irbrc`
```
irb --type-completion
irb --regexp-completion
```

other possibilities
```
--type-completor
--regexp-completor

--use-type-(completion|completor)
--use-regexp-(completion|completor)
```

or ENV?
```
ENV['IRB_COMPLETOR'] = 'type' | 'regexp'
```

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -330,6 +330,10 @@ module IRB # :nodoc:
         @CONF[:USE_AUTOCOMPLETE] = true
       when "--noautocomplete"
         @CONF[:USE_AUTOCOMPLETE] = false
+      when "--regexp-completor"
+        @CONF[:COMPLETOR] = :regexp
+      when "--type-completor"
+        @CONF[:COMPLETOR] = :type
       when /^--prompt-mode(?:=(.+))?/, /^--prompt(?:=(.+))?/
         opt = $1 || argv.shift
         prompt_mode = opt.upcase.tr("-", "_").intern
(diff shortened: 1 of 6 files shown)


lib/irb/init.rb
```diff
@@ -330,6 +330,10 @@ def IRB.parse_opts(argv: ::ARGV)
         @CONF[:USE_AUTOCOMPLETE] = true
       when "--noautocomplete"
         @CONF[:USE_AUTOCOMPLETE] = false
+      when "--regexp-completion"
+        @CONF[:COMPLETOR] = :regexp
+      when "--type-completion"
+        @CONF[:COMPLETOR] = :type
```
st0012, line 336: Can we have some tests under `InitTest` to cover these?

lib/irb/init.rb
```diff
@@ -330,6 +330,10 @@ def IRB.parse_opts(argv: ::ARGV)
         @CONF[:USE_AUTOCOMPLETE] = true
       when "--noautocomplete"
         @CONF[:USE_AUTOCOMPLETE] = false
+      when "--regexp-completion"
```
st0012, line 333: I'd like to use `completor` instead of `completion` so the term is consistent between flags and config names, which will be easier to remember.

tompng:
renamed and add test

st0012 (approved):

PR #758, opened 2023-11-11
Fix thread leak in test

tompng:
Ruby's leak check will catch thread (and file descriptor reading RBS) leak in `TestIRB::ContextTest#test_build_completor`

This pull request will wait for IRB:TypeCompletion::Types preloading thread to stop.


tompng:
Not needed after #772

PR #759, opened 2023-11-13
Fallback to non-colored inspect when coloring takes too long

osyoyu:
Inspecting large objects may take a long time since IRB colors the entire output of #inspect all at once.
This patch lets IRB to give up coloring after a certain time and fall back to non-colored output.
Changes can be observed by inspecting large Arrays, for example `[1] * 1000000`.

The default timeout is 2 seconds, but this can be configured through `IRB.conf[:INSPECT_COLORING_TIMEOUT]`.
Passing nil to this option will disable timeouts (IRB will never fall back to non-colored output).

### DIFF
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -61,6 +61,8 @@ module IRB
       @io = nil
 
       self.inspect_mode = IRB.conf[:INSPECT_MODE]
+      @inspect_coloring_timeout = IRB.conf[:INSPECT_COLORING_TIMEOUT]
+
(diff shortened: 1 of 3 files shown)


st0012:
Thanks for the PR 👍 
However, it will not be the best solution to the problem because using `timeout` will cause hanging problems with `debug` for the `irb:rdbg` integration. See https://github.com/ruby/debug/issues/877

Maybe we should improve colour printer instead?

tompng:
Thank you for this pull request and also for raising this issue.
The problem is reduced with another approach in #1040, by showing preview of first page quickly.

Issue #751, opened 2023-11-08
Support `hist` command


st0012:
## Description

Support `hist` command to display input history.

### Output format

```
0: foo
1: ls foo
2: some_method
```

And the output should be paged with `Pager.page_content`.

### Command options

At the beginning, supporting only `-g` option should already cover most cases. The implementation will be similar to [ls command's](https://github.com/ruby/irb/blob/master/lib/irb/cmd/ls.rb#L17-L24).

PR #761, opened 2023-11-13
Implement `history` command

garyhtou:
This PR adds the `history` command (aliased as `hist`) which provides a list of previous IRB inputs. This supports multiline inputs but truncates them to two lines. Lines are displayed in reverse to show the most recent inputs first.

This command also supports grep (`-g` or `-G` option); similar to the `ls` command.

image

Closes https://github.com/ruby/irb/issues/751

### DIFF
--- a/lib/irb/cmd/history.rb
+++ b/lib/irb/cmd/history.rb
@@ -23,8 +23,8 @@ module IRB
         formatted_inputs = irb_context.io.class::HISTORY.each_with_index.reverse_each.map do |input, index|
           header = "#{index}: "
 
-          first_line, *other_lines = input.split("\n") || [""]
-          first_line.prepend header
+          first_line, *other_lines = input.split("\n")
+          first_line = "#{header}#{first_line}"
 
           truncated_lines = other_lines.slice!(1..) # Show 1 additional line (2 total)
           other_lines << "..." if truncated_lines&.any?
--- a/lib/irb/cmd/history.rb
+++ b/lib/irb/cmd/history.rb
@@ -13,15 +13,13 @@ module IRB
       description "Shows the input history. `-g [query]` or `-G [query]` allows you to filter the output."
 
       def self.transform_args(args)
-        if match = args&.match(/\A(?<args>.+\s|)(-g|-G)\s+(?<grep>[^\s]+)\s*\n\z/)
-          args = match[:args]
-          "#{args}#{',' unless args.chomp.empty?} grep: /#{match[:grep]}/"
-        else
-          args
-        end
+        match = args&.match(/(-g|-G)\s+(?<grep>[^\s]+)\s*\n\z/)
+        return unless match
+
+        "grep: /#{match[:grep]}/"
       end
 
-      def execute(*arg, grep: nil)
+      def execute(grep: nil)
         formatted_inputs = irb_context.io.class::HISTORY.each_with_index.reverse_each.map do |input, index|
           header = "#{index}: "
 
--- a/lib/irb/cmd/history.rb
+++ b/lib/irb/cmd/history.rb
@@ -35,7 +35,7 @@ module IRB
             " " * header.length + line
           end
 
-          StringIO.new.tap { |io| io.puts(first_line, *other_lines) }.string
+          [first_line, *other_lines].join("\n") + "\n"
         end
 
         Pager.page_content(formatted_inputs.join)
--- a/lib/irb/cmd/history.rb
+++ b/lib/irb/cmd/history.rb
@@ -13,7 +13,7 @@ module IRB
       description "Shows the input history. `-g [query]` or `-G [query]` allows you to filter the output."
 
       def self.transform_args(args)
-        match = args&.match(/(-g|-G)\s+(?<grep>[^\s]+)\s*\n\z/)
+        match = args&.match(/(-g|-G)\s+(?<grep>.+)\s*\n\z/)
         return unless match
 
         "grep: /#{match[:grep]}/"
--- a/lib/irb/cmd/history.rb
+++ b/lib/irb/cmd/history.rb
@@ -16,7 +16,7 @@ module IRB
         match = args&.match(/(-g|-G)\s+(?<grep>.+)\s*\n\z/)
         return unless match
 
-        "grep: /#{match[:grep]}/"
+        "grep: #{Regexp.new(match[:grep]).inspect}"
       end
 
       def execute(grep: nil)
--- a/lib/irb/cmd/history.rb
+++ b/lib/irb/cmd/history.rb
@@ -18,7 +18,7 @@ module IRB
           header = "#{index}: "
           header_length = header.size
 
-          lines = input.split("\n")
+          lines = input.split("\n") || [""]
           first_line = header + lines[0]
 
           truncated_input = lines[1..1]
--- a/lib/irb/cmd/history.rb
+++ b/lib/irb/cmd/history.rb
@@ -10,11 +10,19 @@ module IRB
   module ExtendCommand
     class History < Nop
       category "IRB"
-      description "Show the input history."
+      description "Shows the input history. `-g [query]` or `-G [query]` allows you to filter the output."
 
-      def execute(*)
-        output = StringIO.new
-        irb_context.io.class::HISTORY.each_with_index.reverse_each do |input, index|
+      def self.transform_args(args)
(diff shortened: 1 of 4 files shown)


lib/irb/cmd/history.rb
```diff
@@ -0,0 +1,40 @@
...
+          first_line = header + lines[0]
+
+          truncated_input = lines[1..1]
+          truncated_input << "..." if lines.size > 2
+
+          truncated_input = truncated_input.map do |line|
+            " " * header_length + line
+          end
+
+          output.puts [first_line, *truncated_input].join("\n")
```
tompng, line 31: `puts [a, *b].join("\n")` is almost same as `puts(a, *b)`.
When the second line is `''`, `puts ['first-line', ''].join("\n")` will print 1 line but `puts('first-line', '')` will print 2 lines. I think 2 lines are expected in this case.

lib/irb/cmd/history.rb
```diff
@@ -0,0 +1,40 @@
...
+      description "Show the input history."
+
+      def execute(*)
+        output = StringIO.new
+        irb_context.io.class::HISTORY.each_with_index.reverse_each do |input, index|
+          header = "#{index}: "
+          header_length = header.size
+
+          lines = input.split("\n")
+          first_line = header + lines[0]
```
tompng, line 22: If input is an empty string, `lines[0]` will be `nil`.

In some rare case, HISTORY could contain empty string.
```ruby
$ irb --nomultiline
irb(main):001> Readline.readline '>', true
> (press enter)
=> ""
irb(main):002> irb_context.io.class::HISTORY.to_a.last(2)
=> ["", "irb_context.io.class::HISTORY.to_a.last(2)"]
```
 When this history is once saved to `.irb_history`, HISTORY will always contain empty string. (I have "" in my .irb_history)

st0012:
@garyhtou 👋 I think once you addressed these 2 feedback, we'll be very close to merging this PR and make it part of Ruby 3.3 😉

garyhtou:
Oops, sorry for leaving this hanging. Thanks for the ping!

I'll wrap this up later tonight :)

lib/irb/cmd/history.rb
```diff
@@ -0,0 +1,49 @@
...
+          first_line.prepend header
+
+          truncated_lines = other_lines.slice!(1..) # Show 1 additional line (2 total)
+          other_lines << "..." if truncated_lines&.any?
+
+          other_lines.map! do |line|
+            " " * header.length + line
+          end
+
+          StringIO.new.tap { |io| io.puts(first_line, *other_lines) }.string
```
garyhtou, line 38: This line's a bit ugly. There might be a better write to rewrite it while addressing your comment about `puts a, *b` vs `puts [a, *b].join("\n")`.

In https://github.com/ruby/irb/pull/761/commits/61dc0e54a1179f97952fa3ced66ca52ce4a7035d I rewrote it as `output.puts first_line, *truncated_input`. However, after adding the `-g` option, I lost the need for the `output` StringIO.

lib/irb/cmd/history.rb
garyhtou: Hopefully https://github.com/ruby/irb/pull/761/commits/214fc8acd46976e10aa74a36068342354343e95d handles that issue

garyhtou:
It appears the CI failure might be unrelated

https://github.com/ruby/irb/pull/790

lib/irb/cmd/history.rb
```diff
@@ -0,0 +1,49 @@
...
+      def self.transform_args(args)
+        if match = args&.match(/\A(?<args>.+\s|)(-g|-G)\s+(?<grep>[^\s]+)\s*\n\z/)
+          args = match[:args]
+          "#{args}#{',' unless args.chomp.empty?} grep: /#{match[:grep]}/"
+        else
+          args
+        end
+      end
+
+      def execute(*arg, grep: nil)
```
tompng, line 24: I think it's better to not have `*arg` until implementing arg to history command
```diff
-      def execute(*arg, grep: nil)
+      def execute(grep: nil)
```

lib/irb/cmd/history.rb
tompng: Yeah, it would be better if  it can be simply done without StringIO.

lib/irb/cmd/history.rb
```diff
@@ -0,0 +1,49 @@
...
+        formatted_inputs = irb_context.io.class::HISTORY.each_with_index.reverse_each.map do |input, index|
+          header = "#{index}: "
+
+          first_line, *other_lines = input.split("\n") || [""]
```
tompng, line 28: The `|| [""]` has no effect because `input.split("\n")` will always return an array, (not nil or false)
and if input is `""`, split will return `[]` and `first_line.prepend header` will raise NoMethodError.

One simple idea is
```ruby
first_line = "#{header}#{first_line}"
```
It works even if input is `""` and first_line is `nil`

lib/irb/cmd/history.rb
```diff
@@ -0,0 +1,49 @@
...
+module IRB
+  # :stopdoc:
+
+  module ExtendCommand
+    class History < Nop
+      category "IRB"
+      description "Shows the input history. `-g [query]` or `-G [query]` allows you to filter the output."
+
+      def self.transform_args(args)
+        if match = args&.match(/\A(?<args>.+\s|)(-g|-G)\s+(?<grep>[^\s]+)\s*\n\z/)
```
tompng, line 16: I think we can accept `\s` in regexp. `history -g if match` will be useful. Pry's `hist --grep` accepts spaces.

lib/irb/cmd/history.rb
```diff
@@ -0,0 +1,49 @@
...
+          args = match[:args]
+          "#{args}#{',' unless args.chomp.empty?} grep: /#{match[:grep]}/"
```
tompng, line 18: How about `grep: #{Regexp.new(match[:grep]).inspect}` or `grep: #{match[:grep].dump}` (pass as string) so that grep regexp can include `/`

lib/irb/cmd/history.rb
garyhtou: Good idea!

- Allowing whitespace: https://github.com/ruby/irb/pull/761/commits/c48b70c3c07d184aac8e94394c69621bda23054c
- Removing `args`: https://github.com/ruby/irb/pull/761/commits/6ebd08c142e01c00f5cea67201e1b1c4a4e5da90

lib/irb/cmd/history.rb
garyhtou: Done! https://github.com/ruby/irb/pull/761/commits/e3bafe0655fd0fc95303373e3d1f77864a28337a

lib/irb/cmd/history.rb
garyhtou: Thank you for such a thorough review!

https://github.com/ruby/irb/pull/761/commits/5b64a5be9abe985161aebe92b37af14d4f45b9c6

lib/irb/cmd/history.rb
```diff
@@ -0,0 +1,49 @@
...
+          other_lines << "..." if truncated_lines&.any?
+
+          other_lines.map! do |line|
+            " " * header.length + line
+          end
+
+          StringIO.new.tap { |io| io.puts(first_line, *other_lines) }.string
+        end
+
+        formatted_inputs = formatted_inputs.grep(grep) if grep
```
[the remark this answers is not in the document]
garyhtou: Good catch! I had no idea `filter_map` existed; very useful!

Hopefully this fixes it: https://github.com/ruby/irb/pull/761/commits/9e38cb28949e3b35da89975619ec00098eefb2cf

lib/irb/cmd/history.rb
garyhtou: What are your thoughts on this?

```ruby
formatted = [first_line, *other_lines].join("\n")
formatted += "\n" unless formatted.end_with?("\n")
formatted
```

It's not the cleanest, but should hopefully mimic `puts`'s new line formatting.
> Writes a newline after each that does not already end with a newline sequence.
https://docs.ruby-lang.org/en/master/ARGF.html#method-i-puts

Is there a better way to write this?

garyhtou: Here you go! https://github.com/ruby/irb/pull/761/commits/f2c3d9af8fbb859c40a71dae5f0477f03214c23d

tompng (approved):
Looks great! Thank you :+1:

garyhtou:
Thank you @tompng and @st0012 for helping guide me through this!

PR #763, opened 2023-11-15
require 'irb/version to test completion of constant `IRB::VERSION`

tompng:
because test fails in ruby ci


hsbt:
https://github.com/ruby/ruby/pull/8925 is fine. I'm +1 to merge thsi.

st0012 (approved):

PR #768, opened 2023-11-18
Hint debugger command in `irb:rdbg` session

st0012:
## Background

In https://github.com/ruby/debug/pull/1024#issuecomment-1811928142, @ko1 raised the concern that users may accidentally hit `debug` commands when trying to evaluate variables with certain names. I then talked to a few IRB users during RubyConf and some of them confirmed that this did happen to them before. So I think we should explore methods to avoid such confusion.

## Description

When user enters `irb:rdbg` session, they don't get the same hint that the `debug` gem provides, like

```
(rdbg) n    # next command
```

This means that users may accidentally execute commands when they want to retrieve the value of a variable.

So this commit adds a Reline output modifier to add a simiar hint:

```
irb:rdbg(main):002> n # debug command
```

It is not exactly the same as `debug`'s because in this case the importance is to help users distinguish between value evaluation and debugger command execution.

### DIFF
--- a/lib/irb/debug.rb
+++ b/lib/irb/debug.rb
@@ -57,6 +57,24 @@ module IRB
           DEBUGGER__::ThreadClient.prepend(SkipPathHelperForIRB)
         end
 
+        if !@output_modifier_defined && !DEBUGGER__::CONFIG[:no_hint]
+          irb_output_modifier_proc = Reline.output_modifier_proc
+
+          Reline.output_modifier_proc = proc do |output, complete:|
+            unless output.strip.empty?
+              cmd = output.split(/\s/, 2).first
+
+              if DEBUGGER__.commands.key?(cmd)
+                output = output.sub(/\n$/, " # debug command\n")
+              end
+            end
+
+            irb_output_modifier_proc.call(output, complete: complete)
+          end
+
+          @output_modifier_defined = true
+        end
+
         true
       end
 
--- a/test/irb/yamatanooroti/test_rendering.rb
+++ b/test/irb/yamatanooroti/test_rendering.rb
@@ -379,6 +379,48 @@ class IRB::RenderingTest < Yamatanooroti::TestCase
     assert_match(/foobar/, screen)
   end
 
+  def test_debug_integration_hints_debugger_commands
+    write_irbrc <<~'LINES'
+      IRB.conf[:USE_COLORIZE] = false
+    LINES
+    script = Tempfile.create(["debug", ".rb"])
+    script.write <<~RUBY
+      puts 'start IRB'
+      binding.irb
+    RUBY
+    script.close
+    start_terminal(40, 80, %W{ruby -I#{@pwd}/lib #{script.to_path}}, startup_message: 'start IRB')
+    write("debug\n")
+    write("n")
+    close
+
+    screen = result.join("\n").sub(/\n*\z/, "\n")
+    assert_include(screen, "irb:rdbg(main):002> n # debug command")
+  ensure
+    File.unlink(script) if script
+  end
+
+  def test_debug_integration_doesnt_hint_non_debugger_commands
+    write_irbrc <<~'LINES'
+      IRB.conf[:USE_COLORIZE] = false
+    LINES
+    script = Tempfile.create(["debug", ".rb"])
+    script.write <<~RUBY
+      puts 'start IRB'
+      binding.irb
+    RUBY
+    script.close
+    start_terminal(40, 80, %W{ruby -I#{@pwd}/lib #{script.to_path}}, startup_message: 'start IRB')
+    write("debug\n")
+    write("foo")
+    close
+
+    screen = result.join("\n").sub(/\n*\z/, "\n")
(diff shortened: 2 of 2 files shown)


lib/irb/debug.rb
```diff
@@ -57,6 +57,19 @@ def DEBUGGER__.capture_frames(*args)
           DEBUGGER__::ThreadClient.prepend(SkipPathHelperForIRB)
         end
 
+        Reline.output_modifier_proc = proc do |output, _|
+          next output if output.strip.empty?
+
+          cmd = output.split(/\s/, 2).first
+
+          if DEBUGGER__.commands.keys.grep(/\A#{Regexp.escape(cmd)}/).any?
```
tompng, line 65: `w = 1` is not a command but debug hint is show. 
```
From: a.rb @ line 1 :

 => 1: binding.irb

irb(main):001> debug
irb:rdbg(main):002> w = 1 # debug command
```
How about `DEBUGGER__.commands.key?(cmd)`

lib/irb/debug.rb
```diff
@@ -57,6 +57,19 @@ def DEBUGGER__.capture_frames(*args)
...
+            hint = Color.colorize(" # debug command", [:DIM])
+            output.sub(/\n$/, hint)
+          else
+            output
```
tompng, line 69: Looks like IRB's syntax highlight is disabled here. Can we call IRB's default output_modifier logic here?

st0012:
Good catch on both. I've updated 👍

lib/irb/debug.rb
```diff
@@ -57,6 +57,22 @@ def DEBUGGER__.capture_frames(*args)
...
 
+        unless DEBUGGER__::CONFIG[:no_hint]
+          irb_output_modifier_proc = Reline.output_modifier_proc
+
+          Reline.output_modifier_proc = proc do |output, complete:|
+            unless output.strip.empty?
+              cmd = output.split(/\s/, 2).first
+
+              if DEBUGGER__.commands.key?(cmd)
+                output = output.sub(/\n$/, " # debug command")
```
tompng, line 68: `output = output.sub(/\n$/, " # debug command\n")`

The line count is changed in this operation.
IRB crashes when you type `info[OPTION+ENTER]`
```
irb(main):001> debug
irb:rdbg(main):002> info【OPTION+ENTER】
/Users/tomoya.ishida/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/reline-0.4.0/lib/reline/line_editor.rb:1128:in `+': no implicit conversion of nil into String (TypeError)
```

lib/irb/debug.rb
```diff
@@ -57,6 +57,22 @@ def DEBUGGER__.capture_frames(*args)
           DEBUGGER__::ThreadClient.prepend(SkipPathHelperForIRB)
         end
 
+        unless DEBUGGER__::CONFIG[:no_hint]
+          irb_output_modifier_proc = Reline.output_modifier_proc
```
tompng, line 61: This part is called multiple times (when `continue` hits next `binding.irb`)
```ruby
# a.rb
5.times do
  binding.irb
end
```

```
ruby -I path/to/irb/lib a.rb
irb(main):001> debug
(called)
irb:rdbg(main):002> c # debug command
(called)
irb:rdbg(main):003> c # debug command # debug command
(called)
irb:rdbg(main):004> c # debug command # debug command # debug command
```
And in this case, irb_output_modifier_proc is chained. irb_output_modifier_proc calls previous irb_output_modifier_proc.

I think moving irb's original output_modifier_proc to a method and call it from here is be better.

st0012: I understand that'd be the case, but I actually can't reproduce it 🤔 

```
irb(main):001> debug
irb:rdbg(main):002> c # debug command

From: test.rb @ line 2 :

    1: 5.times do
 => 2:   binding.irb
    3: end

irb:rdbg(main):003> c # debug command

From: test.rb @ line 2 :

    1: 5.times do
 => 2:   binding.irb
    3: end

irb:rdbg(main):004> c # debug command

From: test.rb @ line 2 :

    1: 5.times do
 => 2:   binding.irb
    3: end

irb:rdbg(main):005> c # debug command

From: test.rb @ line 2 :

    1: 5.times do
 => 2:   binding.irb
    3: end

irb:rdbg(main):006>
```

> I think moving irb's original output_modifier_proc to a method and call it from here is better.

Will do 👍

tompng: I forgot to write, it can be reproduced after
```
output = output.sub(/\n$/, " # debug command")
```
is changed to 
```
output = output.sub(/\n$/, " # debug command\n")
```

st0012: I ended up using a flag to check if the `debug` modifier block has been defined or not instead. Calling an input method object's method here feels like a bad coupling.

lib/irb/debug.rb
st0012: Good catch, thank you!

tompng (approved):

PR #770, opened 2023-11-19
adding the -s flag to show_source

### LINKED ISSUE #718: Support `-s` option in `edit` and `show_source` commands
## Description

In `Pry`, users can use `show-source Bar#foo -s` to access `Foo#foo`'s source of the example below:

```rb
class Foo
  def foo
  end
end

class Bar < Foo
  def foo
    super
  end
end
```

This could be extremely helpful when debugging at times and it'd be great to support it in IRB.

paulreece:
This enhancement allows a user to add the -s flag if they want to access a methods origin definition.  It allows for chaining of multiple esses to further go up the classes as needed.

With the following code:
```ruby
class Foo
  def foo
  end
end

class Bar < Foo
  def foo
    super
  end
end
```

If you call `show_source Bar#foo -s` it will return:
```ruby
From: test.rb:4

  def foo
  end

```

Closes #718.

### DIFF
--- a/lib/irb/cmd/show_source.rb
+++ b/lib/irb/cmd/show_source.rb
@@ -27,11 +27,18 @@ module IRB
           puts "Error: Expected a string but got #{str.inspect}"
           return
         end
-
-        source = SourceFinder.new(@irb_context).find_source(str)
+        if str.include? " -s"
+          str, esses = str.split(" -")
+          s_count = esses.count("^s").zero? ? esses.size : 1
+          source = SourceFinder.new(@irb_context).find_source(str, s_count)
+        else
+          source = SourceFinder.new(@irb_context).find_source(str)
+        end
 
         if source
           show_source(source)
+        elsif s_count
+          puts "Error: Couldn't locate a super definition for #{str}"
         else
           puts "Error: Couldn't locate a definition for #{str}"
         end
--- a/lib/irb/source_finder.rb
+++ b/lib/irb/source_finder.rb
@@ -16,7 +16,7 @@ module IRB
       @irb_context = irb_context
     end
 
-    def find_source(signature)
+    def find_source(signature, s_count = nil)
       context_binding = @irb_context.workspace.binding
       case signature
       when /\A[A-Z]\w*(::[A-Z]\w*)*\z/ # Const::Name
@@ -26,14 +26,13 @@ module IRB
       when /\A(?<owner>[A-Z]\w*(::[A-Z]\w*)*)#(?<method>[^ :.]+)\z/ # Class#method
         owner = eval(Regexp.last_match[:owner], context_binding)
         method = Regexp.last_match[:method]
-        if owner.respond_to?(:instance_method)
-          methods = owner.instance_methods + owner.private_instance_methods
-          file, line = owner.instance_method(method).source_location if methods.include?(method.to_sym)
-        end
+        return unless owner.respond_to?(:instance_method)
+        file, line = method_target(owner, s_count, method, "owner")
       when /\A((?<receiver>.+)(\.|::))?(?<method>[^ :.]+)\z/ # method, receiver.method, receiver::method
         receiver = eval(Regexp.last_match[:receiver] || 'self', context_binding)
         method = Regexp.last_match[:method]
-        file, line = receiver.method(method).source_location if receiver.respond_to?(method, true)
+        return unless receiver.respond_to?(method, true)
+        file, line = method_target(receiver, s_count, method, "receiver")
       end
       if file && line && File.exist?(file)
         Source.new(file: file, first_line: line, last_line: find_end(file, line))
@@ -59,6 +58,25 @@ module IRB
         end
       end
       first_line
+    end
+
+    def method_target(owner_receiver, s_count, method, type)
+      case type
+      when "owner"
+        target_method = owner_receiver.instance_method(method)
+        return target_method.source_location unless s_count
+      when "receiver"
+        if s_count
+          target_method = owner_receiver.class.instance_method(method)
+        else
+          target_method = method
+          return owner_receiver.method(method).source_location
+        end
+      end
+      s_count.times do |s|
+        target_method = target_method.super_method if target_method
+      end
+      target_method.nil? ? nil : target_method.source_location
     end
   end
 end
--- a/test/irb/test_cmd.rb
+++ b/test/irb/test_cmd.rb
@@ -469,6 +469,144 @@ module TestIRB
       assert_match(%r[/irb\/init\.rb], out)
     end
 
+    def test_show_source_method_s
+      code = <<~RUBY
+        class Baz
+          def foo
+          end
+        end
+      
+        class Bar < Baz
+          def foo
+            super
+          end
+        end
+      RUBY
+      File.write("#{@tmpdir}/bazbar.rb", code)
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbar.rb'\n",
+        "show_source Bar#foo -s",
+      )
+      assert_match(%r[bazbar.rb:2\n\n  def foo\n  end\n\n=> nil\n], out)
+      File.delete("#{@tmpdir}/bazbar.rb")
+    end
+
+    def test_show_source_method_multiple_s
+      code = <<~RUBY
+        class Baz
+          def fob
+          end
+        end
+      
+        class Bar < Baz
+          def fob
+            super
+          end
+        end
+
+        class Bob < Bar
+          def fob
+            super
+          end
+        end
+      RUBY
+      File.write("#{@tmpdir}/bazbarbob.rb", code)
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbarbob.rb'\n",
+        "show_source Bob#fob -ss",
+      )
+      assert_match(%r[bazbarbob.rb:2\n\n  def fob\n  end\n\n=> nil\n], out)
+      File.delete("#{@tmpdir}/bazbarbob.rb")
+    end
+
+    def test_show_source_method_no_instance_method
+      code = <<~RUBY
+        class Baz
+        end
+      
+        class Bar < Baz
+          def fee
+            super
+          end
+        end
+      RUBY
+      File.write("#{@tmpdir}/bazbar.rb", code)
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbar.rb'\n",
+        "show_source Bar#fee -s",
+      )
+      assert_match(%r[Error: Couldn't locate a super definition for Bar#fee\n], out)
+      File.delete("#{@tmpdir}/bazbar.rb")
+    end
+
+    def test_show_source_method_exceeds_super_chain
+      code = <<~RUBY
+        class Baz
+          def fow
+          end
+        end
+      
+        class Bar < Baz
+          def fow
+            super
+          end
+        end
+      RUBY
+      File.write("#{@tmpdir}/bazbar.rb", code)
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbar.rb'\n",
+        "show_source Bar#fow -ss",
+      )
+      assert_match(%r[Error: Couldn't locate a super definition for Bar#fow\n], out)
+      File.delete("#{@tmpdir}/bazbar.rb")
+    end
+
+    def test_show_source_method_accidental_characters
+      code = <<~RUBY
+        class Baz
+          def fol
+          end
+        end
+      
+        class Bar < Baz
+          def fol
+            super
+          end
+        end
+      RUBY
+      File.write("#{@tmpdir}/bazbar.rb", code)
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbar.rb'\n",
+        "show_source Bar#fol -sddddd",
+      )
+
+      assert_match(%r[bazbar.rb:2\n\n  def fol\n  end\n\n=> nil\n], out)
+      File.delete("#{@tmpdir}/bazbar.rb")
+    end
+
+    def test_show_source_receiver_super
+      code = <<~RUBY
+        class Baz
+          def fot
+          end
+        end
+      
+        class Bar < Baz
+          def fot
+            super
+          end
+        end
+      RUBY
+      File.write("#{@tmpdir}/bazbar.rb", code)
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbar.rb'\n",
+        "bar = Bar.new",
+        "show_source bar.fot -s"
+      )
+      assert_match(%r[bazbar.rb:2\n\n  def fot\n  end\n\n=> nil\n], out)
+      File.delete("#{@tmpdir}/bazbar.rb")
+    end
+
     def test_show_source_string
       out, err = execute_lines(
         "show_source 'IRB.conf'\n",


paulreece:
Sorry for all of the pushes! My test file had a namespace clash with the `Foo` class that was being used by the other tests.  I just changed it to `Baz` instead. :)

test/irb/test_cmd.rb
```diff
@@ -469,6 +469,17 @@ def test_show_source_method
...
     end
 
+    def test_show_source_method_s
+      File.write("#{@tmpdir}/bazbar.rb", "class Baz\n  def foo\n  end\nend\n\nclass Bar < Baz\n  def foo\n    super\n  end\nend\n")
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbar.rb'\n",
+        "show_source Bar#foo -s",
+      )
+      assert_empty err
+      p out
```
st0012, line 479: I think this is a leftover from debugging?

test/irb/test_cmd.rb
```diff
@@ -469,6 +469,17 @@ def test_show_source_method
       assert_match(%r[/irb\/init\.rb], out)
     end
 
+    def test_show_source_method_s
+      File.write("#{@tmpdir}/bazbar.rb", "class Baz\n  def foo\n  end\nend\n\nclass Bar < Baz\n  def foo\n    super\n  end\nend\n")
```
st0012, line 473: Could we use heredoc for this?

test/irb/test_cmd.rb
```diff
@@ -469,6 +469,17 @@ def test_show_source_method
       assert_match(%r[/irb\/init\.rb], out)
     end
 
+    def test_show_source_method_s
```
st0012, line 472: Can we have more tests to cover:

- Multiple super level like `-sss`
- When the level of flags, like `-sss`, exceeds the actual super chain of the method
- When the super owner doesn't support `instance_method` method
- When `-s` is accidentally mixed with other characters, like `-sd`
    - I think in this case it should be seen as `-s`

st0012 (commented):
Thank you for the PR! Just few comments about testing.

test/irb/test_cmd.rb
paulreece: Yes, apologies!

test/irb/test_cmd.rb
paulreece: Yes, apologies!

test/irb/test_cmd.rb
paulreece: These are all great edge cases.  I'll get to work on implementing them today!

paulreece: Okay I added functionality and tests for all of these including `heredoc` for the code.  Let me know what you think!

test/irb/test_cmd.rb
```diff
@@ -469,6 +469,116 @@ def test_show_source_method
...
+            super
+          end
+        end
+      RUBY
+      File.write("#{@tmpdir}/bazbar.rb", code)
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbar.rb'\n",
+        "show_source Bar#foo -s",
+      )
+      assert_match(%r[Error: Couldn't locate a definition for Bar#foo -s\n=> nil\n], out)
```
st0012, line 536: Is it possible to change the message to `Couldn't locate the super definition for Bar#foo`?

lib/irb/source_finder.rb
```diff
@@ -18,6 +18,11 @@ def initialize(irb_context)
 
     def find_source(signature)
       context_binding = @irb_context.workspace.binding
+      if signature.include? " -s"
+        signature, esses = signature.split(" -")
+        esses.count("^s").zero? ? s_count = esses.size : s_count = 1
```
st0012, line 23:
```suggestion
        s_count = esses.count("^s").zero? ? esses.size : 1
```

lib/irb/source_finder.rb
```diff
@@ -26,6 +31,13 @@ def find_source(signature)
       when /\A(?<owner>[A-Z]\w*(::[A-Z]\w*)*)#(?<method>[^ :.]+)\z/ # Class#method
         owner = eval(Regexp.last_match[:owner], context_binding)
         method = Regexp.last_match[:method]
+        if esses
```
st0012, line 34: That's use `s_count` for this condition?

lib/irb/source_finder.rb
```diff
@@ -26,6 +31,13 @@ def find_source(signature)
...
+          super_owner = owner
+          s_count.times {|s| super_owner = super_owner.superclass}
```
tompng, line 36: if there are too many -ssss, `super_owner.superclass` raises ``undefined method `superclass' for nil:NilClass``

lib/irb/source_finder.rb
```diff
@@ -26,6 +31,13 @@ def find_source(signature)
...
+          if super_owner.respond_to?(:instance_method)
```
tompng, line 37: `super_owner` is always `nil` or `Class` here.
The original intention of `if owner.respond_to?(:instance_method)` is to handle a case that owner is not a class.
```ruby
show_source RUBY_VERSION#method
```

How about moving `if esses` logic inside `if owner.respond_to?(:instance_method)` ?
```ruby
if owner.respond_to?(:instance_method)
  if esses
     ...
  end
  methods = owner.instance_methods + ...
end
```

lib/irb/source_finder.rb
paulreece: Much clearer.

lib/irb/source_finder.rb
paulreece: We can use it, it will probably be clearer that way.

lib/irb/source_finder.rb
paulreece: I added this test in there already in the previous commit:
```ruby
    def test_show_source_method_exceeds_super_chain
      code = <<~RUBY
        class Baz
          def foo
          end
        end
      
        class Bar < Baz
          def foo
            super
          end
        end
      RUBY
      File.write("#{@tmpdir}/bazbar.rb", code)
      out, err = execute_lines(
        "irb_load '#{@tmpdir}/bazbar.rb'\n",
        "show_source Bar#foo -ss",
      )
      assert_match(%r[Error: Couldn't locate a definition for Bar#foo -ss\n=> nil\n], out)
    end
```
Does that cover the use case you're referring to? If not I'm happy to add another.

lib/irb/source_finder.rb
paulreece: This is cool context you've given for the conditional.  Right now the `signature` getting passed to the method is a string including the `-s` . For example if I call `show_source Baz#foo -s` in irb.  This method will receive `"Baz#foo -s"` as the `signature` string and error out in the case statement.  That's why I have the parsing logic living at the top there.

Would you suggest that we change the `Regex` in the conditional for this method and do the parsing there? Then we could switch the order of the case statements and resolve it that way.

```ruby
case signature
      when /\A[A-Z]\w*(::[A-Z]\w*)*\z/ # Const::Name
        eval(signature, context_binding) # trigger autoload
        base = context_binding.receiver.yield_self { |r| r.is_a?(Module) ? r : Object }
        file, line = base.const_source_location(signature)
      when /\A(?<owner>[A-Z]\w*(::[A-Z]\w*)*)#(?<method>[^ :.]+)\z/ # Class#method
        owner = eval(Regexp.last_match[:owner], context_binding)
        method = Regexp.last_match[:method]
        if s_count
          super_owner = owner
          s_count.times {|s| super_owner = super_owner.superclass}
          if super_owner.respond_to?(:instance_method)
            owner = super_owner
          end
        end
        if owner.respond_to?(:instance_method)
          methods = owner.instance_methods + owner.private_instance_methods
          file, line = owner.instance_method(method).source_location if methods.include?(method.to_sym)
        end
```

tompng: I was thinking of something like this. Parsing logic should be in top. Not change need for regexp.
```ruby
if owner.respond_to?(:instance_method) && s_count # to ensure owner is a class or module.
  # Maybe we need to check if owner is not a module, or support module too.
  super_owner = owner
  s_count.times do |s|
    # When super_owner reaches BasicObject, its superclass is nil.
    # To avoid calling nil.superclass, add `if super_owner.superclass` or `break unless super_owner` or use `&.`
    super_owner = super_owner.superclass
  end
  # super_owner is class or nil. To distinguish class and nil, respond_to?(:instance_method) is not needed.
  # Perhaps we can just `return unless super_owner` and show `Couldn't locate a definition`
  if super_owner needed.
    owner = super_owner
  end
end
if owner.respond_to?(:instance_method) # This check can be combined with the above one
  methods = owner.instance_methods + owner.private_instance_methods
  file, line = owner.instance_method(method).source_location if methods.include?(method.to_sym)
end
```
After two `if owner.respond_to?(:instance_method)` is combined, it will be like this, which I wanted to say in the previous comment
```ruby
if owner.respond_to?(:instance_method)
  if s_count # it was `if esses` before
     ...
  end
  methods = owner.instance_methods + ...
end
```
Or maybe using early return is more simple
```ruby
return unless owner.respond_to?(:instance_method)
if s_count
  ...
end
methods = ...
```

paulreece: Ah I understand now.  Sorry I misunderstood your original comment but I appreciate the second one for all of the cool ideas you mentioned!

lib/irb/cmd/show_source.rb
```diff
@@ -58,4 +58,4 @@ def bold(str)
       end
     end
   end
-end
+end
```
ima1zumi, line 61: The new line at the end of the file has been removed. Could you add it?

test/irb/test_cmd.rb
paulreece: Just added this.  I moved the parsing to `ShowSource#execute" to accamodate this.

paulreece:
FYI: the `truffleruby` tests seem to be erroring out with some gem dependency errors

lib/irb/cmd/show_source.rb
```diff
@@ -27,11 +27,18 @@ def execute(str = nil)
...
+          s_count = esses.count("^s").zero? ? esses.size : 1
+          source = SourceFinder.new(@irb_context).find_source([str, s_count])
+        else
+          source = SourceFinder.new(@irb_context).find_source(str)
+        end
 
         if source
           show_source(source)
+        elsif s_count
+          puts "Error: Couldn't locate a definition for #{str}"
```
st0012, line 41: Should it be this instead?
```suggestion
          puts "Error: Couldn't locate a super definition for #{str}"
```

paulreece: Hmm, I'm not sure honestly.  Want me to ask on Mastodon and see what feedback I get?

st0012: If we don't want to provide a different message for super definition, why do we need this `elsif` branch then 🤔 The message is exactly the same as the `else` condition's.

paulreece: Lol, fair point!

Well what do you think?  Is it more helpful with the `elsif` branch adding `super` to the Error message or simply removing the `elsif` and having it return as it normally would when it hits the `else`?

st0012: I think mentioning super is necessary because we do find the signatured method's source.

For example, in Pry:

- If it doesn't find the method's source, it prints `Error: Couldn't locate a definition for Foo#baz`
- But if it doesn't find the method's super method, it prints: `Error: No superclass found for #<UnboundMethod: Bar#baz() test.rb:6>`

I think we can tweak the message later, but IMO they have to be different.

paulreece: Makes sense, changing now.

st0012 (approved):
I have a few ideas for small improvements and adopting the same flag to `edit`. But I think it's good to merge now. I'll let @tompng do another pass before pushing the button though.

@paulreece Thank you so much for the work on this PR, as well as all the feedback you provided during RubyConf ❤️

paulreece:
> @paulreece Thank you so much for the work on this PR, as well as all the feedback you provided during RubyConf ❤️

@st0012 you're welcome!  It was a pleasure working with you and @tompng on this PR.  I'm happy to try to address other items from the RubyConf issues or other open bug issues if you'd like.  I really enjoy how constructive working in this repo is. :)

st0012:
@paulreece It's not listed in the issues yet, but I'd really like to make `show_source` command provide more metadata about the method, like the ones from Pry:

```
Visibility: public
Signature: baz()
Number of lines: 3
```

You are definitely welcome to work on any other issues :-) But this one just came up to me today.

paulreece:
>It's not listed in the issues yet, but I'd really like to make `show_source` command provide more metadata about the method, like the ones from Pry:
> 
> ```
> Visibility: public
> Signature: baz()
> Number of lines: 3
> ```
> 


Awesome!  I'll get started on this one!

lib/irb/source_finder.rb
```diff
@@ -18,6 +18,9 @@ def initialize(irb_context)
 
     def find_source(signature)
       context_binding = @irb_context.workspace.binding
+      if signature.kind_of?(Array)
+        signature, s_count = signature
+      end
```
tompng, line 23: Can this be a separate method parameter?
```ruby
def find_source(signature, s_count = nil)
```

paulreece: Done!

lib/irb/source_finder.rb
```diff
@@ -26,10 +26,18 @@ def find_source(signature)
...
+          super_owner = owner
+          s_count.times  do |s| 
+            if super_owner.superclass
+              super_owner = super_owner.superclass
+            end
+          end
+          owner = super_owner
         end
+        methods = owner.instance_methods + owner.private_instance_methods
+        file, line = owner.instance_method(method).source_location if methods.include?(method.to_sym)
```
st0012, line 40: Sorry that I only found this now: I think it's also possible to get the method through `super_method`? Something like:

```rb
target_method = owner.instance_method(method)

s_count.times do |s|
  target_method = target_method.super_method if target_method
end

file, line = target_method.source_location if target_method
```

lib/irb/source_finder.rb
```diff
@@ -26,10 +26,18 @@ def find_source(signature)
...
       when /\A((?<receiver>.+)(\.|::))?(?<method>[^ :.]+)\z/ # method, receiver.method, receiver::method
         receiver = eval(Regexp.last_match[:receiver] || 'self', context_binding)
```
st0012, line 42: If we can use the above approach, we can support these signatures too.

paulreece: Just so I'm understanding this correctly.  We would be combining the two different case statements here when I implement the above approach? Or would I be adding this `target_method` logic into each one?

```ruby
 when /\A(?<owner>[A-Z]\w*(::[A-Z]\w*)*)#(?<method>[^ :.]+)\z/ # Class#method

 when /\A((?<receiver>.+)(\.|::))?(?<method>[^ :.]+)\z/ 
```

st0012: Not combining the case statements. I imagine you can extract the super lookup logic into a private method and use it in both cases instead.

paulreece: Got it

test/irb/test_cmd.rb
```diff
@@ -469,6 +469,116 @@ def test_show_source_method
...
+            super
+          end
+        end
+      RUBY
+      File.write("#{@tmpdir}/bazbar.rb", code)
+      out, err = execute_lines(
+        "irb_load '#{@tmpdir}/bazbar.rb'\n",
+        "show_source Bar#foo -s",
+      )
+      assert_match(%r[Error: Couldn't locate a super definition for Bar#foo\n], out)
```
tompng, line 536: CI (ruby-core setup) is failing in this line.
You can reproduce it by removing `def test_show_source_method_multiple_s` and run `rake test TEST=test/irb/test_cmd.rb` in local.

The problem is that after other test, `Object::Baz` still remains with method `foo` defined.
In this case, `class Baz; end` is not a class definition but an open-class.
The test accidentally passed in other CI because source_location file `bazbarbob.rb` was already deleted.

I think the easiest way to fix this is to add these in all added tests
```ruby
ensure
  Object.remove_const :Bar
  Object.remove_const :Baz
  Object.remove_const :Bob
end
```

lib/irb/source_finder.rb
paulreece: Implemented this and refactored `find_source` a bit as well.  Lmk what you think!

test/irb/test_cmd.rb
paulreece: I ended up just changing the method names in each test since it wouldn't let me access the `remove_const` method now these pass even when commenting out `def test_show_source_method_multiple_s`

TruffleRuby CI is failing because of a gem I believe:
```ruby
In Gemfile:
  debug was resolved to 1.8.0, which depends on
    irb was resolved to 1.9.1, which depends on
      rdoc was resolved to 6.6.0, which depends on
        psych was resolved to 5.1.1.1, which depends on
          stringio
Error: The process '/home/runner/.rubies/truffleruby-head/bin/bundle' failed with exit code 5
```

st0012: (Don't worry about truffleruby for now, we're aware of the problem and have contacted related teams)

tompng (approved):

st0012 (approved):
Thank you so much!

jemmaissroff:
👋 Looks like this PR has introduced test failures in the ruby/ruby master build (https://github.com/ruby/ruby/commits/master). Flagging in case you weren't aware

st0012:
Thanks for flagging. I've just addressed the issues (hopefully) in https://github.com/ruby/irb/pull/793. But I think the commit was not synced to ruby/ruby?

jemmaissroff:
Hmm interesting, it does seem it wasn't synced. Oddly too, it doesn't appear in the #alerts-sync channel in RubyLang slack which usually indicates sync failures. Perhaps we can try syncing it manually?

paulreece:
Sorry about that y'all.  @st0012 Do you have any suggestions for the next PR I open in order to avoid the bug issues that the tests caused on this one?  Maybe I setup the tests a bit differently?

st0012:
Just add new tests to the new file added in #793 should be enough 😄

PR #771, opened 2023-11-19
Enable completion setting through `IRB_COMPLETOR`

ima1zumi:
I propose introducing the capability to set the IRB completion kinds via an environment variable, specifically `IRB_COMPLETOR=type`.
This feature aims to enhance the Rails console experience by allowing Rails users to specify their preferred completion more conveniently.

Currently, when using the Rails console, there's no straightforward way to set the type completion across a Rails application repository.
It's possible to configure this setting by placing a `.irbrc` file at the project root. However, using a `.irbrc` file is not ideal as it allows for broad configurations and can potentially affect the production environment.
My suggestion focuses on allowing users to set the completion to `type` in a minimal.

This enhancement would be particularly beneficial for teams writing RBS in their Rails applications.
This type completor, integrated with RBS, would enhance completion accuracy, improving the Rails console experience.

### DIFF
--- a/test/irb/test_init.rb
+++ b/test/irb/test_init.rb
@@ -120,6 +120,34 @@ module TestIRB
       IRB.conf[:USE_AUTOCOMPLETE] = orig_use_autocomplete_conf
     end
 
+    def test_completor_environment_variable
+      orig_use_autocomplete_env = ENV['IRB_COMPLETOR']
+      orig_use_autocomplete_conf = IRB.conf[:COMPLETOR]
+
+      ENV['IRB_COMPLETOR'] = nil
+      IRB.setup(__FILE__)
+      assert_equal(:regexp, IRB.conf[:COMPLETOR])
+
+      ENV['IRB_COMPLETOR'] = 'regexp'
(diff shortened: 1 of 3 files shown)


tompng (approved):

test/irb/test_init.rb
```diff
@@ -120,6 +120,26 @@ def test_use_autocomplete_environment_variable
...
+      IRB.setup(__FILE__)
+      assert_equal(:regexp, IRB.conf[:COMPLETOR])
+
+      ENV['IRB_COMPLETOR'] = 'regexp'
+      IRB.setup(__FILE__)
+      assert_equal(:regexp, IRB.conf[:COMPLETOR])
+
+      ENV['IRB_COMPLETOR'] = 'type'
+      IRB.setup(__FILE__)
+      assert_equal(:type, IRB.conf[:COMPLETOR])
```
st0012, line 137: Should we add a test case here to check that `--type-completor` flag takes precedence if both are specified?
I think it's already the case but having a test will express that it's the intentional behaviour.

st0012 (approved):

Issue #734, opened 2023-10-18
Better completion using types


tompng:
## Background
Current RegexpCompletor has many limitations. It can't complete chained methods. Many wrong completions.
For better completion, we need to parse the incomplete input and perform type analysis.
We need to use `RBS` and an error tolerant parser like `Prism`.

## PullRequest
WIP pull request is opend in #708.
Fallbacks to RegexpCompletor if Prism is not available.

## Performance

RBS loading performance depends on `rbs_collection.yml` and project's rbs size under `sig/` dir. It takes 0.168721s when there is no yml and sig.
Completion performane depends on code size editing in irb. For example below, it only analyze `"n.times do\n  _1."` with binding that knows the actual value of `n`.
```ruby
irb(main):123> n.times do
irb(main):124>   _1.█
```
Analyzing 1161 lines takes 0.072726s. For comparision, coloring code takes 0.057202s
```ruby
code = File.read 'lib/irb/type_completion/type_analyzer.rb'
# Force to analyze every single nodes
class IRB::TypeCompletion::TypeAnalyzer::DigTarget; def dig?(node)=true; end
measure
IRB::TypeCompletion::Completor.new.completion_candidates code, '.', '', bind: binding;
# => processing time: 0.072726s
IRB::Color.colorize_code(code);
# => processing time: 0.057202s
```

### Comparision with current RegexpCompletor

As far as I know, the new completor can complete every case that RegexpCompletor can complete, even if RBS is not available.
```ruby
# Completion without RBS that RegexpCompletor can't complete

def foo(n = 0)
  a = n; a = 1.0 if b; a.█ # Integer and Float methods
end

class User < ApplicationRecord
  has_█ # has_many, has_one
  def bar
    sa█ # save, save!
  end
end

module IRB
  A█ # IRB.constants and Object.constants
  @@a█ # IRB's class variable names
end
```

### With RBS

When RBS is available, it can complete more.
```ruby
# Completion with RBS

'a'.bytes.shuffle.join.encoding.█ # Encoding methods

10.times.map(&:to_s).map { _1.to_sym }.map(&:█) # Symbol methods

object.instance_eval do
  @a█ # object's instance method names
  self.█ # object's methods
end

%w[a b c].each_with_index do
  s = [:s] * _1
  a = [:a] * _2
  s[0].█ # String methods
  a[0].█ # Symbol methods
end
```

### In Rails apps
[...]

PR #772, opened 2023-11-20
Split TypeCompletion to a gem

tompng:
## Description

TypeCompletion requires Prism and RBS. Since IRB is a default gem and IRB supports ruby>=2.7.0, we can't write Prism dependency(ruby>=3.0) to gemspec.

Splitting to gem allows to simplify the version checks and other. (Other problems are described in https://github.com/ruby/irb/issues/734#issuecomment-1801853235)

## API of repl_type_completor
https://github.com/ruby/repl_type_completor/blob/main/sig/repl_type_completor.rbs
```rbs
module ReplTypeCompletor
  VERSION: String

  class Result
    def completion_candidates: () -> Array[String]
    def doc_namespace: (String) -> String?
  end

  def self.analyze: (String, binding: Binding, ?filename: String?) -> Result?

  def self.rbs_load_error: () -> Exception?
  def self.last_completion_error: () -> Exception?

  def self.rbs_load_started?: () -> bool
  def self.rbs_loaded?: () -> bool
  def self.load_rbs: () -> void
  def self.preload_rbs: () -> void
  def self.info: () -> String
end
```

### Future extending plans
The `completion_candidates` and `doc_namespace` is insufficient when we want to
- Show a hint in completion dialog (example: `abort (method)` `alias (keyword)` `and (keyword)` `at_exit (method)` `a_local_var (lvar)`)
- Show something in document dialog (example: RBS signature) when there is no document
- Add another dialog to show something (example: possible types of receiver in method call)

We can add a method to `ReplCompletion::Result` to extend.

```ruby
# These are just an idea

result.categorized_candidates
#=>
[
  { type: :lvar, candidates: ['abc, 'abcde'] },
  { type: :keyword, candidates: ['and', 'alias'] },
  { type: :call, candidates: ['abort', 'at_exit'], receiver: Kernel },
  { type: :call, candidates: ['aaa', 'aab'], receiver: main.singleton_class },
]

result.match_info(matched)
# =>
{
  type: :lvar,
  value_classes: [Array, Hash],
  value_type_inspect: 'Array[String] | Hash[Symbol, String | Integer]',
  other_useful_information: something
}
```

### Duplicate implementation with IRB
Gem repl_type_completor has a reimplementation of `IRB::InputCompletor#retrieve_files_to_require_from_load_path` and `IRB::InputCompletor#retrieve_files_to_require_relative_from_current_dir`.
It is duplicated but I think it should be calculated in the separated gem.

Also IRB can't delete these methods until dropping RegexpCompletor.

### `filename:` option
We need source file path to correctly complete require_relative paths.
`require_relative` will load relative path from `__dir__`, not from `Dir.pwd`.
`IRB.CurrentContext.irb_path` is passed to `eval(string, binding, filename, lineno)` and completor needs this value.

```ruby
# Using binding.source_locaiton[0] as a filename is not sufficient because:
irb(main):001> IRB.CurrentContext.workspace.binding.source_location[0]
=> "/path/to/irb/lib/irb/workspace.rb" # wrong base path for require_relative
irb(main):002> __FILE__
=> "(irb)" # In this case, looks like `Dir.pwd` is used instead
```


lib/irb/context.rb
```diff
@@ -176,26 +176,19 @@ def initialize(irb, workspace = nil, input_method = nil)
...
-        warn 'TypeCompletion requires RUBY_VERSION >= 3.0.0'
+      if RUBY_ENGINE == 'truffleruby'
+        # Avoid SynatxError. truffleruby does not support endless method definition yet.
+        warn 'TypeCompletor is not supported on TruffleRuby yet'
         return
       end
+
       begin
-        require 'prism'
+        TypeCompletor.new(irb_path)
```
st0012, line 187: Can we move `require 'repl_type_completor'` here? It will help readers to understand what the `LoadError` rescue is primarily for.

test/irb/test_context.rb
```diff
@@ -668,7 +668,16 @@ def test_build_completor
...
+        begin
+          require 'repl_type_completor'
+          type_completor_available = true
+        rescue LoadError
+        end
+        if type_completor_available
+          # Load RBS before testing to avoid completor preload RBS in a thread
+          ReplTypeCompletor.load_rbs unless ReplTypeCompletor.rbs_loaded?
+          assert_equal 'IRB::TypeCompletor', @context.send(:build_completor).class.name
+        end
```
st0012, line 680: I feel we can move this to `test_type_completor.rb` and make it an integration test. This way we won't need the condition.

lib/irb/completion.rb
```diff
@@ -93,6 +93,29 @@ def retrieve_files_to_require_relative_from_current_dir
     end
   end
 
+  class TypeCompletor < BaseCompletor # :nodoc:
+    def initialize(source_file = nil)
+      require 'repl_type_completor'
+      ReplTypeCompletor.preload_rbs
+      @source_file = source_file
```
st0012, line 100: Does this mean in the entire IRB session, user only gets the completion targeting one source file? For example, if I put a `binding.irb` under `app/models/user.rb`, will it run all analysis only around `user.rb`?

tompng: Nice catch. Actually, this initialization  (build_type_completor and also `IRB::Irb.new`) will be called in each `binding.irb`.
But I found that `@irb_path` is not fully initialized yet when build_completor is called and `context.irb_path` is overwritten after `IRB::Irb` initialization.
```ruby
# In irb.rb:1050
binding_irb = IRB::Irb.new(workspace)
binding_irb.context.irb_path = irb_path
binding_irb.run(IRB.conf)
```

I will change to `IRB::TypeCompletor.new(context)` and call `ReplTypeCompletor.analyze(preposing + target, binding: bind, filename: context.irb_path)` on completion phase.

test/irb/test_context.rb
tompng: Sounds nice, moved.
Changed to test the `irb_info` result and also test if `@completor.completion_candidates` is really working

lib/irb/completion.rb
st0012: Yeah I think that's better because with `irb:rdbg`, `irb_path` could be updated if user moves around with the debugger. If we store it as a state in InputMethod, it won't be synced correctly.

Rakefile
```diff
@@ -13,7 +13,7 @@ desc "Run each irb test file in isolation."
 task :test_in_isolation do
   failed = false
 
-  FileList["test/irb/test_*.rb", "test/irb/type_completion/test_*.rb"].each do |test_file|
+  FileList["test/irb/**/test_*.rb"].each do |test_file|
```
st0012, line 16: Ah I forgot to change this in #794 🤦‍♂️

st0012 (approved):

Issue #779, opened 2023-11-22
Encoding::ConverterNotFoundError: ASCII-8BIT to UTF-8


taketo1113:
## Description

It raises multiple error `Encoding::ConverterNotFoundError` and actual exception of `SQLite3::CantOpenException`.

This repository is an example of reproduction.
https://github.com/taketo1113/irb-test

Error messages:
```
$ bundle exec irb
irb(main):001> require './test.rb'
irb(main):002> Test.new.exec
1
2
3
...
246
247
248
bundler: failed to load command: irb (/Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/bin/irb)
/Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:678:in `initialize': code converter not found (ASCII-8BIT to UTF-8) (Encoding::ConverterNotFoundError)
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:678:in `new'
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:678:in `encode_with_invalid_byte_sequence'
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:723:in `handle_exception'
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:529:in `rescue in block (2 levels) in eval_input'
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:509:in `block (2 levels) in eval_input'
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:823:in `signal_status'
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:508:in `block in eval_input'
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:590:in `block in each_top_level_statement'
	from /Users/taketo/irb-test/tmp/bundle/ruby/3.2.0/gems/irb-1.9.1/lib/irb.rb:585:in `loop'
[...]

PR #780, opened 2023-11-22
Handle handle_exception's exception

### LINKED ISSUE #233: Ctrl+c can exit irb while in handle_exception phase of REPL.
## Description

if a very lengthy pretty-printing is being displayed, then during at least some portion of that time, Ctrl+c will cause irb to exit, losing all progress in whatever the user was doing, whereas the user would expect Ctrl+c to be trapped as usual.

## Result of irb_info

Ruby version: 3.0.1
IRB version: irb 1.3.5 (2021-04-03)
InputMethod: ReidlineInputMethod with Reline 0.2.5

## Terminal Emulator

MacOS Terminal.app

## Setting Files

no

tompng:
Fix #233 and #779

Fix IRB crash when exception occurred inside handle_exception.

- Ctrl+C is pressed while printing exception
- Exception's encoding can't be converted
- Exception's method message, full_message, inspect raise error

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -743,7 +743,14 @@ module IRB
         message = message.gsub(/\(irb\):(?<num>\d+):in `<(?<frame>top \(required\))>'/)  { "(irb):#{$~[:num]}:in `<main>'" }
         puts message
       end
-      print "Maybe IRB bug!\n" if irb_bug
+      puts 'Maybe IRB bug!' if irb_bug
+    rescue Exception => handler_exc
+      begin
+        puts exc.inspect
+        puts "backtraces are hidden because #{handler_exc} was raised when processing them"
+      rescue Exception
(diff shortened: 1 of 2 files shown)


lib/irb.rb
```diff
@@ -744,6 +744,12 @@ def handle_exception(exc)
         puts message
       end
       print "Maybe IRB bug!\n" if irb_bug
+    rescue Exception
+      begin
+        print "#{exc.inspect}\n"
```
st0012, line 749: I think we should hint users that something happened in the `handle_exception`, like:

```suggestion
    rescue Exception => handler_exc
      begin
        puts exc.inspect
        puts "backtraces are hidden because #{handler_exc} was raised when processing them"
```

tompng: 👍  Updated

st0012 (approved):

PR #781, opened 2023-11-22
Fix failure of more command with -R option

hogelog:
I believe the more command generally does not accept the raw option (-R).
As an example, the more command man included in util-linux shown below does not have the -R option.
https://man7.org/linux/man-pages/man1/more.1.html
When I run pager with irb in an environment where the -less command does not exist, I get the following error

```console
root@12afc3afe8da:/app# irb
irb(main):001> ls File
more: invalid option -- 'R'
Try 'more --help' for more information.
=> nil
irb(main):002> 
```

After installing the irb gem with this Pulll Request modification, I was able to successfully execute pagenation with more.

### DIFF
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -62,7 +62,7 @@ module IRB
           pager = Shellwords.split(pager)
           next if pager.empty?
 
-          if pager.first == 'less' || pager.first == 'more'
+          if pager.first == 'less'
             pager << '-R' unless pager.include?('-R')
           end
 


hogelog:
The test is failing. Seems like perhaps a fix in @tompng's <https://github.com/tompng/irb/tree/prism_match_write_case_match> branch would fix it.

hogelog:
Rebased onto < https://github.com/ruby/irb/commit/943c14b12ed6c06f6785c2a7e8abc55e6532b322> and force pushed to pass test.

tompng (approved):
Thank you for the pull request.

In my macOS, `more` command accepted `-R` option because it was actually `less` (!)
```
$ more --version
less 581.2 (POSIX regular expressions)
Copyright (C) 1984-2021  Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Home page: https://greenwoodsoftware.com/less
```

hogelog:
Thank you for merging this!

> In my macOS, `more` command accepted `-R` option because it was actually `less` (!)

Yes, exactly what I was confused about at first too😅

PR #782, opened 2023-11-23
Fix exception(backtrace=nil) prints nothing

tompng:
When an exception that has backtrace=nil is raised, IRB prints nothing.

```
irb(main):001> raise StandardError.new('error').tap{def _1.backtrace()=nil}
irb(main):002> raise _
irb(main):003> _
=> #<StandardError: error>
irb(main):004> 
```

I removed `if exc.backtrace` condition. [files?w=1](https://github.com/ruby/irb/pull/782/files?w=1)
Any error on printing error with backtrace will be rescued after #780.

The original test properly ensures stderr to be empty, but it was not properly testing stdout.
```ruby
assert_in_out_err(bundle_exec + %w[-rirb -W0 -e IRB.start(__FILE__) -- -f --], <<-IRB, /Exception: foo/, [])
  e = Exception.new("foo")
  puts e.inspect
  def e.backtrace; nil; end
  raise e
IRB
```
```ruby
e = Exception.new("foo")  # prints "=> #<Exception: foo>" that makes test pass
puts e.inspect            # prints "#<Exception: foo>" and "nil" that also makes test pass
def e.backtrace; nil; end # prints "=> :backtrace"
raise e                   # should print some error message and this should be tested but it doesn't
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -697,7 +697,7 @@ module IRB
     end
 
     def handle_exception(exc)
-      if exc.backtrace && exc.backtrace[0] =~ /\/irb(2)?(\/.*|-.*|\.rb)?:/ && exc.class.to_s !~ /^IRB/ &&
+      if exc.backtrace[0] =~ /\/irb(2)?(\/.*|-.*|\.rb)?:/ && exc.class.to_s !~ /^IRB/ &&
          !(SyntaxError === exc) && !(EncodingError === exc)
         # The backtrace of invalid encoding hash (ex. {"\xAE": 1}) raises EncodingError without lineno.
         irb_bug = true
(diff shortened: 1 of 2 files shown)


lib/irb.rb
```diff
@@ -697,52 +697,49 @@ def encode_with_invalid_byte_sequence(str, enc)
     end
 
     def handle_exception(exc)
-      if exc.backtrace && exc.backtrace[0] =~ /\/irb(2)?(\/.*|-.*|\.rb)?:/ && exc.class.to_s !~ /^IRB/ &&
```
st0012, line 700: I don't think we should change this. With the same example, it'll cause NoMethodError:

```
irb(main):001> raise StandardError.new('error').tap{def _1.backtrace()=nil}
#<StandardError: error>
backtraces are hidden because undefined method `[]' for nil:NilClass was raised when processing them
irb(main):002> 
```

tompng: Yes, it will cause NoMethodError, and I think it's OK because it will be rescued and prints something.

This is the original issue https://bugs.ruby-lang.org/issues/9063 and pull-request https://github.com/ruby/ruby/pull/434.
Nil backtrace was a bug of ActionView::Template::Error, but backtrace check was introduced because
> irb shouldn't die even if caught Exception is something wrong (sorah wrote in bugs.ruby-lang.org)

`raise Exception.new.tap{def _1.message()=nil}` will raise NoMethodError.
`def _1.full_message(*)=nil` `def _1.full_message()='forgot to write arg'` also raise error.
I think nil backtrace is not a common case that IRB should specially handle.

st0012: Fair enough 👍

st0012 (approved):

PR #783, opened 2023-11-24
Support disabling pager

st0012:
With either `IRB.conf[:NO_PAGER] = true` or `--no-pager` commnad line flag.

I decided use `--no-pager` instead of `--use-pager` because it matches with Pry and git's command line flags.

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -82,6 +82,7 @@ module IRB # :nodoc:
     @CONF[:USE_LOADER] = false
     @CONF[:IGNORE_SIGINT] = true
     @CONF[:IGNORE_EOF] = false
+    @CONF[:USE_PAGER] = true
     @CONF[:EXTRA_DOC_DIRS] = []
     @CONF[:ECHO] = nil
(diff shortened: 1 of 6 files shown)


tompng:
The same option with `git` and `pry` seems good. However, IRB's other `--no` prefixed options are `--noecho` `--noinspect` `--nomultiline`. I think `--nopager` is more consistent.

Some other IRB's config use `USE_` prefixes.
How about default `IRB.conf[:USE_PAGER] = true(default)` and command line option (--nopager or --no-pager) turns it to false?

st0012:
> I think --nopager is more consistent.

IMO users are more likely to follow the convention from other tools, instead of trying to derive it from existing IRB flags. And for this reason I feel the existing `--no[name]` convention is not desirable as it creates unnecessary friction for users.

> How about IRB.conf[:USE_PAGER] = true(default)

👍

tompng (approved):

heaven:
I hate this so much, both the multiline and the pager.

Issue #495, opened 2023-01-05
Add a pager to evaluation result


nickjer:
## Description

Currently when IRB outputs something with a large amount of data such as `MyRecord.all` in Rails it will pause for a long time and then dump everything to the screen at once. This can be a bit of a pain when coming from Pry where it comes with a [pager](https://github.com/pry/pry/blob/master/lib/pry/pager.rb) that will instantly return and allow the user to navigate around in the data.

You can reproduce this wait with...

```ruby
> [1] * 1_000_000
```

I was wondering if IRB could add a pager similar to Pry. If so, I'd be happy to explore it but also don't want to step on any toes if this is a no go.

st0012:
I gave it a try but unfortunately simply paging the result won’t solve the issue. The slowness comes from IRB using `ColorPrinter.pp` to generate the inspected value string all at once. And value inspection behavior is customisable in IRB so it’s not something we can easily change.

That being said, paging the generated long string itself doesn’t seem difficult to achieve. Do you think if we implement only that part for now will already help? Or the main concern is just the rendering speed?

nickjer:
That is unfortunate. The rendering speed is definitely the main issue, but there is still significant value in the pager.

For instance, I regularly use the pager in Pry to find text or scroll back-and-forth using the vim keybindings. Basically it makes navigating large chunks of output so much easier. That being said, if it makes its way into IRB that would be amazing.

> And value inspection behavior is customisable in IRB so it’s not something we can easily change.

So when you say it is customizable, does that mean it can be disabled through configuration (e.g., like an environment variable)?

st0012:
Sorry for the long delay.

> So when you say it is customizable, does that mean it can be disabled through configuration (e.g., like an environment variable)?

Yes, you can configure it with `IRB.conf[:INSPECT_MODE]` in your `.irbrc`. Here's a list of [default inspectors](https://github.com/ruby/irb/blob/master/lib/irb/inspector.rb#L111-L129) in IRB. But you can also define your own with the same APIs.

I think to dramatically improve the speed in your example input, setting it to `IRB.conf[:INSPECT_MODE] = :to_s # or :raw or false` should do the trick.

osyoyu:
What do you think about adding a timeout to the coloring process? I've sent a patch for that in https://github.com/ruby/irb/pull/759 .

[...]

PR #784, opened 2023-11-24
Page evaluation result's output

st0012:
This will make it easier to work with long output that exceeds the terminal's height.
Partially solving the problem mentioned in https://github.com/ruby/irb/issues/495

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -20,6 +20,7 @@ require_relative "irb/color"
 require_relative "irb/version"
 require_relative "irb/easter-egg"
 require_relative "irb/debug"
+require_relative "irb/pager"
 
 # IRB stands for "interactive Ruby" and is a tool to interactively execute Ruby
 # expressions read from the standard input.
@@ -859,11 +860,12 @@ module IRB
           end
         end
       end
+
       if multiline_p && @context.newline_before_multiline_output?
-        printf @context.return_format, "\n#{str}"
-      else
-        printf @context.return_format, str
+        str = "\n" + str
       end
+
+      Pager.page_content(format(@context.return_format, str), retain_content: true)
     end
 
     # Outputs the local variables to this current session, including
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -7,9 +7,9 @@ module IRB
     PAGE_COMMANDS = [ENV['RI_PAGER'], ENV['PAGER'], 'less', 'more'].compact.uniq
 
     class << self
-      def page_content(content)
+      def page_content(content, **options)
         if content_exceeds_screen_height?(content)
-          page do |io|
+          page(**options) do |io|
             io.puts content
           end
         else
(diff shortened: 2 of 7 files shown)


lib/irb.rb
```diff
@@ -859,11 +860,12 @@ def output_value(omit = false) # :nodoc:
...
       end
+
       if multiline_p && @context.newline_before_multiline_output?
-        printf @context.return_format, "\n#{str}"
-      else
-        printf @context.return_format, str
+        str = "\n" + str
       end
+
+      Pager.page_content(format(@context.return_format, str))
```
tompng, line 868: After quit `less`, previous output is disappeared.
```ruby
irb(main):001> 1000.times.to_a
irb(main):002> _.size
=> 1000
```
I think this disappearing feature is useful for some command's pager output but not for inspect output.

Pry uses `-R` `-X` and `-F` option

```
$ less --help
  -F  ........  --quit-if-one-screen
                  Quit if entire file fits on first screen.
  -X  ........  --no-init
                  Don't use termcap init/deinit strings.
```

st0012: Yeah I initially thought it's good to not show the paged long output as it could push up the history a lot. But in the case of smaller terminal, that could be annoying instead.

But I'm not sure if we should use `-X` to implement the behaviour. If we were to support remote IRB, we need to centralise the output to a specific IO object, which may not be stdout.

Will push my alternative idea shortly.

lib/irb/pager.rb
```diff
@@ -7,18 +7,21 @@ class Pager
...
         else
           $stdout.puts content
         end
       end
 
-      def page
+      def page(retain_content: false)
         if IRB.conf[:USE_PAGER] && STDIN.tty? && pager = setup_pager
+          # if retain_content is true, we want to print the content to the designated IO too, which is $stdout for now
+          yield $stdout if retain_content
```
tompng, line 23: It looks like `more` behaves similar to `less -X -F`
`echo hello | more` exits just after printing.
`ruby -e "puts 100.times.to_a" | more` and press `q`, output remains.

When user uses `more`, print and show pager will double-print the output.

st0012: I see. I can't find `more` on macOS so I can't test it 😞 
But given this I decided to implement content retaining  with `-X` flag for now as we're already appending the `-R` flag. I didn't add `-F` flag though as I think the current pager already does the size check itself?

st0012: It's weird that on my machine the `-R -X` does exactly what I expect. But on CI the same command doesn't preserve the output after quitting pager:

```
Failure: test_long_evaluation_output_is_preserved_after_paging(IRB::RenderingTest::TestVTerm):
  </(a{80}\n){6}/> was expected to be =~
  <"start IRB\n" +
  "irb(main):001> 'a' * 80 * 11\n" +
  "Paging content with command: [\"less\", \"-R\", \"-X\"]\n" +
  "WARNING: terminal is not fully functional\n" +
  "Press RETURN to continue\n" +
  "irb(main):002> 'foo' + 'bar'\n" +
  "=> \"foobar\"\n" +
  "irb(main):003>\n">.
```

st0012: To get consistent result on both local and CI environment, I think we need to start setting `TERM` in rendering tests. And since we want to make these tests as close to most users' setup as possible, maybe it's better to set it to something like `xterm-256color` instead of `dumb`?

tompng (approved):

Issue #775, opened 2023-11-20
basic usability


zenspider:
I'm having so many problems with irb. Basic usability issues. For example, if your config includes:

```
IRB.conf[:PROMPT_MODE] = :SIMPLE
```

and you type `help`, you literally can't tell that you've just entered `ri` mode. 

- [x] `help` should help. 

- [x] `ri` should `ri`. 
`show_doc` just shouldn't be.

- [x] `cwws` and `chws` are _terribly_ named. 
Consider `cd` and... I dunno, `pwd`?

- [x] `chws` doesn't accept `..` and I can't figure out how to get out.

`chws` doesn't accept `-h`. 

- [x] `show_cmds` doesn't detail anything. There's no way to see help for a command.

- [x] `show_cmds` doesn't show any of the aliases.

- [x] `show_source` doesn't seem to be able to find ANY source for anything defined in the irb session.

That's 10 minutes worth of investigation prompted by @st0012 saying this thing was ready and should go in our Gemfiles. _I just don't think it is ready_. I used to LOVE irb and resisted using pry for years, but ever since the irb overhaul, I have found it to be very unusable.

Please sit down with someone unfamiliar with `irb` and watch them fire it up and try to use it. **Every** friction point, **every** confusion, **every** broken expectation should be written up as an issue, prioritized, and fixed.

st0012:
Except that there's no plan (nor against) to accept `-h` argument for commands (as users can already do `help [cmd]`), I think all other problems mentioned here have been addressed. Thanks again for the feedback 🙂 
Please open new issues if you have other ideas 😄

PR #788, opened 2023-11-25
Display aliases in help message

st0012:
Similar to Pry, it displays aliases in the help message with a dedicated section. With the current default aliases, it looks like:

```
...other sections...

Aliases
  $              Alias for `show_source`
  @              Alias for `whereami`
```

Addresses a point in #775 

> show_cmds doesn't show any of the aliases.

### DIFF
--- a/lib/irb/cmd/show_cmds.rb
+++ b/lib/irb/cmd/show_cmds.rb
@@ -16,6 +16,12 @@ module IRB
         commands_info = IRB::ExtendCommandBundle.all_commands_info
         commands_grouped_by_categories = commands_info.group_by { |cmd| cmd[:category] }
 
+        user_aliases = irb_context.instance_variable_get(:@user_aliases)
+
+        commands_grouped_by_categories["Aliases"] = user_aliases.map do |alias_name, target|
+          { display_name: alias_name, description: "Alias for `#{target}`" }
+        end
+
         if irb_context.with_debugger
           # Remove the original "Debugging" category
           commands_grouped_by_categories.delete("Debugging")
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -146,8 +146,20 @@ module IRB
         @newline_before_multiline_output = true
       end
 
-      @command_aliases = IRB.conf[:COMMAND_ALIASES]
+      @user_aliases = IRB.conf[:COMMAND_ALIASES].dup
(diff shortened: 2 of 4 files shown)


lib/irb/cmd/show_cmds.rb
```diff
@@ -16,6 +16,12 @@ def execute(*args)
         commands_info = IRB::ExtendCommandBundle.all_commands_info
         commands_grouped_by_categories = commands_info.group_by { |cmd| cmd[:category] }
 
+        user_aliases = irb_context.instance_variable_get(:@user_aliases)
```
st0012, line 19: Since it's purely internal and only for this command at the moment, I don't plan to expose an API for it.

And I don't directly read `IRB.conf[:COMMAND_ALIASES]` here because if the user updates it in the session, help message and the actual behaviour will be out of sync.

lib/irb/context.rb
```diff
@@ -146,9 +146,21 @@ def initialize(irb, workspace = nil, input_method = nil)
         @newline_before_multiline_output = true
       end
 
-      @command_aliases = IRB.conf[:COMMAND_ALIASES]
+      @user_aliases = IRB.conf[:COMMAND_ALIASES]
+      @command_aliases = @user_aliases.merge(KEYWORD_ALIASES)
```
tompng, line 150: This used to work, but it does not now.
```ruby
irb(main):001> IRB.conf[:COMMAND_ALIASES][:ss] = :show_source
irb(main):002> ss Set#include?
```
It does not work when you write it directly into IRB console, but show_cmds shows not-working aliases.
```
Aliases
  $              Alias for `show_source`
  @              Alias for `whereami`
  ss             Alias for `show_source`
```
The problem is that when `IRB.conf[:COMMAND_ALIASES]` change, `@user_aliases` will change (because referencing same Hash instance) but `@command_aliases` does not change.

I think one of these is better
```ruby
# Use a snapshot of COMMAND_ALIESES.
# Both `@user_alias` and `@command_aliases` does not change when `IRB.conf[:COMMAND_ALIASES]` change.
@user_aliases = IRB.conf[:COMMAND_ALIASES].dup
```

```ruby
# Both `@user_alias` and `command_aliases` change when `IRB.conf[:COMMAND_ALIAS]` change.
def command_aliases() = @user_aliases.merge(KEYWORD_ALIASES)
```

tompng (approved):

PR #790, opened 2023-11-27
Mitigate truffleruby failure by using unreleased stringio gem

st0012:
The fix has been merged but not released yet: https://github.com/ruby/stringio/pull/71

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,3 +22,6 @@ if RUBY_VERSION >= "3.0.0"
   gem "rbs"
   gem "prism", ">= 0.18.0"
 end
+
+# Remove this after stringio 3.1.0 is released
+gem "stringio", github: "ruby/stringio"


eregon:
Shouldn't be necessary anymore (stringio 3.1.0 was released)

eregon:
But unfortunately there seems to be some not updated cache or so yet in rubygems.org :/
https://github.com/ruby/stringio/pull/71#issuecomment-1829565743

eregon:
The failure on https://github.com/ruby/irb/actions/runs/7017641177/job/19091354167?pr=790#step:3:87 is due to the debug gem using RBASIC()->flags too.
I'll make a PR.

eregon:
https://github.com/ruby/debug/pull/1039

st0012:
@eregon Thanks for the context and the PR on `debug` 🙏 
Since the goal is to test compilation against TruffleRuby directly in `debug` repo, I added #796 to skip installing `debug` with TruffleRuby in IRB.

PR #793, opened 2023-11-28
Change show_source tests into integration tests

st0012:
It's always tricky to write `show_source` tests as the state change would affect `ruby/ruby`'s CI. So this change will stop them from running on `ruby/ruby` (because we skip integration tests there) but also make sure each test don't affect each other.

I also spotted a small bug when improving tests and fixed it here too.


lib/irb/source_finder.rb
```diff
@@ -77,6 +77,8 @@ def method_target(owner_receiver, s_count, method, type)
         target_method = target_method.super_method if target_method
       end
       target_method.nil? ? nil : target_method.source_location
+    rescue NameError
```
st0012, line 80: We can do more sophisticated rescue for individual `instance_method` and `method` calls. But for now with all the tests I think we can assume it's unlikely the logic itself would cause `NoMethodError`.

tompng (approved):
:+1:

PR #807, opened 2023-12-09
Simplify show_source's super calculation

st0012:
We can simplify `show_source` command's `-s` flag handling by removing the possibility to have it `nil`. 

I'm also working on a more sophisticated flag parsing mechanism for all commands. And this change will help simplify that solution too.

### DIFF
--- a/lib/irb/cmd/show_source.rb
+++ b/lib/irb/cmd/show_source.rb
@@ -27,17 +27,14 @@ module IRB
           puts "Error: Expected a string but got #{str.inspect}"
           return
         end
-        if str.include? " -s"
-          str, esses = str.split(" -")
-          s_count = esses.count("^s").zero? ? esses.size : 1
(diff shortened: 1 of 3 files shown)


lib/irb/cmd/show_source.rb
```diff
@@ -27,17 +27,14 @@ def execute(str = nil)
...
-        end
+
+        str, esses = str.split(" -")
+        super_level = esses ? esses.count("s") : 0
+        source = SourceFinder.new(@irb_context).find_source(str, super_level)
 
         if source
           show_source(source)
-        elsif s_count
+        elsif super_level
```
tompng, line 37: I think comparing with 0 is missing. `super_level` is changed to be always a number.

st0012: Good catch 👍 
I also added a test case for it.

tompng (approved):

Issue #674, opened 2023-08-09
Improve `.irbrc`'s loading behaviour


st0012:
I noticed 2 "surprising" `.irbrc` loading behaviour recently:

### Lack of hierarchy on rc file loading

Unlike many other tools that have "project level" and "user level" configurations, IRB only accepts a single `.irbrc` file [through a fixed loading order](https://github.com/ruby/irb/blob/153b1e9d1cab945e4b21eba12c1654b5876e17e6/lib/irb/init.rb#L392-L411). 

For example, it doesn't load `project/.irbrc` if `~/.irbrc` is present.

This is unintuitive to most users as popular tools like `Pry` or `VS Code` both have "user level" and "project level" configurations. And they usually load "user level" configs first, and allow "project level" configs to override them.

```
$ pry        
~/.pryrc loaded
project/.pryrc loaded
[1] pry(main)> 
```

### `.irb_history` is always placed at the same place as `.irbrc` is defined

This is also not an intuitive design. Most of the time, `.irb_history` should not be inside the project folder. So the current reality is, if a project wants to define `.irbrc` for all its developers, `.irb_history` needs to be added to `.gitignore` or the project needs to configure `IRB.conf[:HISTORY_FILE]` ([example](https://github.com/Shopify/ruby-lsp/pull/858)).

ktopolski:
Hello 👋 

First time `irb` contributor here - just wanted to say thank you for maintaining Ruby's core tool! 🙏

I'd love to contribute to give back to the community so I picked first part of the issue and made a PR. I feel like I'm walking in the dark a bit, but managed to test it on the "local"(github codespace) environment and it passes automatic tests as well. I wasn't able to verify the _yamatanooroti_ due to codespace limitations - might get back to it on local environment next week. 

I'm extremely open to feedback - would love to contribute more to irb in the future!

elfham:
Since #859, `~/.config/irb/irbrc` is loaded twice when `$XDG_CONFIG_HOME` is set.

```
% irb
irb(main):001> IRB.rc_files.select { |rc| File.exist?(rc) }
=> []
irb(main):002> exit
% echo "puts __FILE__" > ~/.config/irb/irbrc
% irb
/home/elfham/.config/irb/irbrc
irb(main):001> IRB.rc_files.select { |rc| File.exist?(rc) }
=> ["/home/elfham/.config/irb/irbrc"]
irb(main):002> exit
% export XDG_CONFIG_HOME=~/.config
% irb
[...]

PR #808, opened 2023-12-09
Improve irbrc loading order

ktopolski:
Ensures that "project level" `.irbrc` will always take precedence over "user level" `.irbrc`. 

Addresses part of #674

### DIFF
--- a/test/irb/test_init.rb
+++ b/test/irb/test_init.rb
@@ -50,6 +50,27 @@ module TestIRB
       end
     end
 
+    def test_rc_file_loading_order
+      tmpdir = @tmpdir
+      Dir.chdir(tmpdir) do
+        # tmpdir is ENV["HOME"], so another dir is needed
+        subdir = "#{tmpdir}/mydir"
+        FileUtils.mkdir_p("#{tmpdir}/mydir")
+        Dir.chdir(subdir) do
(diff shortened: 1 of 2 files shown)


test/irb/test_init.rb
```diff
@@ -50,6 +50,27 @@ def test_rc_file
       end
     end
 
+    def test_rc_file_loading_order
+      tmpdir = @tmpdir
+      Dir.chdir(tmpdir) do
+        # tmpdir is ENV["HOME"], so another dir is needed
+        subdir = "#{tmpdir}/mydir"
```
st0012, line 57: For clarify, let's name it `project_dir` instead.

```rb
project_dir = "#{tmpdir}/project"
```

test/irb/test_init.rb
```diff
@@ -50,6 +50,27 @@ def test_rc_file
...
+        FileUtils.mkdir_p("#{tmpdir}/mydir")
```
st0012, line 58: Then we can just use the above local here:

```
FileUtils.mkdir_p(project_dir)
```

test/irb/test_init.rb
ktopolski: Good catch, done 👍

test/irb/test_init.rb
ktopolski: Done, thank you!

PR #810, opened 2023-12-12
Prevent a warning: setting Encoding.default_external

mame:
This is a sync to changes in ruby/ruby master


st0012 (approved):
Thanks for the fix 👍

Issue #752, opened 2023-11-08
Upgrade tracer integration to work with the latest tracer gem


st0012:
The [`tracer` gem](https://github.com/ruby/tracer) has received a big upgrade this year, but IRB's [tracer integration](https://github.com/ruby/irb/blob/master/lib/irb/ext/tracer.rb) hasn't been updated accordingly. So it's now causing issues like #694.

This ticket aims to:

1. Update the tracer integration so it works with the latest `tracer` gem and the current IRB implementation.
2. Add tests for the tracer integration.

PR #811, opened 2023-12-12
Remove documents about deprecated/WIP features and some slight adjustments

st0012:


lib/irb.rb
```diff
@@ -698,45 +698,6 @@
 # and <tt>$DEBUG</tt> to +true+;
 # these have no effect on \IRB output.
 #
-# === Tracer
```
st0012, line 701: Tracer integration needs some rework: https://github.com/ruby/irb/issues/752

I'll add the documentation back once that's completed.

lib/irb.rb
```diff
@@ -780,12 +741,6 @@
 #
 # Note that the configuration file entry overrides the command-line options.
 #
-# :stopdoc:
-# === \IRB Loader
```
st0012, line 784: I don't feel IRB Loader should be exposed to end users.

lib/irb.rb
```diff
@@ -850,12 +805,6 @@
 #
 #   IRB.conf[:AP_NAME] = 'my_ap_name'
 #
-# :stopdoc:
-# === \IRB Library Directory
```
st0012, line 854: This config is only used by the current tracer integration to skip IRB related traces. We don't need to disclose it in the doc.

lib/irb.rb
```diff
@@ -872,12 +821,6 @@
 # IRB.conf[:LC_MESSAGES]
 # :startdoc:
 #
-# :stopdoc:
-# === Single-IRB Mode
```
st0012, line 876: We're deprecating the multi-irb feature soon so let's not introducing the single/multi IRB concept.

lib/irb.rb
```diff
@@ -887,49 +830,7 @@
...
-#   * measure enables the mode to measure processing time. measure :off disables it.
-# * $, show_source
-#   * Show the source code of a given method or constant.
-# * @, whereami
-#   * Show the source code around binding.irb again.
-# * debug
-#   * Start the debugger of debug.gem.
-# * break, delete, next, step, continue, finish, backtrace, info, catch
-#   * Start the debugger of debug.gem and run the command on it.
+# Please use the `show_cmds` command to see the list of available commands.
```
st0012, line 833: We need to find a way to sync commands into the doc here. But for now, users can find them through readme or through the `show_cmds` command.

lib/irb.rb
```diff
@@ -986,67 +866,6 @@
 #   If +line_no+ is a negative, the return value +line_no+ many lines before
 #   the most recent return value.
 #
-# ==== Example using IRB Sessions
```
st0012, line 989: The example here has a heavy focus on the multi-irb feature we plan to deprecate, which is a bit misleading.

tompng (approved):

PR #813, opened 2023-12-13
Warn and do nothing if block is passed to measure command

tompng:
Ruby 3.3 is a good opportunity to add a warning message for measure command with block.
(Although, I think no one reads the warning because (maybe) no one use it.)

I searched the commit that added measure with block but it had no desciption. https://github.com/ruby/irb/pull/183 https://github.com/ruby/irb/pull/184

Measure command with block is
- Not documented
- Misleading. Usage is not `measure{heavy_logic_to_be_measured}`.
- Hard to use as a command
- Has alternative way
- Just making IRB's command system complicated

Usage of measure with block was
```ruby
# Type this in IRB session.
irb> measure do |context, code, line_no, &block|
irb>   t = Time.now
irb>   result = block.call
irb>   puts "time: #{Time.now - t}"
irb>   result
irb> end
# We already have "TIME" measure, so the actual measure command will be longer and complicated than this example.
```

Alternative API of measure with block
```ruby
# Can write this to .irbrc
irb> IRB.conf[:MEASURE_PROC][:FOOBAR] = measure_logic_proc
# Enable it
irb> measure :foobar

# Same as `measure do end`.
irb> IRB.set_measure_callback do |*, &block| measure_logic end
```

Alternative workaround for measure with block
```ruby
irb> t=Time.now; measure_target_code; Time.now - t

irb> def m; t = Time.now; [yield, Time.now - t]; end
irb> m { measure_target_code }
```
Measure with block will register a session-only measure method that you don't need to write to `.irbrc`` for reusing. For this purpose, these workaround is enough and easier than using measure command with block. That's why I think no one use it.

### DIFF
--- a/lib/irb/cmd/measure.rb
+++ b/lib/irb/cmd/measure.rb
@@ -12,32 +12,29 @@ module IRB
         super(*args)
       end
 
-      def execute(type = nil, arg = nil, &block)
+      def execute(type = nil, arg = nil)
         # Please check IRB.init_config in lib/irb/init.rb that sets
         # IRB.conf[:MEASURE_PROC] to register default "measure" methods,
         # "measure :time" (abbreviated as "measure") and "measure :stackprof".
+
+        if block_given?
+          warn 'Configure IRB.conf[:MEASURE_PROC] to add custom measure methods.'
+          return
+        end
+
         case type
         when :off
-          IRB.conf[:MEASURE] = nil
           IRB.unset_measure_callback(arg)
         when :list
           IRB.conf[:MEASURE_CALLBACKS].each do |type_name, _, arg_val|
             puts "- #{type_name}" + (arg_val ? "(#{arg_val.inspect})" : '')
           end
         when :on
-          IRB.conf[:MEASURE] = true
-          added = IRB.set_measure_callback(type, arg)
+          added = IRB.set_measure_callback(arg)
           puts "#{added[0]} is added." if added
         else
-          if block_given?
-            IRB.conf[:MEASURE] = true
-            added = IRB.set_measure_callback(&block)
-            puts "#{added[0]} is added." if added
-          else
-            IRB.conf[:MEASURE] = true
-            added = IRB.set_measure_callback(type, arg)
-            puts "#{added[0]} is added." if added
-          end
+          added = IRB.set_measure_callback(type, arg)
+          puts "#{added[0]} is added." if added
         end
         nil
       end
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -215,6 +215,7 @@ module IRB # :nodoc:
       added = [:TIME, IRB.conf[:MEASURE_PROC][:TIME], arg]
     end
     if added
+      IRB.conf[:MEASURE] = true
       found = IRB.conf[:MEASURE_CALLBACKS].find{ |m| m[0] == added[0] && m[2] == added[2] }
       if found
         # already added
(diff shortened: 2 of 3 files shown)


lib/irb/cmd/measure.rb
```diff
@@ -12,32 +12,29 @@ def initialize(*args)
...
             puts "- #{type_name}" + (arg_val ? "(#{arg_val.inspect})" : '')
           end
         when :on
-          IRB.conf[:MEASURE] = true
-          added = IRB.set_measure_callback(type, arg)
+          added = IRB.set_measure_callback(nil, nil)
           puts "#{added[0]} is added." if added
         else
-          if block_given?
-            IRB.conf[:MEASURE] = true
```
tompng, line 33: Changing `IRB.conf[:MEASURE]` is moved to `IRB.set_measure_calback` and `IRB.unset_measure_calback`.
This makes `IRB.set_measure_calback do end` to be a complete alternative of `measure do end`

lib/irb/cmd/measure.rb
```diff
@@ -12,32 +12,29 @@ def initialize(*args)
...
-          IRB.conf[:MEASURE] = nil
           IRB.unset_measure_callback(arg)
         when :list
           IRB.conf[:MEASURE_CALLBACKS].each do |type_name, _, arg_val|
             puts "- #{type_name}" + (arg_val ? "(#{arg_val.inspect})" : '')
           end
         when :on
-          IRB.conf[:MEASURE] = true
-          added = IRB.set_measure_callback(type, arg)
+          added = IRB.set_measure_callback(nil, nil)
```
tompng, line 33: `set_measure_callback(:on, nil)` does nothing (unless `IRB.conf[:MEASURE_PROC][:ON]` is set)
IRB was using "TIME" when `IRB.conf[:MEASURE] = true` is set and measure callback is not set.

Fixes this bug:
```
irb> measure :on
# Message `"TIME" is added.` is not show
irb> 1
processing time: 0.000032s
=> 1 # but "TIME" is added
```

lib/irb/init.rb
```diff
@@ -235,6 +236,7 @@ def IRB.unset_measure_callback(type = nil)
       type_sym = type.upcase.to_sym
       IRB.conf[:MEASURE_CALLBACKS].reject!{ |t, | t == type_sym }
     end
+    IRB.conf[:MEASURE] = nil if IRB.conf[:MEASURE_CALLBACKS].empty?
```
tompng, line 239: `measure :off, :measure_type` had a bug
```ruby
# lib/irb/cmd/measure.rb:20
when :off
  IRB.conf[:MEASURE] = nil
  IRB.unset_measure_callback(arg)
```
That disables all measure even if arg is specified.
```
irb> measure :foo
"FOO" is added
irb> measure :bar
"BAR" is added
irb> measure :off, :foo
# removes :foo and turns all measure off
irb> 42
# no measure
irb> measure
"TIME" added. (and BAR is re-enabled)
```

lib/irb/cmd/measure.rb
```diff
@@ -12,32 +12,29 @@ def initialize(*args)
...
-          IRB.conf[:MEASURE] = nil
           IRB.unset_measure_callback(arg)
         when :list
           IRB.conf[:MEASURE_CALLBACKS].each do |type_name, _, arg_val|
             puts "- #{type_name}" + (arg_val ? "(#{arg_val.inspect})" : '')
           end
         when :on
-          IRB.conf[:MEASURE] = true
-          added = IRB.set_measure_callback(type, arg)
+          added = IRB.set_measure_callback(arg)
```
tompng, line 33: `set_measure_callback(:on, arg)` had no effect (unless IRB.conf[:MEASURE_PROC][:ON] is set)
This will fix a bug that `measure :on` was enabling TIME without showing `"TIME" is added.` message.

This will also allow `measure :on, :foobar`, an opposite of `measure :off, :foobar`

st0012 (approved):
It's such a relief to see none of the recent posts/tweets about `measure` ([example](https://jemma.dev/blog/irb-measure)) mentioning the block argument 😄 

I think this can help us avoid a condition in #805 too.

PR #823, opened 2023-12-22
Add up command

k0kubun:
Using binding.irb, I sometimes run into a situation where I want to see the caller frame's context. This PR supports `up` command to seamlessly enter the `rdbg:irb` mode just like other commands that do so.

There are also `down` and `frame`. However, I think we don't need `down` in the IRB side because you don't need to use it before `up`. Similarly, you likely don't use `frame` before running `backtrace`. So I think it's fine to leave them as is.


k0kubun:
Maintainers seem to want to solve this by allowing to configure such commands easily, so closing this.

PR #824, opened 2023-12-24
Command implementation not by method

### LINKED ISSUE #592: binding.irb pollutes the context object's methods
## Description

Given the code:

```rb
class Foo
  def foo
    binding.irb
    puts "Hello"
  end
end

Foo.new.foo
```

If we run `ls` from `binding.irb`, we'd get 

```
Foo.methods: 
  backtrace     bindings  bt                  cb        chws                         conf     context  continue  cws              cwws      debug      delete                     edit                         exit              fg         finish      help      info      
  irb_bindings  irb_cb    irb_change_binding  irb_chws  irb_current_working_binding  irb_cwb  irb_cws  irb_cwws  irb_pop_binding  irb_popb  irb_popws  irb_print_working_binding  irb_print_working_workspace  irb_push_binding  irb_pushb  irb_pushws  irb_pwb   irb_pwws  
  irb_quit      jobs      kill                ls        measure                      popb     popws    pushb     pushws           pwws      quit       show_cmds                  show_doc                     show_source       source     step        whereami  workspaces
IRB::ExtendCommandBundle#methods: 
  install_alias_method  irb       irb_backtrace  irb_break  irb_catch  irb_change_workspace  irb_context  irb_continue       irb_current_working_workspace  irb_debug    irb_debug_info  irb_delete    irb_edit         irb_exit    irb_fg    irb_finish    irb_help      
  irb_info              irb_jobs  irb_kill       irb_load   irb_ls     irb_measure           irb_next     irb_pop_workspace  irb_push_workspace             irb_require  irb_show_cmds   irb_show_doc  irb_show_source  irb_source  irb_step  irb_whereami  irb_workspaces
Foo#methods: foo
locals: _
=> nil
```

This shows that IRB's command methods got added to the object, which is a pretty bad side-effect. This is because IRB directly calls `extend(ExtendCommandBundle)` on the `context.main` [here](https://github.com/ruby/irb/blob/master/lib/irb.rb#L468).

I think we should stop treating commands as methods to properly solve this, which will also make command extension API easier to implement.

tompng:
Command implementation without using `obj.extend IRB::ExtendCommandBundle`. Command is not a method anymore.
Fixes #592
Continuation of #547

Command `command arg` was executed by `eval("#{command} #{arg}")` or `eval("command #{transformed_args(arg)}")`.
This pull request will change command execution to `load_command(command).execute(arg)`.

## No main object method polluting by default
Now, IRB has no helper method by default.
~If more than one helper method is defined in ExtendCommandBundle, (like Rails console adds `app` `new_session` `reload!` `helper` `controller`), IRB will reluctantly pollute main object.~ (postponed)

## What kind of input is a command?

IRB has command override policy `NO_OVERRIDE` `OVERRIDE_PRIVATE_ONLY` and `OVERRIDE_ALL` for defined methods.
New command recognition is:
- Every command is single line, so multiline input is not a command
- `command_name arg` might be a command
- check private_method and public_method with the same name exists or not, and follow the override policy.

| input | situation | type(before) | reason(before) | type(after) | reason(after) |
| --- | --- | --- | --- | --- | --- |
| `exit` | private method `Kernel.exit` is defined | command | alias `irb_exit` to `exit` because it's ORVERRIDE_PRIVATE_ONLY | command | command name exists and it's OVERRIDE_PRIVATE_ONLY |
| `exit` | public method `exit` defined by user | expression | does not alias `irb_exit` to `exit` | expression | command name exists but it's OVERRIDE_PRIVATE_ONLY |
| `irb_measure` | | command | ruby evaluates as method because command method exists | command | command name exists |
| `measure` | local var `measure` defined | expression | ruby evaluates as lvar | command | command name exists |
| `measure off` | local var `measure` defined | command | ruby evaluates as method call | command | command name exists |
| `info 'a' + 'b'` | context.main is Logger.new | wrong expression | `logger.info("'a' + 'b'")` because IRB transform args even if command is not installed by NO_OVERRIDE policy | expression | `logger.info('a' + 'b')` because info is NO_OVERRIDE |
| `show_source * -s` | local var `show_source` defined | command | `show_source("* -s")` because transorm_args is defined | command | command name exists |
| `show_source += 1` | | command | `show_source("+= 1")` because transform_args is defined | command | command name exists |
| `info += 1` | | expression | ruby evaluates as opassign | command | command name exists |

## Command completion
Command was a method, so completor can complete it without extra effort.
Now, command is not a method, so this pull request adds command name completion to both RegexpCompletor and TypeCompletor.

## `conf` and `context` method
`conf` `context` method can be used as `conf.eval_history = 100` before.
It is now a command only to show configuration.

The source code comment says it just displays the configuration, and navigates to use `IRB.conf` for modifying it.
```
# Displays current configuration.
#
# Modifying the configuration is achieved by sending a message to IRB.conf.
def irb_context
  IRB.CurrentContext
end

@ALIASES = [
  [:context, :irb_context, NO_OVERRIDE],
  [:conf, :irb_context, NO_OVERRIDE],
  ...
]
```
But a different thing is written in the document that you can use `conf.eval_history = N`.

We can:
- Fix the document (My proposal)
- Show a message when input is `/\A(conf|context)\./`v
- Add a special path to allow it, e.g. `context.evaluate(code.gsub(/\A(conf|context)\./, 'IRB.CurrentContext.'))`
- Keep `conf` and `context` as a method. Polluting main object problem remains.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -929,7 +929,7 @@ module IRB
     # Creates a new irb session
     def initialize(workspace = nil, input_method = nil)
       @context = Context.new(self, workspace, input_method)
-      @context.workspace.load_commands_to_main
+      @context.workspace.load_helper_methods_to_main
       @signal_status = :IN_IRB
       @scanner = RubyLex.new
       @line_no = 1
@@ -950,7 +950,7 @@ module IRB
     def debug_readline(binding)
       workspace = IRB::WorkSpace.new(binding)
       context.replace_workspace(workspace)
-      context.workspace.load_commands_to_main
+      context.workspace.load_helper_methods_to_main
       @line_no += 1
 
       # When users run:
@@ -1028,7 +1028,7 @@ module IRB
               return statement.code
             end
 
-            @context.evaluate(statement.evaluable_code, line_no)
+            statement.execute(@context, line_no)
 
             if @context.echo? && !statement.suppresses_echo?
               if statement.is_assignment?
@@ -1084,10 +1084,7 @@ module IRB
         end
 
         code << line
-
-        # Accept any single-line input for symbol aliases or commands that transform
-        # args
-        return code if single_line_command?(code)
+        return code if command?(code)
 
         tokens, opens, terminated = @scanner.check_code_state(code, local_variables: @context.local_variables)
         return code if terminated
@@ -1114,23 +1111,46 @@ module IRB
       end
 
       code.force_encoding(@context.io.encoding)
-      command_or_alias, arg = code.split(/\s/, 2)
-      # Transform a non-identifier alias (@, $) or keywords (next, break)
-      command_name = @context.command_aliases[command_or_alias.to_sym]
-      command = command_name || command_or_alias
-      command_class = ExtendCommandBundle.load_command(command)
-
-      if command_class
-        Statement::Command.new(code, command, arg, command_class)
+      if (command, arg = parse_command(code))
+        command_class = ExtendCommandBundle.load_command(command)
+        Statement::Command.new(code, command_class, arg)
       else
         is_assignment_expression = @scanner.assignment_expression?(code, local_variables: @context.local_variables)
         Statement::Expression.new(code, is_assignment_expression)
       end
     end
 
-    def single_line_command?(code)
-      command = code.split(/\s/, 2).first
-      @context.symbol_alias?(command) || @context.transform_args?(command)
+    ASSIGN_OPERATORS = %w[= += -= *= /= %= **= &= |= &&= ||= ^= <<= >>=]
+    COMMAND_LIKE_ASSIGN_REGEXP = /\A[a-z_]\w* #{Regexp.union(ASSIGN_OPERATORS)}( |$)/
+
+    def parse_command(code)
+      command_name, arg = code.strip.split(/\s/, 2)
+      return unless code.lines.size == 1 && command_name
+      return if COMMAND_LIKE_ASSIGN_REGEXP.match?(code)
+
+      arg ||= ''
+      command = command_name.to_sym
+      # Command aliases are always command. example: $, @
+      if (alias_name = @context.command_aliases[command])
+        return [alias_name, arg]
+      end
+
+      # Check visibility
+      local_variable = @context.local_variables.include?(command)
+      public_method = !!Kernel.instance_method(:public_method).bind_call(@context.main, command) rescue false
+      private_method = !public_method && !!Kernel.instance_method(:method).bind_call(@context.main, command) rescue false
+      if ExtendCommandBundle.execute_as_command?(
+        command,
+        public_method: public_method,
+        private_method: private_method,
+        local_variable: local_variable
+      )
+        [command, arg]
+      end
+    end
+
+    def command?(code)
+      !!parse_command(code)
     end
 
     def configure_io
@@ -1148,9 +1168,7 @@ module IRB
               false
             end
           else
-            # Accept any single-line input for symbol aliases or commands that transform
-            # args
-            next true if single_line_command?(code)
+            next true if command?(code)
 
             _tokens, _opens, terminated = @scanner.check_code_state(code, local_variables: @context.local_variables)
             terminated
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1124,7 +1124,7 @@ module IRB
     COMMAND_LIKE_ASSIGN_REGEXP = /\A[a-z_]\w* #{Regexp.union(ASSIGN_OPERATORS)}( |$)/
 
     def parse_command(code)
-      command_name, arg = code.strip.split(/\s/, 2)
+      command_name, arg = code.strip.split(/\s+/, 2)
       return unless code.lines.size == 1 && command_name
       return if COMMAND_LIKE_ASSIGN_REGEXP.match?(code)
 
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1128,13 +1128,9 @@ module IRB
       end
     end
 
-    ASSIGN_OPERATORS = %w[= += -= *= /= %= **= &= |= &&= ||= ^= <<= >>=]
-    COMMAND_LIKE_ASSIGN_REGEXP = /\A[a-z_]\w* #{Regexp.union(ASSIGN_OPERATORS)}( |$)/
-
     def parse_command(code)
       command_name, arg = code.strip.split(/\s+/, 2)
       return unless code.lines.size == 1 && command_name
-      return if COMMAND_LIKE_ASSIGN_REGEXP.match?(code)
 
       arg ||= ''
       command = command_name.to_sym
@@ -1144,15 +1140,9 @@ module IRB
       end
 
       # Check visibility
-      local_variable = @context.local_variables.include?(command)
       public_method = !!Kernel.instance_method(:public_method).bind_call(@context.main, command) rescue false
       private_method = !public_method && !!Kernel.instance_method(:method).bind_call(@context.main, command) rescue false
-      if ExtendCommandBundle.execute_as_command?(
-        command,
-        public_method: public_method,
-        private_method: private_method,
-        local_variable: local_variable
-      )
+      if ExtendCommandBundle.execute_as_command?(command, public_method: public_method, private_method: private_method)
         [command, arg]
       end
     end
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1028,7 +1028,15 @@ module IRB
               return statement.code
             end
 
-            statement.execute(@context, line_no)
+            case statement
+            when Statement::EmptyInput
+              # Do nothing
+            when Statement::Expression
+              @context.evaluate(statement.code, line_no)
+            when Statement::Command
+              ret = statement.command_class.execute(@context, statement.arg)
+              @context.set_last_value(ret)
+            end
 
             if @context.echo? && !statement.suppresses_echo?
               if statement.is_assignment?
--- a/lib/irb/command/base.rb
+++ b/lib/irb/command/base.rb
@@ -10,6 +10,10 @@ module IRB
   module Command
     class CommandArgumentError < StandardError; end
 
+    def self.extract_ruby_args(*args, **kwargs)
+      throw :EXTRACT_RUBY_ARGS, [args, kwargs]
+    end
+
     class Base
       class << self
         def category(category = nil)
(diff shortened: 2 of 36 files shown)


lib/irb/command/base.rb
```diff
@@ -29,19 +33,13 @@ def help_message(help_message = nil)
...
         def highlight(text)
           Color.colorize(text, [:BOLD, :BLUE])
         end
       end
 
-      def self.execute(irb_context, *opts, **kwargs, &block)
-        command = new(irb_context)
-        command.execute(*opts, **kwargs, &block)
+      def self.execute(irb_context, arg)
+        new(irb_context).execute(arg)
```
st0012, line 42: Should we always `rstrip` command arg here? I can't think of any case that we'd want trailing spaces. I think it'd also avoid issues like #905

lib/irb/command/help.rb
```diff
@@ -6,18 +6,9 @@ class Help < Base
...
-          else # Otherwise, consider the input as a String for convenience
-            args.strip.dump
-          end
-        end
-      end
-
-      def execute(command_name = nil)
+      def execute(arg)
+        # Accept string literal for backward compatibility
+        command_name = unwrap_string_literal(arg)
```
st0012, line 11: I think I added the string literal support by mistake since the new help command doesn't need it. So we probably don't need to unwrap the string literal here.

lib/irb/completion.rb
```diff
@@ -86,6 +86,14 @@ def retrieve_files_to_require_from_load_path
         )
     end
 
+    def command_completions(preposing, target)
+      if preposing.empty? && !target.empty?
+        IRB::ExtendCommandBundle.command_names.select { _1.start_with?(target) }
```
st0012, line 91: Maybe `IRB::ExtendCommandBundle.command_names` can be (lazily) cached somewhere so we don't compute it repeatedly?

lib/irb/statement.rb
```diff
@@ -61,17 +61,16 @@ def is_assignment?
         @is_assignment
       end
 
-      def evaluable_code
-        @code
+      def execute(context, line_no)
+        context.evaluate(@code, line_no)
```
st0012, line 65: IMO, statement objects should only provide information and the evaluation should still happen in the `Irb` objects.

lib/irb/completion.rb
tompng: I think it's already cached
```ruby
def self.command_names
  command_override_policies.keys.map(&:to_s)
end
```
`command_override_policies` is cached, `.keys.map(&:to_s)` is not a heavy instruction compared to `select{_1.start_with?(target)}`, so the effect of caching names is small, I think.

lib/irb/statement.rb
tompng: Then we need to introduce this kind of branch, because we can't provide evaluable_code for command.
```ruby
case statement
when Statement::Expression
  context.evaluate(statement.code, line_no)
when Statement::Command
  ret = statement.command_class.execute(context, statement.arg)
  context.set_last_value(ret)
end
```
I think `def execute` is better than this case-when.

lib/irb/command/base.rb
st0012: But if we `rstrip` the arg here, then we don't need to call them manually in commands like `Break`?

lib/irb/command/base.rb
```diff
@@ -52,7 +50,26 @@ def initialize(irb_context)
...
+        else
+          str
+        end
+      end
+
+      def ruby_args(arg)
+        # Use throw and catch to handle arg that includes `;`
+        # For example: "1, kw: (2; 3); 4" will be parsed to [[1], { kw: 3 }]
+        catch(:EXTRACT_RUBY_ARGS) do
+          @irb_context.workspace.binding.eval "IRB::ExtendCommand.extract_ruby_args #{arg}"
```
st0012, line 68: Let's use the new class:

```suggestion
          @irb_context.workspace.binding.eval "IRB::Command.extract_ruby_args #{arg}"
```

lib/irb/statement.rb
st0012: I still prefer having the case statement inside `Irb#eval_input` than making statements prefer the evaluation. Otherwise we're risking having a wrong coupling like RubyLex had, which is hard to eradicate as we both know.

Also, in the future (maybe even in the same release), we'll change commands to NOT have a return value. When that happens, it'd be even less beneficial to encapsulate things in statements.

lib/irb/workspace.rb
```diff
@@ -108,8 +108,10 @@ def initialize(*main)
     # <code>IRB.conf[:__MAIN__]</code>
     attr_reader :main
 
-    def load_commands_to_main
-      main.extend ExtendCommandBundle
+    def load_helper_methods_to_main
+      if ExtendCommandBundle.has_helper_method? && !(class<<main;ancestors;end).include?(ExtendCommandBundle)
```
st0012, line 112: I think it's fine to extend the context even if `ExtendCommandBundle` has no methods as it'll simplify tests (e.g. we don't need 2 cases to test workspace commands). We'll also come up with a better solution for this when we support helper methods.

test/irb/test_type_completor.rb
```diff
@@ -9,6 +9,8 @@
   return
 end
 
+require 'irb/context'
+require 'irb/command'
 require 'irb/completion'
```
st0012, line 14: Does it make sense to just require entire irb here?

lib/irb.rb
```diff
@@ -1114,23 +1111,46 @@ def build_statement(code)
...
+
+      arg ||= ''
+      command = command_name.to_sym
+      # Command aliases are always command. example: $, @
+      if (alias_name = @context.command_aliases[command])
+        return [alias_name, arg]
+      end
+
+      # Check visibility
+      local_variable = @context.local_variables.include?(command)
```
st0012, line 1139: While it's considerate to take local variables into account, I feel local variables should never override a command's invocation because:
- It's the behaviour of current IRB, Pry, and `debug`, so I'd expect most users to be aware of this limitation already.
- Under the current behaviour, users can still use `p <var>` to retrieve the variable when a conflict happens.
- However, under the this version users will NOT be able to call a command if a same-name local is present.

lib/irb/command/base.rb
tompng: There is no need and no effect adding rstrip here. `arg` does not end with space, ensured in `CommandArgTest#test_arg`.

I added rstrip in Break#execute because:
```ruby
arg = "" # this arg does not end with " ", already stripped
pre_cmds = "continue #{arg}" #=> "continue " # this string might end with " " if arg is an empty string
pre_cmds = "continue #{arg}".rstrip #=> "continue"
```

lib/irb.rb
tompng: I think the limitation is fine, but if we're going to fix #803, to be consistent, I think local variables should take precedence over methods.

```ruby
info = 1 # assignment is considered not a command in this pull request
=> 1
info + 1 # this will be a command call if we don't consider local variables
=> 2
```

st0012: Yeah then it's probably better to leave the behaviour in #803 as is. I'll close it if you also agree.

lib/irb/workspace.rb
tompng: Done!
I think it's beneficial to keep main object clean as possible, but I'll consider it later after helper method support

lib/irb.rb
tompng: Although there's many things to consider, I want to keep #803 open because it will be an enhancement.

lib/irb/command/help.rb
```diff
@@ -6,27 +6,17 @@ class Help < Base
...
-            args
-          else # Otherwise, consider the input as a String for convenience
-            args.strip.dump
-          end
-        end
-      end
-
-      def execute(command_name = nil)
+      def execute(command_name)
+        # Accept string literal for backward compatibility
```
st0012, line 10: I think this has been dropped?

```
irb(main):001> help "irb_info"
Can't find command `"irb_info"`. Please check the command name and try again.

=> nil
irb(main):002> help irb_info
Show information about IRB.
=> nil
irb(main):003> 
```

```suggestion
```

lib/irb/command/ls.rb
```diff
@@ -20,27 +20,35 @@ class Ls < Base
...
-      def self.transform_args(args)
-        if match = args&.match(/\A(?<args>.+\s|)(-g|-G)\s+(?<grep>[^\s]+)\s*\n\z/)
-          args = match[:args]
-          "#{args}#{',' unless args.chomp.empty?} grep: /#{match[:grep]}/"
+      def execute(arg)
+        if match = arg.match(/\A(?<args>.+\s|)(-g|-G)\s+(?<grep>.+)$/)
+          if match[:args].empty?
+            use_main = true
+          else
+            obj = @irb_context.workspace.binding.eval(match[:args])
```
st0012, line 28: Nit: we can give the capture group with a clearer name:

```suggestion
        if match = arg.match(/\A(?<target>.+\s|)(-g|-G)\s+(?<grep>.+)$/)
          if match[:target].empty?
            use_main = true
          else
            obj = @irb_context.workspace.binding.eval(match[:target])
          end
```

lib/irb/statement.rb
```diff
@@ -16,7 +16,7 @@ def should_be_handled_by_debugger?
       raise NotImplementedError
     end
 
-    def evaluable_code
+    def execute(context, line_no)
```
st0012, line 19: I think this method should be removed?

test/irb/test_command.rb
```diff
@@ -210,6 +210,98 @@ def test_irb_info_lang
     end
   end
 
+  class CustomCommandTestCase < CommandTestCase
+    def setup
+      super
+      execute_lines("show_cmds\n") # To ensure command initialization is done
```
st0012, line 216: nit: let's use `help` here:

```suggestion
      execute_lines("help\n") # To ensure command initialization is done
```

lib/irb/command/help.rb
tompng: I dropped it in this commit https://github.com/ruby/irb/pull/824/commits/47c2210555e565bd0895e563275db0e263a59258
https://github.com/ruby/irb/pull/824#discussion_r1546233617

Do you think we should revert it?

st0012: No I meant since it's dropped, the comment isn't true anymore and can be removed?

st0012 (approved):
👍

lib/irb/command/help.rb
tompng: Thanks! removed

PR #825, opened 2023-12-25
Implement `exit_program` (`!!!`) command

sambostock:
Pry includes an `exit-program` command, also aliased as `quit-program` and `!!!` which calls `Kernel.exit`. This is useful when `binding.pry` has been used in a loop.

This implements a similar `exit_program` command, also aliased as `quit_program` and `!!!`.


lib/irb/extend-command.rb
```diff
@@ -197,6 +197,14 @@ def irb_context
         :irb_history, :History, "cmd/history",
         [:history, NO_OVERRIDE],
         [:hist, NO_OVERRIDE],
+      ],
+
+      [
+        :irb_exit_program, :ExitProgram, 'cmd/exit_program',
+        [:exit_program, NO_OVERRIDE],
+        [:quit_program, NO_OVERRIDE],
+        [:'exit-program', NO_OVERRIDE], # FIXME: This doesn't work properly; seems to match normal "exit" command
```
sambostock, line 206: This doesn't seem to work, but maybe we don't care?

I'd added these to match the commands in Pry, but I see we use `show_source` rather than `show-source`, so maybe we only want `exit_program` and `quit_program`.

st0012:
With #851, the new `exit!` command is now essentially `exit_program` without taking an status argument. Do you want to rebase this PR and then introduce the `!!!` as an alias to it?

PR #829, opened 2023-12-31
test_recovery_sigint: Ensure precondition is met

sorah:
1. Ensure precondition is met on test_recovery_sigint. It depends on its process has SIG_DFL sigaction for SIGINT. When its parent process set SIG_IGN which inherits to children, then this test fails. This patch ensures this precondition met regardless of inherited sigaction from its parent.
2. Add another variant of test_recovery_sigint to ensure IRB to preserve existing SIGINT handler other than SIG_DFL.

----

My build process had unintentionally had SIG_IGN for SIGINT so this test started to fail, and took time to understand the cause. But in general it'd be better to ensure precondition in each test case, so sending this patch with additional scenario.


tompng (approved):
👍

PR #832, opened 2024-01-02
Avoid completing empty input

st0012:
The candidate list for empty input is all methods + all variables + all constants + all keywords. It's a long list that is not useful.

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -406,7 +406,13 @@ module IRB
         else
           select_message(receiver, message, candidates.sort)
         end
-
+      when /^\s*$/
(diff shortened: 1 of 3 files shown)


st0012:
Hmm I don't really understand how this would change the dialog tests' result 🤔

tompng:
It's reline's bug that pasting `"code1\C-icode2"` is treated as `"\C-icode1code2"`

Changing `write("IR\C-i")` to `write("IR"); write("\C-i")` might make the test pass.

st0012:
Yep it works, thank you 🙌

tompng (approved):

PR #844, opened 2024-01-25
Add some table of contents item in README

ydah:


st0012:
Thanks for the PR! But I intentionally excluded them from the ToC because they're not helpful information to IRB users thus not worth taking 2 extra lines. And for contributors, linking to the contributing section should be enough IMO.

Issue #625, opened 2023-07-04
IRBRC file is loaded every time ruby reaches binding.irb


tompng:
## Description

If you have `.irbrc` file
```ruby
puts 'irbrc file loaded'
```

You can see that `binding.irb` loads `.irbrc` every time.

```ruby
$ irb -f
irb(main):001:0> 4.times{binding.irb}
irbrc file loaded
irb(main):001:0> exit
irbrc file loaded
irb(main):001:0> exit
irbrc file loaded
irb(main):001:0> exit
irbrc file loaded
irb(main):001:0> exit
=> 4
irb(main):002:0> exit
$
```

`binding.irb` loads `.irbrc` even if command line option `-f` (Don't initialize from configuration file) is specified.
I think it should follow `-f` option. Maybe it should load only once.
For reference, IRB's command `irb` (Start a child IRB) does not load `.irbrc`.

## Result of irb_info
```
irb(main):001:0> irb_info
Ruby version: 3.2.2
IRB version: irb 1.7.1 (2023-07-01)
InputMethod: RelineInputMethod with Reline 0.3.5
.irbrc path: /Users/tomoya.ishida/.irbrc
RUBY_PLATFORM: arm64-darwin22
LANG env: ja_JP.UTF-8
East Asian Ambiguous Width: 1
```

## Terminal Emulator
Terminal.app

## Setting Files

```
# ~/.irbrc
puts 'irbrc file loaded'
```

PR #848, opened 2024-01-31
Respect parent command line arguments in nested IRB scopes

nunosilva800:
When `binding.irb` is called, the new session is setup without command line arguments:
https://github.com/ruby/irb/blob/a641746b18faa35ef32eec55255894fffcd4b063/lib/irb.rb#L1516-L1517

This means that the child session does not respect the parent command line arguments.
Therefore when using `$ irb -f` (to ignore RC file), and opening a child session with `biding.irb`, the new session will load the RC file (if it exists).

See more in:
- https://github.com/ruby/irb/issues/625

The solution I present here is to hold a copy of ARGV in `@CONF`, and use that to setup child sessions.

Closes https://github.com/ruby/irb/issues/625

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1513,8 +1513,8 @@ class Binding
   #
   # See IRB for more information.
   def irb(show_code: true)
-    # Setup IRB with the current file's path and no command line arguments
-    IRB.setup(source_location[0], argv: [])
+    # Setup IRB with the current file's path and current session command line arguments
+    IRB.setup(source_location[0], argv: IRB.conf[:ARGV])
     # Create a new workspace using the current binding
     workspace = IRB::WorkSpace.new(self)
     # Print the code around the binding if show_code is true
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -43,9 +43,9 @@ module IRB # :nodoc:
 
   # initialize config
   def IRB.setup(ap_path, argv: ::ARGV)
-    IRB.init_config(ap_path)
+    IRB.init_config(ap_path, argv: argv)
     IRB.init_error
-    IRB.parse_opts(argv: argv)
+    IRB.parse_opts
     IRB.run_config
     IRB.load_modules
 
@@ -55,7 +55,7 @@ module IRB # :nodoc:
   end
 
   # @CONF default setting
-  def IRB.init_config(ap_path)
+  def IRB.init_config(ap_path, argv: ::ARGV)
     # class instance variables
     @TRACER_INITIALIZED = false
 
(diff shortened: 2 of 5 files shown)


lib/irb/init.rb
```diff
@@ -244,7 +245,8 @@ def IRB.init_error
   end
 
   # option analyzing
-  def IRB.parse_opts(argv: ::ARGV)
+  def IRB.parse_opts
+    argv = @CONF[:ARGV].dup || []
```
nunosilva800, line 249: Problems in tests:

1. https://github.com/ruby/irb/blob/a641746b18faa35ef32eec55255894fffcd4b063/test/irb/test_init.rb#L179
2. https://github.com/ruby/irb/blob/a641746b18faa35ef32eec55255894fffcd4b063/test/irb/test_init.rb#L206

Both of these assert that the ARGV array passed as argument is mutated. Is this really a behaviour that is important? Isn't it better to assert that the IRB.conf[...] values are properly set?

lib/irb.rb
```diff
@@ -1513,8 +1513,8 @@ class Binding
   #
   # See IRB for more information.
   def irb(show_code: true)
-    # Setup IRB with the current file's path and no command line arguments
-    IRB.setup(source_location[0], argv: [])
+    # Setup IRB with the current file's path and current session command line arguments
+    IRB.setup(source_location[0], argv: IRB.conf[:ARGV])
```
nunosilva800, line 1517: Don't really know how to test this, other than manually steps:

```sh
$ mkdir -p tmp
$ echo "puts 'irbrc file loaded'" > tmp/.irbrc 
$ bundle exec rake install
$ % cd tmp && echo "binding.irb" | irb -f && cd ..
```

Before:

```
Switch to inspect mode.
binding.irb
irbrc file loaded
Switch to inspect mode.

nil
```

After:

```
Switch to inspect mode.
binding.irb
Switch to inspect mode.

nil
```


---

I thought about doing this test:

```rb
    def test_new_session_inherits_parent_config
      write_rc <<~RUBY
        puts 'irbrc file loaded'
      RUBY

      write_ruby <<~'RUBY'
        require 'irb'
        IRB.start
      RUBY

      output = run_ruby_file(["-f"]) do
        type "binding.irb"
        type "exit!"
        type "exit!"
      end

      refute output.include?("irbrc file loaded")
    end
```

And change the helper method `ruby_ruby_file` to:
```rb
    def run_ruby_file(cmd_args=[], &block)
      cmd = [EnvUtil.rubybin, "-I", LIB, @ruby_file.to_path] + cmd_args
```

But the test times-out. I think it's because `safe_gets` or writing to the command in `PTY.spawn` is not working this way :/

nunosilva800:
Closing in favour of 
- https://github.com/ruby/irb/pull/850

PR #849, opened 2024-01-31
Add rubocop with a few basic styling rules

st0012:
While I dislike excessive linting warnings/errors we sometimes experience in projects, I think having a tool to enforce some universally agreed styles is beneficial in the long term. So I want to start introducing a few very basic rubocop rules that has been widely followed in this codebase (see `.rubocop.yml`).

As we gradually adopt more rules, the CI job and our editor (through LSP servers like Ruby LSP) will make them easy to follow and enforce as well.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,17 @@ jobs:
     with:
       engine: cruby-truffleruby
       min_version: 2.7
-
+  lint:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


st0012:
Errors are caused by TruffleRuby suites and not related to the changes here.

tompng (approved):
:+1:

ima1zumi (approved):
👍

PR #850, opened 2024-01-31
Skip re-setup when creating a child session

nunosilva800:


st0012 (approved):
This is amazing, thank you!

Issue #612, opened 2023-06-25
IRB doesn't save input history when exited with `exit!`


st0012:
Because IRB [relies on `IRB.conf[:AT_EXIT]` config](https://github.com/ruby/irb/blob/master/lib/irb/ext/save-history.rb#L55), which will be run in `at_exit`, to save a session's input history. When users exit the session with `exit!`, the input is not written.

### Reproduction Steps

1. Open an IRB session and run

    ```
    irb(main):001:0> a = 1
    => 1
    irb(main):002:0> b = 2
    => 2
    irb(main):003:0> exit
    ```
1. Open another session and type up-arrow to check the above history
1. In the same session, run
    
    ```
    irb(main):001:0> c = 3
    => 3
    irb(main):002:0> d = 4
    => 4
    irb(main):003:0> exit! # with a ! this time
    ```
1. Open another session and type up-arrow to view history

#### Expected Behaviour

Input from step 3 should be part of the history

#### Actual Behaviour

Input from step 3 is lost

tompng:
This is the expected behavior of `Kernel.exit!`. Skips all at_exit handlers and ensure statements.
I think we can add `exit!` command that executes `Kernel.exit`

tompng:
Just for my case, most of the time I use `exit!` is when I want to just Kernel.exit because exit is overridden in irb.
like
```ruby
loop{
  ...
  binding.irb # want to exit the program after debugging/investigating. don't want to skip cleanup (ensure, at_exit)
}
```

PR #851, opened 2024-01-31
Introduce exit! command

ignacio-chiazzo:
# Problem
Currently, `exit!` does not save the history of the console' input like the following example

```console
irb(main):001:0> a = 1
=> 1
irb(main):002:0> b = 2
=> 2
irb(main):003:0> exit
```

Open another session and type up-arrow to check the above history
`irb(main):001:0> b = 2 # (history is saved!)`

If we run the same for `exit!`, the history isn't saved.
Closes [#612](https://github.com/ruby/irb/issues/612)

# Solution

I extended the command `exit!` and used a new class `ExitForcedAction`. Note that we can't use `Exit!` as the class name since `!` isn't allowed.

This will call the `irb_exit!` method, which will throw the same exception as exit (`throw :IRB_EXIT`) and the main method `run` catches it and calls `save_history``.

# Demo

In the following example, I test that the history is saved and `Kernel.exit!` exits the console even when the session is nested.

https://github.com/ruby/irb/assets/11672878/f6993f51-e729-4f4c-9ce4-b40a9137c925

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -889,6 +889,10 @@ module IRB
     throw :IRB_EXIT
   end
 
+  def IRB.irb_exit!(*)
+    throw :IRB_EXIT
+  end
+
   # Aborts then interrupts irb.
   #
   # Will raise an Abort exception, or the given +exception+.
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -886,11 +886,11 @@ module IRB
 
   # Quits irb
   def IRB.irb_exit(*)
-    throw :IRB_EXIT
+    throw :IRB_EXIT, false
   end
 
   def IRB.irb_exit!(*)
-    throw :IRB_EXIT!
+    throw :IRB_EXIT, true
   end
 
   # Aborts then interrupts irb.
@@ -984,21 +984,18 @@ module IRB
       end
 
       begin
-        @forced_exit = false
+        forced_exit = false
 
-        catch(:IRB_EXIT) do
-          catch(:IRB_EXIT!) do
-            eval_input
-          end
-          @forced_exit = true
+        forced_exit = catch(:IRB_EXIT) do
+          eval_input
         end
       ensure
         trap("SIGINT", prev_trap)
         conf[:AT_EXIT].each{|hook| hook.call}
 
-        if @forced_exit
+        if forced_exit
           context.io.save_history if supports_history_saving
-          Kernel.exit!
+          Kernel.exit!(0)
         else
           context.io.save_history if save_history
         end
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -890,7 +890,7 @@ module IRB
   end
 
   def IRB.irb_exit!(*)
-    throw :IRB_EXIT
+    throw :IRB_EXIT!
   end
 
   # Aborts then interrupts irb.
@@ -972,7 +972,8 @@ module IRB
       conf[:IRB_RC].call(context) if conf[:IRB_RC]
       conf[:MAIN_CONTEXT] = context
 
-      save_history = !in_nested_session && conf[:SAVE_HISTORY] && context.io.support_history_saving?
+      supports_history_saving = conf[:SAVE_HISTORY] && context.io.support_history_saving?
+      save_history = !in_nested_session && supports_history_saving
 
       if save_history
         context.io.load_history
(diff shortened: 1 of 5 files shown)


st0012:
Calling `exit!` now behaves differently then `exit`:

**exit**

```
$ be ruby test.rb

From: test.rb @ line 1 :

 => 1: binding.irb
    2: binding.irb

irb(main):001> exit

From: test.rb @ line 2 :

    1: binding.irb
 => 2: binding.irb

irb(main):001> exit
```

**exit!**

```
$ be ruby test.rb

From: test.rb @ line 1 :

 => 1: binding.irb
    2: binding.irb

irb(main):001> exit! # the process exists right away
```

I think the solution should also maintain this behaviour while saving the history.

lib/irb.rb
```diff
@@ -889,6 +889,10 @@ def IRB.irb_exit(*)
     throw :IRB_EXIT
   end
 
+  def IRB.irb_exit!(*)
+    throw :IRB_EXIT!
```
st0012, line 893: We can pass return value when throwing. So if we do `throw :IRB_EXIT, true` here, I think `Irb#run` can be simplified to:

```rb
      # ....
      force_quit = false

      begin
        force_quit = catch(:IRB_EXIT) do
          eval_input
        end
      ensure
        trap("SIGINT", prev_trap)
        conf[:AT_EXIT].each{|hook| hook.call}
        context.io.save_history if save_history
        Kernel.exit!(0) if force_quit
      end
```

lib/irb/cmd/exit_forced_action.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  module ExtendCommand
+    class ExitForcedAction < Nop
+      category "IRB"
+      description "Exit the current irb session with exit!."
+
+      def execute(*)
+        IRB.irb_exit!
+      rescue UncaughtThrowError
+        Kernel.exit
```
st0012, line 16: Maybe this should be `Kernel.exit!(0)` instead?

lib/irb.rb
```diff
@@ -979,13 +984,21 @@ def run(conf = IRB.conf)
       end
 
       begin
-        catch(:IRB_EXIT) do
+        @forced_exit = false
```
st0012, line 987: Why does it have to be an ivar instead of a local?

lib/irb.rb
```diff
@@ -968,7 +972,8 @@ def run(conf = IRB.conf)
       conf[:IRB_RC].call(context) if conf[:IRB_RC]
       conf[:MAIN_CONTEXT] = context
 
-      save_history = !in_nested_session && conf[:SAVE_HISTORY] && context.io.support_history_saving?
+      supports_history_saving = conf[:SAVE_HISTORY] && context.io.support_history_saving?
+      save_history = !in_nested_session && supports_history_saving
```
st0012, line 976: Why do we need to separate this?

lib/irb/cmd/exit_forced_action.rb
```diff
@@ -0,0 +1,22 @@
...
+
+require_relative "nop"
+
+module IRB
+  # :stopdoc:
+
+  module ExtendCommand
+    class ExitForcedAction < Nop
+      category "IRB"
+      description "Exit the current irb session with exit!."
```
st0012, line 11: I think we can make the effect of this command more clear:

```suggestion
      description "Exit the current process."
```

st0012 (commented):
I think we're pretty close, just want to clarify 2 changes.

lib/irb.rb
ignacio-chiazzo: `save_history` is false when `in_nested_session` is true. When calling `exit!`, we still want to save the history even if the session is nested.

```console
binding.irb
binding.irb
exit! -> should save history
```

```console
binding.irb
binding.irb
exit -> should not save history
a = 3
b = 44
exit -> should save history
```

lib/irb.rb
ignacio-chiazzo: not need to. I just updated this part.

lib/irb.rb
st0012: I see, that's a good catch 👍 
Do you mind adding a test case for this too?

ignacio-chiazzo: It is covered by https://github.com/ruby/irb/pull/851/files#diff-c37538faf5a6aa1413dc2844e6e78c58656b6474985c895e743ed422be4b3875R278-R280. I can change the message to something more explicit if needed.

lib/irb.rb
```diff
@@ -979,13 +984,21 @@ def run(conf = IRB.conf)
...
           eval_input
         end
       ensure
         trap("SIGINT", prev_trap)
         conf[:AT_EXIT].each{|hook| hook.call}
-        context.io.save_history if save_history
+
+        if forced_exit
+          context.io.save_history if supports_history_saving
+          Kernel.exit!(0)
```
st0012, line 998: I'm changing the method to `Kernel.exit` instead as it achieves the same effect we want but still runs `at_exit` callbacks.

I also don't think it'll be confusing to users as after the PR `exit!` will be listed in the help message as a command.

```suggestion
          Kernel.exit(0)
```

st0012 (approved):
Thanks for the PR ❤️

nobu:
This added a new warning during the test.

[/home/runner/work/irb/irb/lib/irb/extend-command.rb:290: warning: delegator does not forward private method #exit!](https://github.com/ruby/irb/actions/runs/7857657441/job/21441803100#step:4:7)

PR #853, opened 2024-01-31
Stop `irb:rdbg` from saving duplicated history when there's no prior history file

st0012:
The issue (https://github.com/ruby/irb/issues/855) is caused by a combination of factors:

1. When user starts an IRB session without a history file, the `@loaded_history_lines` ivar is not defined.
2. If the user then starts the `irb:rdbg` session, the history counter is not set, because the `@loaded_history_lines` is not defined.
3. Because `irb:rdbg` saves the history before passing Ruby expression to the debugger, it saves the history with duplicated lines. The number grows in exponential order.
4. When the user exits the `irb:rdbg` session, the history file could be bloated with duplicated lines.

This commit fixes the issue by resetting the history counter even when `@loaded_history_lines` is not defined.

Closes https://github.com/ruby/irb/issues/855

### DIFF
--- a/test/irb/test_history.rb
+++ b/test/irb/test_history.rb
@@ -251,10 +251,6 @@ module TestIRB
 
   class IRBHistoryIntegrationTest < IntegrationTestCase
     def test_history_saving_with_debug
-      if ruby_core?
-        omit "This test works only under ruby/irb"
-      end
-
       write_history ""
 
       write_ruby <<~'RUBY'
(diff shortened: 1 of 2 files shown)


test/irb/test_history.rb
```diff
@@ -293,6 +289,51 @@ def foo
...
+      ).strip
+
+      assert_equal <<~HISTORY, File.read(history_file)
+        'irb session'
+        next
+        'irb:debug session'
+        step
+        irb_info
+        puts Reline::HISTORY.to_a.to_s
+        q!
```
st0012, line 333: Without the fix, the history would be:

```
+ 'irb session'
+ next
+ 'irb session'
+ next
+ 'irb:debug session'
+ 'irb session'
+ next
+ 'irb:debug session'
+ step
+ 'irb session'
+ next
+ 'irb:debug session'
+ step
+ irb_info
+ puts Reline::HISTORY.to_a.to_s
  'irb session'
  next
  'irb:debug session'
  step
  irb_info
  puts Reline::HISTORY.to_a.to_s
  q!
```

tompng (approved):

PR #856, opened 2024-02-01
Support repeating debugger input by passing empty input to it

st0012:
When typing empty input in `rdbg` sessions, `debug` repeats the last repeatable input so users don't need to type `step` or `next` repeatedly. But currently `irb:rdbg` session doesn't support this as IRB ignores empty input completely.

This PR supports the behaviour by adding representation for empty input, and pass it to the debugger when in the `irb:rdbg` session. I also added a test to check the current empty input behaviour in normal IRB sessions is not changed.

Closes https://github.com/ruby/debug/issues/1063

### DIFF
--- a/test/irb/test_debugger_integration.rb
+++ b/test/irb/test_debugger_integration.rb
@@ -6,7 +6,7 @@ require "tmpdir"
 require_relative "helper"
 
 module TestIRB
-  class DebugCommandTest < IntegrationTestCase
+  class DebuggerIntegrationTest < IntegrationTestCase
     def setup
       super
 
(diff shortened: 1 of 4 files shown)


test/irb/test_debugger_integration.rb
```diff
@@ -434,5 +434,25 @@ def test_multi_irb_commands_are_not_available_after_activating_the_debugger
...
+        puts "foo"
+        puts "bar"
+        puts "baz"
+      ruby
+
+      output = run_ruby_file do
+        type "next"
+        type ""
+        type "\n\r"
+        type "\n"
```
tompng, line 450: `type "\n"` is exactly same as `type ""` because `puts("")` is equal to `puts("\n")`
```ruby
# type(command) pushes command string to `@commands`
# method run_ruby_file in test/irb/helper.rb L129-131
while command = @commands.shift
  write.puts(command)
end
```

`type "\n\r"` writes `"\n\r\n"`. `\r` and `\n` are both assigned to `ed_newline` in `reline/key_actor/emacs.rb`. `type("\n\n\n")` is equivalent to `3.times{type ""}`.

I think we can use `type "\e\n\n"`. (`"\e\n"` is option+enter)

st0012: The idea was to test different ways of sending empty input to make sure we've covered them all with the implementation. But it looks like it's not necessary after all 😅  How about we test it simply with `""`?

tompng: > How about we test it simply with `""`?

I think it's ok 👍 

We can think of a way to pass multiline input in IntegrationTestCase in the future.

Issue #694, opened 2023-08-24
`--tracer` flag (and extension) broken with recent versions of `tracer`


adam12:
## Description

The `--tracer` flag breaks `irb` when used with `0.2.2` version of `tracer`. I believe this was slightly hinted at in #630.

It works when requiring `tracer`, whereas `show_cmds` is extended with tracing options.

## Result of irb_info

```
irb(main):001> irb_info
Ruby version: 3.2.1
IRB version: irb 1.7.4 (2023-07-15)
InputMethod: RelineInputMethod with Reline 0.3.7
RUBY_PLATFORM: arm64-darwin22
LANG env: en_CA.UTF-8
East Asian Ambiguous Width: 1
```

st0012:
@adam12 Yeah sorry I updated the tracer but didn't update the extension in IRB as I expected no one would use it due to lack of update on both sides.

Given that IRB's tracer extension hasn't received updated for many years, so I'd like to make it actually helpful in addition to fixing it for the new tracer. Would you mind sharing your use case of/expectation for it?

adam12:
Should we disable the broken version temporarily? I actually don't know how useful the `--tracer` command line argument is, to be honest. It was just something I noticed while working on something else.

Ideally, it would function similar to `irb/debug`, where the commands are registered but are mostly stubs, where the `tracer` library is required upon first call or warned if unable to load.

PR #857, opened 2024-02-02
Fix usage of tracer gem and add tests

nunosilva800:
Fixes support of option `--tracer` when `gem tracer >= 0.2.0` is installed.

The API for tracer changed a lot from 0.1.x to 0.2.x. For reference, here is the old API: https://github.com/ruby/tracer/blob/v0.1.1/lib/tracer.rb

This PR also adds some tests for this integration:

- `WITH_TRACER=true bundle exec rake test`
The new tests are executed. There is one test that verifies behaviour when `--tracer` is set, but gem is not bundled:
`Skipping test_use_tracer_when_gem_is_unavailable because 'tracer' gem is available.`

- `WITH_TRACER=false bundle exec rake test`
The new tests will be skipped with message `Skipping ContextWithTracerTest because 'tracer' gem is not available. Enable with WITH_TRACER=true.`

Closes https://github.com/ruby/irb/issues/752
Closes https://github.com/ruby/irb/issues/694

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ gem "test-unit-ruby-core"
 
 gem "rubocop"
 
+gem "tracer" if ENV["WITH_TRACER"] == "true"
 gem "debug", github: "ruby/debug", platforms: [:mri, :mswin]
 
 if RUBY_VERSION >= "3.0.0" && !is_truffleruby
--- a/lib/irb/ext/tracer.rb
+++ b/lib/irb/ext/tracer.rb
@@ -3,76 +3,30 @@
 #   irb/lib/tracer.rb -
 #   	by Keiju ISHITSUKA(keiju@ruby-lang.org)
 #
-
+# Loading the gem "tracer" will cause it to extend IRB commands with:
+# https://github.com/ruby/tracer/blob/v0.2.2/lib/tracer/irb.rb
 begin
   require "tracer"
 rescue LoadError
   $stderr.puts "Tracer extension of IRB is enabled but tracer gem wasn't found."
-  module IRB
-    class Context
-      def use_tracer=(opt)
-        # do nothing
-      end
-    end
-  end
   return # This is about to disable loading below
 end
 
 module IRB
-
-  # initialize tracing function
-  def IRB.initialize_tracer
-    Tracer.verbose = false
-    Tracer.add_filter {
-      |event, file, line, id, binding, *rests|
-      /^#{Regexp.quote(@CONF[:IRB_LIB_PATH])}/ !~ file and
-        File::basename(file) != "irb.rb"
-    }
-  end
-
-  class Context
-    # Whether Tracer is used when evaluating statements in this context.
-    #
-    # See +lib/tracer.rb+ for more information.
-    attr_reader :use_tracer
-    alias use_tracer? use_tracer
-
-    # Sets whether or not to use the Tracer library when evaluating statements
-    # in this context.
-    #
-    # See +lib/tracer.rb+ for more information.
-    def use_tracer=(opt)
-      if opt
-        Tracer.set_get_line_procs(@irb_path) {
-          |line_no, *rests|
-          @io.line(line_no)
-        }
-      elsif !opt && @use_tracer
-        Tracer.off
-      end
-      @use_tracer=opt
-    end
-  end
-
   class WorkSpace
     alias __evaluate__ evaluate
     # Evaluate the context of this workspace and use the Tracer library to
     # output the exact lines of code are being executed in chronological order.
     #
-    # See +lib/tracer.rb+ for more information.
-    def evaluate(context, statements, file = nil, line = nil)
-      if context.use_tracer? && file != nil && line != nil
-        Tracer.on
-        begin
+    # See https://github.com/ruby/tracer for more information.
+    def evaluate(statements, file = __FILE__, line = __LINE__)
+      if IRB.conf[:USE_TRACER] == true
+        CallTracer.new(colorize: Color.colorable?).start do
           __evaluate__(statements, file, line)
-        ensure
-          Tracer.off
         end
       else
-        __evaluate__(statements, file || __FILE__, line || __LINE__)
+        __evaluate__(statements, file, line)
       end
     end
   end
-
-  IRB.initialize_tracer
 end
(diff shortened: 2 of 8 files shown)


lib/irb/ext/tracer.rb
```diff
@@ -3,7 +3,8 @@
 #   irb/lib/tracer.rb -
 #   	by Keiju ISHITSUKA(keiju@ruby-lang.org)
 #
-
+# Loading the gem "tracer" will cause it to extend IRB commands with:
+# https://github.com/ruby/tracer/blob/v0.2.2/lib/tracer/irb.rb
```
nunosilva800, line 7: It took me some time to figure this out. Is this helpful?

lib/irb/ext/tracer.rb
```diff
@@ -19,17 +20,6 @@ def use_tracer=(opt)
...
 module IRB
-
-  # initialize tracing function
-  def IRB.initialize_tracer
-    Tracer.verbose = false
-    Tracer.add_filter {
-      |event, file, line, id, binding, *rests|
-      /^#{Regexp.quote(@CONF[:IRB_LIB_PATH])}/ !~ file and
-        File::basename(file) != "irb.rb"
-    }
```
nunosilva800, line 30: `Tracer.verbose` and `Tracer.add_filter` were available in v0.1.x and I could not find an equivalent in the 0.2.x API.

lib/irb/ext/tracer.rb
```diff
@@ -42,37 +32,7 @@ class Context
...
     # See +lib/tracer.rb+ for more information.
     def use_tracer=(opt)
-      if opt
-        Tracer.set_get_line_procs(@irb_path) {
-          |line_no, *rests|
-          @io.line(line_no)
-        }
-      elsif !opt && @use_tracer
-        Tracer.off
-      end
```
nunosilva800, line 52: Same here, `set_get_line_procs`, `off` / `on` are all method of the v0.1.x API, without an equivalent in v0.2.x

lib/irb/ext/tracer.rb
```diff
@@ -42,37 +32,7 @@ class Context
...
-        Tracer.on
-        begin
-          __evaluate__(statements, file, line)
-        ensure
-          Tracer.off
-        end
-      else
-        __evaluate__(statements, file || __FILE__, line || __LINE__)
-      end
-    end
```
nunosilva800, line 74: I could not figure out what this is supposed to do. The cardinality seems wrong: https://github.com/ruby/irb/blob/d8fb3246be699dea102a23410fe7555ae2905f9c/lib/irb/workspace.rb#L116

After correcting the cardinality we can see that this method is invoked, and always falls into the "else" case.

If there is a use case we can write a test for, then we can bring this back.

lib/irb/ext/tracer.rb
st0012: I added that because IRB currently doesn't have integration with it. Do you think dropping it will make things easier here?

lib/irb/ext/tracer.rb
st0012: If the tracer is on, we may be able to do something like:

```rb
CallTracer.new(pattern: /^#{Regexp.quote(@CONF[:IRB_LIB_PATH])}/).start do
  __evaluate__(statements, file, line)
end
```

Gemfile
```diff
@@ -19,6 +19,7 @@ gem "test-unit-ruby-core"
 
 gem "rubocop"
 
+gem "tracer" if ENV["WITH_TRACER"] == "true"
```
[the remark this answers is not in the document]
st0012: I think this feature is not helpful enough to become a dependency atm. So yes I think we can do it like this for now.

nunosilva800:
Rebased.

nunosilva800:
Hmm, just before the rebase we were all green: https://github.com/ruby/irb/actions/runs/7758254084

Is there a way to retry? The [current error does not seem related](https://github.com/ruby/irb/actions/runs/7758339028/job/21159885973?pr=857#step:4:14).

test/irb/test_tracer.rb
```diff
@@ -0,0 +1,84 @@
...
+      # Context still needs this setter to be defined
+      assert irb.context.respond_to?(:use_tracer=)
+    end
+
+    def test_use_tracer_when_gem_is_unavailable
+      IRB.conf[:USE_TRACER] = true
+      begin
+        gem 'tracer'
+        omit "Skipping test_use_tracer_when_gem_is_unavailable because 'tracer' gem is available."
+      rescue Gem::LoadError
```
st0012, line 28: IMO, in general a test case shouldn't rescue any error to test the error handling logic as it only captures part of the behaviour while trying to mimic the rest. Unfortunately we currently don't have a good infra for this kind of dependency-dependant tests. I think we can leave this case untested for now.

test/irb/test_tracer.rb
```diff
@@ -0,0 +1,84 @@
...
+        binding.irb
+      RUBY
+
+      output = run_ruby_file do
+        type "bar(Foo)"
+        type "exit!"
+      end
+
+      assert_include(output, "#depth:")
+      assert_include(output, "Foo.foo")
```
st0012, line 81: With the colorable change I suggested above, we should be able to assert more complete trace messages here.

test/irb/test_tracer.rb
nunosilva800: This is required to skip this test when `USE_TRACER=true` is set, but run it when it is not set / false.

nunosilva800:
Updated with feedback and rebased.

adam12:
I've tested this PR and it works great, thanks!

One observation, and I am not sure if it's even possible, would be to strip off the calls from `IRB::WorkSpace#evaluate` and `Kernel#eval`.

```
irb(main):001> EID.generate_bytes
#depth:16>                  IRB::WorkSpace#evaluate at /Users/adam/code/github.com/ruby/irb/lib/irb/ext/tracer.rb:25:in `block in evaluate'
#depth:17>                   Kernel#eval at /Users/adam/code/github.com/ruby/irb/lib/irb/ext/tracer.rb:25:in `block in evaluate'
#depth:19>                     EID.generate_bytes at (irb):1:in `<top (required)>'
```

st0012 (approved):
Thanks for the fix!
I think some polishing can still be made, but I'll that in a follow up PR 👍

Issue #655, opened 2023-07-23
Use `OptionParser` to parse options


st0012:
Currently IRB parses command line options with `IRB.parse_opts`, which means the output of `irb --help` needs to be manually updated. We should use `OptionParser` instead to make the task easier.

PR #858, opened 2024-02-02
Use OptionParser to parse options

tomash:
Work-in-progress.

Fixes https://github.com/ruby/irb/issues/655

Checklist:

- [x] feature (options and values) parity with old `parse_opts` 
- [ ] remaining argv-based code for running scripts
- [ ] I18n (allowing removal of `lc/help-message`)

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -401,17 +401,17 @@ module IRB # :nodoc:
     load_path = []
     parser = OptionParser.new
 
-    parser.on('-f', "Don't initialize from configuration file.") do |value|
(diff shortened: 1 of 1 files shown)


tomash:
@alexanderadam Grüezi! Go ahead, as you can see from history I haven't worked on it since over a year. I actually forgot about this work, thanks for bringing it up!

The "only" remaining challenge is handling anything passed after "--", which is way outside OptionParser capabilities. My plan for next step was to isolate them for separate parsing and only use OptionParser on the remaining, "vanilla" arguments part.

PR #864, opened 2024-02-06
Polish tracer integration and tests

st0012:
1. I simplified tracer related test and CI setup
2. Also, tracer output now skips IRB's traces

### DIFF
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -61,7 +61,7 @@ module IRB
       @io = nil
 
       self.inspect_mode = IRB.conf[:INSPECT_MODE]
-      self.use_tracer = IRB.conf[:USE_TRACER] if IRB.conf[:USE_TRACER]
+      self.use_tracer = IRB.conf[:USE_TRACER]
       self.use_loader = IRB.conf[:USE_LOADER] if IRB.conf[:USE_LOADER]
       self.eval_history = IRB.conf[:EVAL_HISTORY] if IRB.conf[:EVAL_HISTORY]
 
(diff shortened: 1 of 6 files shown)


lib/irb/context.rb
```diff
@@ -162,7 +162,7 @@ def initialize(irb, workspace = nil, input_method = nil)
     private_constant :KEYWORD_ALIASES
 
     def use_tracer=(val)
-      require_relative "ext/tracer"
+      require_relative "ext/tracer" if val
       @use_tracer = val
```
tompng, line 166: I think this instance variable is never read.
Maybe we can set val to `IRB.conf[:USE_TRACER]` just like `context.use_loader=` does

st0012: Yeah it's not used. But if we remove the setter, that's probably a breaking-change?
If you're fine with dropping it right now (since it has been broken anyway), we can just do

```rb
class Context
  def initialize(...)
    require_relative "ext/tracer" if IRB.conf[:USE_TRACER]
    # ...
  end
end
```

WDYT?

tompng: I'm fine dropping, it

Or just remove the unused instance variable, this won't be a breaking change
```ruby
def use_tracer=(val)
  require_relative "ext/tracer" if val
end
```
Or change the ivar set to conf set. I think this is the original code's intention.
```ruby
def use_tracer=(val)
  require_relative "ext/tracer" if val
  IRB.conf[:USE_TRACER] = val
end
```

st0012: Updated 👍

tompng (approved):

PR #867, opened 2024-02-10
Polish the exit! command and its tests

st0012:
I splitted the changes into individual commits with explanation.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -889,10 +889,6 @@ module IRB
     throw :IRB_EXIT, false
   end
 
-  def IRB.irb_exit!(*)
-    throw :IRB_EXIT, true
-  end
-
   # Aborts then interrupts irb.
   #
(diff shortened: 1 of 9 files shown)


tompng (approved):

lib/irb.rb
```diff
@@ -993,12 +988,8 @@ def run(conf = IRB.conf)
...
         conf[:AT_EXIT].each{|hook| hook.call}
 
-        if forced_exit
-          context.io.save_history if supports_history_saving
-          Kernel.exit(0)
-        else
-          context.io.save_history if save_history
-        end
+        context.io.save_history if save_history
+        Kernel.exit(0) if forced_exit
```
nobu, line 992: Recommend `exit` or `exit(true)` instead of `exit(0)`.

tompng: Thank you. Address it in https://github.com/ruby/irb/pull/868

PR #868, opened 2024-02-11
Fix exit! command warning and method behavior

tompng:
Delegator warning for `irb(main):001> chws 1` can be suppressed.
There was a workaround for exit command. Add workaround for exit!.

Just like #298, `rescue UncaughtThrowError` part should emulate the original method. In this case, we should call `Kernel.exit!`


st0012 (approved):
Thanks for the fix!

PR #871, opened 2024-02-13
Improve constant lookup in SourceFinder

tompng:
Improve show_source of constant to find const owner more precise.

Constant lookup does not depend on `self`, but on `Module.nesting`
```ruby
X = 1
class A
  Y = 1
  class B
    Z = 1
    [X, Y, Z] # these are visible here. Module.nesting #=> [A::B, A]
  end
end

class A::B
  # [X, Z] # Y is not visible here. Module.nesting #=> [A::B]
end
```

In this test code I added, `X`, `Y`, `Z`, `Array` are all visible with value == 1
```ruby
X = 1
module M
  Y = 1
  Z = 2
end
class A
  Z = 1
  Array = 1
  class B
    include M
    Object.new.instance_eval { binding.irb }
  end
end
```
`Module.nesting` is `[A::B, A]` and lookup order seems `A::B → A → A::B.ancestors → A.ancestors → TOPLEVEL(Object)`

### DIFF
--- a/lib/irb/source_finder.rb
+++ b/lib/irb/source_finder.rb
@@ -69,10 +69,18 @@ module IRB
 
     def find_source(signature, super_level = 0)
       case signature
-      when /\A(::)?[A-Z]\w*(::[A-Z]\w*)*\z/ # Const::Name
+      when /\A(::)?[A-Z]\w*(::[A-Z]\w*)*\z/ # ConstName, ::ConstName, ConstPath::ConstName
         eval_receiver_or_owner(signature) # trigger autoload
-        base = @irb_context.workspace.binding.receiver.yield_self { |r| r.is_a?(Module) ? r : Object }
-        file, line = base.const_source_location(signature)
+        *parts, name = signature.split('::', -1)
+        base =
+          if parts.empty? # ConstName
+            find_const_owner(name)
+          elsif parts == [''] # ::ConstName
+            Object
+          else # ConstPath::ConstName
+            eval_receiver_or_owner(parts.join('::'))
+          end
+        file, line = base.const_source_location(name)
       when /\A(?<owner>[A-Z]\w*(::[A-Z]\w*)*)#(?<method>[^ :.]+)\z/ # Class#method
         owner = eval_receiver_or_owner(Regexp.last_match[:owner])
         method = Regexp.last_match[:method]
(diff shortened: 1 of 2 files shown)


lib/irb/source_finder.rb
```diff
@@ -69,10 +69,19 @@ def initialize(irb_context)
...
-        base = @irb_context.workspace.binding.receiver.yield_self { |r| r.is_a?(Module) ? r : Object }
-        file, line = base.const_source_location(signature)
+        *parts, name = signature.split('::', -1)
+        base = (
+          if parts.empty? # ConstName
+            find_const_owner(name)
+          elsif parts == [''] # ::ConstName
+            Object
+          else # ConstPath::ConstName
+            eval_receiver_or_owner(parts.join('::'))
```
st0012, line 81: In this case, the result of this should be the same as `eval_receiver_or_owner(signature)`? If that's the case, maybe we can store that and use it here?

lib/irb/source_finder.rb
```diff
@@ -69,10 +69,19 @@ def initialize(irb_context)
 
     def find_source(signature, super_level = 0)
       case signature
-      when /\A(::)?[A-Z]\w*(::[A-Z]\w*)*\z/ # Const::Name
+      when /\A(::)?[A-Z]\w*(::[A-Z]\w*)*\z/ # ConstName, ::ConstName, ConstPath::ConstName
         eval_receiver_or_owner(signature) # trigger autoload
-        base = @irb_context.workspace.binding.receiver.yield_self { |r| r.is_a?(Module) ? r : Object }
-        file, line = base.const_source_location(signature)
+        *parts, name = signature.split('::', -1)
+        base = (
```
st0012, line 75: nit: I think having these parentheses introduces a new style that didn't exist in this repo before 🤔 ?

lib/irb/source_finder.rb
tompng: If `signature == 'A::B::C'`, `parts.join('::')` will be `'A::B'`
We need to evaluate `A::B` in current binding and get `base = FooBar::A::B` (FooBar part depends on binding's `Module.nesting`)
Now we can find source by calling `base.const_source_location('C')`

st0012: I see, thanks for explaining 👍

st0012 (approved):

lib/irb/source_finder.rb
tompng: Updated to
```ruby
a =
  if cond
  else
  end
```
style

PR #873, opened 2024-02-15
Standardize command related names

st0012:
Before we expose command extension APIs, I think it'd be beneficial to first standardize (or even modernize) the names we use for command related concepts. This includes namespace, base class, and folder names:

1. Rename `lib/irb/extend-command.rb` to `lib/irb/command.rb`
2. Rename `lib/irb/cmd/*.rb` to `lib/irb/command/*.rb`
3. Rename `test/irb/test_cmd.rb` to `test/irb/test_command.rb`
4. Rename `ExtendCommand` to `Command` and alias it to `Command`
6. Rename `Nop` to `Base` and alias it to `Base`

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -9,7 +9,7 @@ require "reline"
 
 require_relative "irb/init"
 require_relative "irb/context"
-require_relative "irb/extend-command"
+require_relative "irb/command"
 
 require_relative "irb/ruby-lex"
 require_relative "irb/statement"
--- a/lib/irb/cmd/nop.rb
+++ b/lib/irb/cmd/nop.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+# This file is just a placeholder for backward-compatibility.
+# Please require 'irb' and inheirt your command from `IRB::Command::Base` instead.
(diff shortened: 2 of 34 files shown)


st0012:
I think we can maybe postpone the raise of deprecation warning. Otherwise, gems need to start requiring IRB release with this PR + start referencing to the new constants to avoid the warning. But at the same time, it won't bring immediate benefit to them, nor us, other than having clearer names in their codebase.

tompng:
It looks like some repository is requiring `"irb/cmd/nop"`. example: ruby/tracer, irbtoos [search result](https://github.com/search?q=%22irb%2Fcmd%2Fnop%22+NOT+is%3Afork+language%3ARuby+NOT+ISHITSUKA&type=code)

How about adding a file `irb/cmd/nop.rb` that only require `irb/command/base`?

st0012:
@tompng That's a good point 👍 
I added the file back but it actually doesn't need to require anything. In both `tracer` and `irbtools` (and I assume other tools too), they are requiring `irb` as well, which currently already loads the base command.

tompng (approved):

PR #877, opened 2024-02-16
Revamp `help` command

st0012:
1. It's now a standalone command implementation and `show_cmds` is simply an alias to it
2. All references to `show_cmds` are now made to `help`
3. It now supports `help <cmd>` to display the command's new `help_message` attribute OR description if its help message is not set

    - Example: Screenshot 2024-02-16 at 21 51 34

5. I added `help_message` to `edit` and `ls` commands and tweaked their description a bit


lib/irb/command/edit.rb
```diff
@@ -8,7 +8,23 @@ module IRB
   module Command
     class Edit < Base
       category "Misc"
-      description 'Open a file with the editor command defined with `ENV["VISUAL"]` or `ENV["EDITOR"]`.'
+      description 'Open a file or source location.'
+      banner <<~BANNER
+        Usage: edit [FILE or constant or method signature]
+
+        Open a file in the editor specified in in #{highlight('ENV["VISUAL"]')} or #{highlight('ENV["EDITOR"]')}
```
tompng, line 15: `specified in in` maybe typo?

lib/irb/command/base.rb
```diff
@@ -22,12 +22,21 @@ def description(description = nil)
           @description
         end
 
+        def banner(banner = nil)
```
tompng, line 25: I'm not familiar with the use of term `banner` for this use. Banner seems like a title or a short message for me.
If there is no blocker, how about `summary(short)` and `description(long)`? It is used in gemspec file.

Other idea for long description name: readme, help, help_message, help_content, detailed_description, manual, document

st0012: I picked `banner` because both Pry and byebug use it for long description in their [commands](https://github.com/pry/pry/blob/2ecbf160f0c833bf7d3425314f46dc1c19b5c03e/lib/pry/commands/change_prompt.rb#L10-L14) or [settings](https://github.com/deivid-rodriguez/byebug/blob/ddad523427d4ccb5c94aa4858b64b4bbf3d0399c/lib/byebug/settings/histsize.rb#L12).

I prefer matching existing tools' naming convention because it means library authors can easily port their features to IRB once we expose extension APIs.

But even without the above factor, I feel banner is also a reasonable name as one of its meanings is: 

> a heading or advertisement appearing on a web page in the form of a bar, column, or box.

In our case, that's the command's detailed introduction in a box-ish area.

> If there is no blocker, how about summary(short) and description(long)? It is used in gemspec file.

If we did adopt it:
- `description` in IRB commands will have the opposite meaning than `Pry`'s
- IRB and tools like tracer, irbtools need to swap all their `description` after the release

While I'm ok with the naming, I wonder if it's worth the additional work?

lib/irb/command/edit.rb
st0012: Fixed 👍

lib/irb/command/base.rb
tompng: If the name is from Pry, I'm against using the word banner because pry's banner is not a detailed description content itself.


Pry uses gem `slop` for parsing command options. Pry's command take the word banner from Slop.
Commit that added `Command.banner` to pry: https://github.com/pry/pry/commit/bb4807724169a8cee5dea3b0870d6a1dd338a781
Slop's API is similar to optionparser which also has `banner`.
```ruby
require 'optparse'
opt = OptionParser.new
opt.on('--foo', 'foo'){}
opt.on('--bar', 'bar'){}
puts opt.help
opt.banner = 'BANNER'
puts opt.help
```
will output
```
Usage: irb [options]
        --foo                        foo
        --bar                        bar
BANNER
        --foo                        foo
        --bar                        bar
```
For optparse and slop, the BANNER line is actually a banner, assuming it to be a short headline title. It can be long, but seems like abuse.

st0012: Ok, then I'd pick `help` instead 👍

tompng (approved):
:+1:

PR #878, opened 2024-02-17
Remove IRB::NotImplementedError

tompng:
`NotImplementedError` is defined globally, and can be used as `raise NotImplementedError`
But inside IRB module, `raise NotImplementedError` raises ArgumentError.
```
irb(main):001* module IRB
irb(main):002*   raise NotImplementedError
irb(main):003> end
gems/irb-1.11.2/lib/irb/lc/ja/error.rb:16:in `initialize': wrong number of arguments (given 0, expected 1) (ArgumentError)
```

Remove `IRB::NotImplementedError` and `IRB::OutputMethod::NotImplementedError` because we don't need it.

### DIFF
--- a/lib/irb/lc/error.rb
+++ b/lib/irb/lc/error.rb
@@ -12,11 +12,6 @@ module IRB
       super("Unrecognized switch: #{val}")
     end
   end
-  class NotImplementedError < StandardError
-    def initialize(val)
-      super("Need to define `#{val}'")
-    end
-  end
   class CantReturnToNormalMode < StandardError
     def initialize
       super("Can't return to normal mode.")
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -20,7 +20,7 @@ module IRB
     #
     # See IO#gets for more information.
     def gets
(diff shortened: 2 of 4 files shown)


lib/irb/lc/error.rb
```diff
@@ -13,8 +13,8 @@ def initialize(val)
     end
   end
   class NotImplementedError < StandardError
```
st0012, line 15: I didn't know IRB has a `NotImplementedError` 🤯 

Should we remove this version and make the JA version simply inherit Ruby's `NotImplementedError`?

tompng: I want to leave it to reduce difference between locale and possibility of locale dependent bug.

st0012:
Actually, I think we don't need a locale-friendly version of `NotImplementedError` because it's an exception that target IRB developers instead of end users. So I prefer dropping `IRB::NotImplementedError` completely.

tompng:
Makes sense.
I Removed class `NotImplementedError` defined in IRB.

st0012 (approved):
Nice cleanup 👍

PR #882, opened 2024-02-23
Unroll extension method generation

st0012:
Given we only have 2 remaining extension setter methods, both of which only take 1 argument and don't have any alias, the current method generation logic is overly complicated.

This PR simplifies the method generation logic by simply defining the methods directly in the `IRB::Context` class.

I also found that `use_loader` is broken and fixed it in a separate commit.

### DIFF
--- a/lib/irb/ext/use-loader.rb
+++ b/lib/irb/ext/use-loader.rb
@@ -4,7 +4,7 @@
 #   	by Keiju ISHITSUKA(keiju@ruby-lang.org)
 #
 
-require_relative "../cmd/load"
+require_relative "../command/load"
 require_relative "loader"
 
 class Object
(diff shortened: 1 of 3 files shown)


tompng (approved):

lib/irb/ext/use-loader.rb
```diff
@@ -49,7 +49,7 @@ def use_loader=(opt)
       if IRB.conf[:USE_LOADER] != opt
         IRB.conf[:USE_LOADER] = opt
         if opt
-          if !$".include?("irb/cmd/load")
+          if !$".include?("irb/command/load")
```
nobu, line 52: This `if` seems doing nothing from the beginning.

PR #885, opened 2024-02-23
Remove useless loaded file check

tompng:
It was doing nothing from the beginning
`$"` never includes `"irb/command/load"` but always includes `"absolute/path/to/irb/command/load.rb"`

https://github.com/ruby/irb/pull/882#discussion_r1500555041


st0012 (approved):
Thanks 👍

PR #886, opened 2024-02-23
Allow defining custom commands in IRB

### LINKED ISSUE #513: Support command extension APIs

st0012:
## Overview

This is a feature that has been requested for a long time. It is now possible to define custom commands in IRB.

Example usage:

```ruby
require "irb/command"

class HelloCommand < IRB::Command::Base
  description "Prints hello world"
  category "My commands"
  help_message "It doesn't do more than printing hello world."

  def execute
    puts "Hello world"
  end
end

IRB::Command.register(:hello, HelloCommand)
```

And the command will appear in the help message:

```
Help
  help           List all available commands. Use `help <command>` to get information about a specific command.

My commands
  hello          Prints hello world

IRB
  ....
```

## Details

- All the default command definition stuff is now moved to `lib/irb/default_commands.rb` so requiring `irb/command` is very lightweight.
- 3 `Command` APIs are introduced 
    - `Command.register(command_name, command_class)` is the official API to register a command. I decided not to accept alias for now as the current aliasing mechanism is messy.
    - `Command._register_with_aliases(name, command_class, *aliases)` is an internal API for IRB to register its existing commands and their aliases.
    - `Command.commands` to get all the registered commands
- Many command handling methods are simplified with the change as well.

I think this is should probably be introduced before #624 because:

- It already simplifies IRB's command registration and will make future refactor easier
- I still haven't figured out how to better display helper methods in the help message
- The community is already familiar with the concept of commands, so I imagine this will be easy to adopt. And with custom commands, it'll help both us and the community understand the role of helper methods.

Closes #513


tompng:
I think this is basically OK :+1:
One concern is that once we release this public customizing API, fixing #592 might be harder than before (to keep public API backward compatibility).

What do you think? I think I'm ready with #824 that fixes #592.


I don't want to make `transform_args` a public API. (We can restrict only for default commands)
If we use the method name `execute(*ruby_args, **ruby_kwargs)`, execution method for raw string that previously used `transform_args` should be a different name like `execute_raw(string_arg)`. I think this method should be named `execute`.

st0012:
@tompng I've rebased the PR and made a few more improvements. Can you give it another look?

tompng (approved):
:+1:

PR #888, opened 2024-02-25
Restructure workspace management

st0012:
Currently, workspace is an attribute of IRB::Context in most use cases. But when some workspace commands are used, like `pushws` or `popws`, a workspace will be created and used along side with the original workspace attribute.

This complexity is not necessary and will prevent us from expanding multi-workspace support in the future.

So this PR introduces a @workspace_stack ivar to IRB::Context so IRB can have a more natural way to manage workspaces.


lib/irb/ext/workspaces.rb
```diff
@@ -6,55 +6,42 @@
...
     # information.
     def push_workspace(*_main)
       if _main.empty?
-        if workspaces.empty?
+        if @workspace_stack.size == 1
           print "No other workspace\n"
-          return nil
+        else
+          # swap the top two workspaces
+          previous_workspace, current_workspace = @workspace_stack.pop
```
tompng, line 20: I think this should be `@workspace_stack.pop(2)`

lib/irb/command/pushws.rb
```diff
@@ -15,7 +15,7 @@ class Workspaces < Base
       description "Show workspaces."
 
       def execute(*obj)
-        irb_context.workspaces.collect{|ws| ws.main}
+        irb_context.instance_variable_get(:@workspace_stack)[0..-2]&.collect{|ws| ws.main}
```
tompng, line 18: Maybe we can remove `[0..-2]&.`. (And we don't need `&.` because `array[0..x]` never be nil.)

The command name `pushws` `popws` and the new message `Can't pop the last workspace on the stack` indicates that current workspace is in the last position of the workspace stack. I think we can simply show the stack here.

Also, showing all the stack makes it clear what `pushws` (noarg) is doing.
```

irb("a"):005> pushws "b"
=> [main, "a", "b"]
irb("b"):006> pushws
=> [main, "b", "a"] # "a" and "b" are visually swapped
irb("a"):007>
```

lib/irb/command/pushws.rb
```diff
@@ -15,7 +15,7 @@ class Workspaces < Base
       description "Show workspaces."
 
       def execute(*obj)
-        irb_context.workspaces.collect{|ws| ws.main}
+        puts irb_context.instance_variable_get(:@workspace_stack).to_s
```
st0012, line 18: Technically this is a breaking change as now `workspaces.each` would break. But since we say it's a command, uses of its return value is not officially supported IMO.

tompng: I think breaking change about it is not a big problem 👍 

Maybe `.collect(&:main)` `.map(&:main)` like before?
The `#<IRB::WorkSpace(`  `)>` part is redundant and long, I think.

```
irb> pushws 42
[#<IRB::WorkSpace (main)>, #<IRB::WorkSpace ("a")>, #<IRB::WorkSpace (#<Object:0x00000001...>)>, #<IRB::WorkSpace (42)>]
```

lib/irb/workspace.rb
```diff
@@ -44,7 +44,7 @@ def initialize(*main)
             end
             Thread.pass
           end
-          @binding = BINDING_QUEUE.pop
+          @binding = BINDING_QUEUE.eval_history
```
tompng, line 47: I think calling `Thread::Queue#pop` here is correct

PR #890, opened 2024-02-26
Use markdown format for docs

st0012:
I used Rails's [rdoc-to-md](https://github.com/rails/rails/blob/main/tools/rdoc-to-md) script for the majority of the conversion, and then manually tweaked a few issues that I found.

2 additional changes:

- `rake rdoc` task is added to simplify the doc generation command.
- Hide top-level methods like `IRB.JobManager` from the generated docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 /pkg/
 /spec/reports/
 /tmp/
+/docs/
 Gemfile.lock
--- a/Rakefile
+++ b/Rakefile
@@ -47,6 +47,7 @@ task :default => :test
 RDoc::Task.new do |rdoc|
   rdoc.title = "IRB"
   rdoc.rdoc_files.include("*.md", "lib/**/*.rb")
+  rdoc.rdoc_files.exclude("lib/irb/xmp.rb")
   rdoc.rdoc_dir = "docs"
   rdoc.markup = "markdown"
   rdoc.options.push("--copy-files", "LICENSE.txt")
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,6 @@
 require "bundler/gem_tasks"
 require "rake/testtask"
+require "rdoc/task"
 
 Rake::TestTask.new(:test) do |t|
   t.libs << "test" << "test/lib"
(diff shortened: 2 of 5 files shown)


adam12:
Out of curiosity, why the switch?

st0012:
I think markdown has basically become the universal format in writing technical content. For example, we use it to write README and contributing guides all the time.

If we keep writing Ruby docs in another format, it means both the maintainers and the contributors need to learn or maintain this knowledge too, which is a huge cost to pay (for example, it makes me think twice before touching the docs). So without seeing a clear benefit of keep using the rdoc format, I think we should stop paying this additional cost and make the document easier to contribute to, at least to most devs.

(As a reference, almost 2 years ago Ruby also changed its contributing doc from rdoc to markdown format, among other things (https://github.com/ruby/ruby/pull/5899))

adam12:
> I think markdown has basically become the universal format in writing technical content. For example, we use it to write README and contributing guides all the time.

100% agree.

> If we keep writing Ruby docs in another format, it means both the maintainers and the contributors need to learn or maintain this knowledge too, which is a huge cost to pay (for example, it makes me think twice before touching the docs). So without seeing a clear benefit of using the rdoc format, I think we should stop paying this additional cost and make the document easier to contribute to, at least to most devs.

I also agree, 100%.


> (As a reference, almost 2 years ago Ruby also changed its contributing doc from rdoc to markdown format, among other things ([ruby/ruby#5899](https://github.com/ruby/ruby/pull/5899)))

I remember seeing this PR, but I don't recall seeing that Markdown for source code documentation was also considered/encouraged in ruby/* projects, but I am presuming this is the maintainer's choice.

Not worth getting into in this thread, but Ruby documentation is my least favourite part of Ruby currently, so any type of PR that seems to resolve some pain around using RDoc currently piques my interest.

hsbt:
I'm +1 to convert this. But I'm not sure what is happened mixing md and rdoc format in `ruby/ruby` repository.

FYI: I have a plan to extract `irb` from `ruby/ruby` repo entirely at https://bugs.ruby-lang.org/issues/20309. After that, `irb` and `reline` don't have some of concerns related `ruby/ruby` repo's limitation.

hsbt (approved):
@st0012 I discussed this at DevMeeting 2024-03-14. No one against this proposal if there is no issue to mix rdoc and markdown formats. Go ahead this.

st0012:
@hsbt Thanks for raising this in the dev meeting 🙌

PR #892, opened 2024-03-01
Escape closing square brackets in regexp

peterzhu2118:
Fixes the following warning:

    test/irb/test_command.rb:546: warning: regular expression has ']' without escape

Introduced in #888.

cc. @st0012


st0012:
Thanks!

PR #893, opened 2024-03-06
Prevent `irb_history`'s creation during HistoryTest

st0012:
Some cases of it currently create `~/.irb_history` files unintentionally while others don't. This is caused by the varying levels of setup/cleanup between them. This commit fixes the issue by wrapping every single test inside a consistent test setup and teardown callbacks.


st0012:
Merging this first to fix Ruby CI failures.

PR #898, opened 2024-03-08
Add disable_irb command to disable debug of binding.irb

hogelog:
This PR is a proposal to add the command `disable_irb`, which is equivalent to the command `disble-pry` in pry.

## Background
I often debug my programs with lots of `binding.irb`. Then, after I have finished debugging, the `binding.irb` I have planted sometimes makes it difficult to terminate the process normally.

Here is an extreme example.

```console
$ ruby -e '100.times{ binding.irb }'
irb(main):001> exit
irb(main):001> exit
irb(main):001> exit
... # 100 times exit required!
```

I often use the following workaround.

```console
$ ruby -e '1000.times{ binding.irb }'
irb(main):001> class ::Binding; def irb; end; end
=> :irb
irb(main):002> exit
```

## Proposal
With `disable_irb` to be added in this PR, it is easy to skip `binding.irb` called in a large number of loops, as shown below.

```console
$ ruby -e '1000.times{ binding.irb }'
irb(main):001> disable_irb
```

### DIFF
--- a/lib/irb/command/disable_irb.rb
+++ b/lib/irb/command/disable_irb.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module IRB
+  # :stopdoc:
+
+  module Command
+    class DisableIrb < Base
+      category "IRB"
+      description "Disable binding.irb."
+
+      def execute(*)
+        ::Binding.define_method(:irb) {}
+        IRB.irb_exit
+      rescue UncaughtThrowError
+        Kernel.exit
+      end
+    end
+  end
+
+  # :startdoc:
+end
--- a/lib/irb/command.rb
+++ b/lib/irb/command.rb
@@ -196,7 +196,12 @@ module IRB # :nodoc:
         :irb_history, :History, "command/history",
         [:history, NO_OVERRIDE],
         [:hist, NO_OVERRIDE],
-      ]
+      ],
+
+      [
+        :irb_disable_irb, :DisableIrb, "command/disable_irb",
+        [:disable_irb, NO_OVERRIDE],
+      ],
     ]
 
 


lib/irb/command/disable_irb.rb
```diff
@@ -0,0 +1,21 @@
...
+    class DisableIrb < Base
+      category "IRB"
+      description "Disable binding.irb."
+
+      def execute(*)
+        ::Binding.define_method(:irb) {}
+        IRB.irb_exit
+      rescue UncaughtThrowError
+        Kernel.exit
+      end
```
tompng, line 16: This part is not needed.

`exit` command has this rescue part. It is a workaround for this bug https://bugs.ruby-lang.org/issues/18234.
Root cause of the bug is https://github.com/ruby/irb/issues/592.
Unlike `exit` command and `Kernel.exit` method, `disable_irb` method is not defined in Kernel.

hogelog: Thanks for the review!
I removed the unnecessary code https://github.com/ruby/irb/pull/898/commits/6724bf42a2b32223a0d67bec4fc0fcb5962aed2f

tompng (approved):

hogelog:
Yay, thank you 🎉

paracycle:
Pry used to have a really nice `!!!` command that would unconditionally exit the process and I'd been using `exit!` (which comes from `Kernel`) to do the same in `irb`. That works really well without needing this kind of extra command.

Actually, if any extra commands will be added, I would really like to see the equivalent of `!!!` to be added.

tompng:
We already have `exit!` command equivalent to Pry's `!!!`. Added in https://github.com/ruby/irb/pull/851
The difference between `Kernel.exit!` is
`Kernel.exit!` skips every callbacks like ensure statement, at_exit, IRB's history saving, and all other cleanups.
`exit!` command executes `Kernel.exit` internally. Does not skip callbacks, saves IRB's history correctly.

paracycle:
@tompng Awesome, thanks for linking to that PR, this is great to hear. Indeed `Kernel.exit!` was just a workaround for me.

Given that `exit!` exists as a command now, is `disable_irb` still something useful?

st0012:
`disable_irb` is a combination of the `exit` command and disabling all subsequent `binding.irb` invocations. So when used, the program still runs but just ignores later breakpoints.

brandondrew:
I've always solved this problem by using `binding.irb unless $continue` in the first place... then I can just set `$continue = true` to skip all subsequent `binding.irb` calls.

I guess `disable_irb` is slightly more convenient... maybe...

chaadow:
image

I get this :/ , I don't know if I should enable something, but i'm on the latest stable IRB version ( 1.13.1 )

st0012:
@chaadow sorry that I accidentally dropped it when working on the command refactorings, which was not detected because there were no tests for it. I've opened https://github.com/ruby/irb/pull/964 to address it.

PR #901, opened 2024-03-15
Fix irb_history saved to current directory

### LINKED ISSUE #674: Improve `.irbrc`'s loading behaviour

tompng:
Fixes #674 with refactoring

IRB.rc_files is used for irbrc searching and irb_history searching. I think two separate logic is combined into one and making the code complicated.

### Changes
- Split `IRB.rc_files(ext)` to `IRB.irb_files` and `IRB.rc_file(ext)`
- history will be saved to `dirname(ENV['IRBRC'])`, `ENV['XDG_CONFIG_HOME']` or `ENV['HOME']`
- `IRB.conf[:RC_NAME_GENERATOR]` is removed because it's impossible to configure before loading irbrc
- Fix irbrc loaded twice bug mentioned in https://github.com/ruby/irb/issues/674#issuecomment-1994482350

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -395,10 +395,8 @@ module IRB # :nodoc:
   # Run the config file
   def IRB.run_config
     if @CONF[:RC]
-      rc_files.each do |rc|
-        # Because rc_file always returns `HOME/.irbrc` even if no rc file is present, we can't warn users about missing rc files.
-        # Otherwise, it'd be very noisy.
-        load rc if File.exist?(rc)
+      irbrc_files.each do |rc|
+        load rc
       rescue StandardError, ScriptError => e
         warn "Error loading RC file '#{rc}':\n#{e.full_message(highlight: false)}"
       end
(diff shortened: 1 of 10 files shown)


test/irb/test_init.rb
```diff
@@ -35,132 +40,82 @@ def test_setup_with_minimum_argv_does_not_change_dollar0
...
-      verbose, $VERBOSE = $VERBOSE, nil
       tmpdir = @tmpdir
       Dir.chdir(tmpdir) do
         ENV["XDG_CONFIG_HOME"] = "#{tmpdir}/xdg"
-        IRB.conf[:RC_NAME_GENERATOR] = nil
-        assert_equal(tmpdir+"/.irbrc", IRB.rc_file)
-        assert_equal(tmpdir+"/.irb_history", IRB.rc_file("_history"))
+        reset_rc_name_generators
+        assert_empty(IRB.irbrc_files)
+        assert_equal("#{ENV["HOME"]}/.irb_history", IRB.rc_file("_history"))
```
st0012, line 48: Isn't this the same as `tmpdir`? I understand that it's to emphasise that this value should be home, but I feel it could be misleading if we use `ENV["HOME"]` and `tmpdir` for the same value.

test/irb/test_init.rb
```diff
@@ -35,132 +40,82 @@ def test_setup_with_minimum_argv_does_not_change_dollar0
...
-      FileUtils.mkdir_p("#{tmpdir}/irb.rc")
-      FileUtils.mkdir_p("#{tmpdir}/$irbrc")
-
-      Dir.chdir(tmpdir) do
-        IRB.conf[:RC_NAME_GENERATOR] = nil
-        assert_includes IRB.rc_files, tmpdir+"/.irbrc"
-        assert_includes IRB.rc_files, tmpdir+"/_irbrc"
-        assert_includes IRB.rc_files, tmpdir+"/irb.rc"
-        assert_includes IRB.rc_files, tmpdir+"/$irbrc"
+        ENV['HOME'] = tmpdir
```
st0012, line 106: This is unnecessary because it's already done in `setup`.
```suggestion
```

test/irb/test_init.rb
```diff
@@ -35,132 +40,82 @@ def test_setup_with_minimum_argv_does_not_change_dollar0
...
+        ENV['XDG_CONFIG_HOME'] = "#{ENV['HOME']}/.config"
+        FileUtils.mkdir_p(ENV['HOME'])
```
st0012, line 108: I think it's not needed because `@tmpdir` is already present?

```suggestion
```

lib/irb/init.rb
```diff
@@ -395,42 +395,54 @@ def IRB.parse_opts(argv: ::ARGV)
...
-  def IRB.rc_file(ext = IRBRC_EXT)
-    warn "rc_file is deprecated, please use rc_files instead."
-    rc_files(ext).first
-  end
 
-  def IRB.rc_files(ext = IRBRC_EXT)
-    if !@CONF[:RC_NAME_GENERATOR]
-      @CONF[:RC_NAME_GENERATOR] ||= []
-      existing_rc_file_generators = []
+  def IRB.prepare_rc_name_generators
```
st0012, line 408: Maybe we can declare this as private?
Also, I feel `prepare_irbrc_name_generators` is a clearer name.

lib/irb/init.rb
```diff
@@ -446,13 +458,16 @@ def IRB.rc_file_generators
...
       yield proc{|rc| home+"/.irb#{rc}"}
-      yield proc{|rc| home+"/.config/irb/irb#{rc}"}
+      if xdg_config_home.nil? || xdg_config_home.empty?
+        yield proc{|rc| home+"/.config/irb/irb#{rc}"}
+      end
     end
+  end
+
+  # possible irbrc files in current directory
+  def IRB.generate_current_dir_irbrc_files
```
st0012, line 468: Let's make this private too.

test/irb/test_init.rb
tompng: I'll remove this `test_rc_file` because history file is tested more in another test `test_rc_files`

lib/irb/init.rb
tompng: Done.
Also changed rc_file_generators to private too

st0012 (approved):
Technically this is a breaking change as it removes `RC_NAME_GENERATOR` from `IRB.conf` entries. But from the [search](https://github.com/search?q=RC_NAME_GENERATOR&type=code&p=1) on GH, most (if not all) usages of it should have been using `ENV["IRBRC"]` instead. The number of its usages also looks small (maybe a few dozens?).
Therefore, I think we can bite the bullet and remove it to make things simpler.

tompng:
I think `IRB.conf[: RC_NAME_GENERATOR]` is already changed from Proc to Array and released as v1.12.0

PR #902, opened 2024-03-15
Explicitly show `--echo-on-assignment` is default on help

okuramasafumi:
Other options with `no` version (boolean flags) display with `(default)` but `--echo-on-assignment` doesn't have that.

### DIFF
--- a/lib/irb/lc/help-message
+++ b/lib/irb/lc/help-message
@@ -15,7 +15,7 @@ Usage:  irb.rb [options] [programfile] [arguments]
   --echo            Show result (default).
   --noecho          Don't show result.
   --echo-on-assignment
-                    Show result on assignment.
+                    Show result on assignment (default).
   --noecho-on-assignment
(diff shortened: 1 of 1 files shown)


tompng:
Thank you for the PR.
Actually, `--truncate-echo-on-assignment` is the default and `--echo-on-assignment` is not.

```
$ irb
irb(main):001> a=1000.times.to_a
=>
[0,
...
```

```
$ irb --truncate-echo-on-assignment
=>
[0,
...
```

```
$ irb --echo-on-assignment
irb(main):001> a=1000.times.to_a
=> 
[0,
 1,
 2,
 3,
 4,
```

```
$ irb --noecho-on-assignment
irb(main):001> a=1000.times.to_a
irb(main):002> 
```

okuramasafumi:
Ah, my mistake, so it's not boolean but 3-values option. Thank you!

PR #903, opened 2024-03-16
docs(help): Add latest options to ja help message

okuramasafumi:


lib/irb/lc/ja/help-message
```diff
@@ -9,8 +9,15 @@ Usage:  irb.rb [options] [programfile] [arguments]
   -W[level=2]       ruby -W と同じ.
   --context-mode n  新しいワークスペースを作成した時に関連する Binding
 		    オブジェクトの作成方法を 0 から 3 のいずれかに設定する.
+  --extra-doc-dir   ドキュメントダイアログにディレクトリを追加する.
```
tompng, line 12: I think this message is slightly incorrect.

English version is `Add an extra doc dir for the doc dialog.`.
I think this message will be `Add a directory to document dialog`
How about `追加のドキュメントディレクトリを指定する.` ?

tompng:
Could you also add help message for `--no-pager`?
This is the last missing translation. Adding this will make it perfect!

lib/irb/lc/ja/help-message
ima1zumi, line 12: The original text is difficult to understand little. `--extra-doc-dir` means the directories that are to be loaded during session startup.
ref: https://docs.ruby-lang.org/en/master/IRB.html#module-IRB-label-RI+Documentation+Directories

日本語で書くと'指定したディレクトリのドキュメントを読み込む'くらいの意味になると思います。

okuramasafumi:
@tompng @ima1zumi Added `--no-pager` and adopted the translation idea!

tompng (approved):
:+1:

ima1zumi (approved):
👍

PR #907, opened 2024-03-22
Switch to StdioInputMethod when TERM is 'dumb'

### LINKED ISSUE #68: --inf-ruby-mode semantics
Can we make it imply `--nosingleline` and `--nomultiline`, but not anything else?

In particular, we (the inf-ruby upstream) decided not to use it because it also disables continuation prompts. Currently `inf-ruby` uses `--prompt default --noreadline` instead.

But with newer features here that we can't use from Emacs, it would be helpful if `--inf-ruby-mode` was an option one could use to avoid version checks. In particular, if `irb --inf-ruby-mode --prompt default` was a valid usage in the new IRB.

### LINKED ISSUE #113: Change the editor module using an environment variable?
## Description

The latest version of IRB is broken when used through Emacs comint (the `inf-ruby` package).

The standalone `irb` works okay-ish if passed `--nomultiline`, (EDIT:) and for Rails one can do `rails console -- --nomultiline` (with recent enough rails). Any other program that similarly embeds Irb is still affected, and in all cases, when a script has to make a choice whether to pass `--nomultiline`, it creates a performance penalty on top of it.

If IRB read some environment variable and disabled multiline editing (singleline too maybe, it doesn't seem any effect here), that would be great.

## Terminal Emulator

Emacs comint.

## Setting Files

No. Changing `IRB.conf[:USE_MULTILINE]` through `.irbrc` works, but I can hardly ask every user to do that.

dgutov:
This works around Reline's misbehavior on low-capability terminals, such as when IRB is launched inside Emacs (which can look like ruby/reline#616, even though that issue is about non-tty usage). It should also resolve #68 and resolve #113 which were filed out of similar need.

The existing solution we use in Emacs with [inf-ruby](https://github.com/nonsequitur/inf-ruby/) is to detect the current version of the tool and launch it with both `--nomultiline` and `--nosingleline` when we think it supports those options. But that only covers one entry point -- having REPL launched from one of the few preset commands in `inf-ruby`, whereas all the other cases (custom console script by a user; or `binding.irb` invocation inside a test suite) remain problematic, requiring case-by-case solutions.

The provided simple change looks like a significant improvement in my testing.


dgutov:
N.B. the older "inverted triangle" issue seems to have been fixed in Reline, somewhere between the versions `0.1.5` and `0.4.3`. But the weird input repetition is still a problem.

lib/irb/context.rb
```diff
@@ -151,6 +151,10 @@ def initialize(irb, workspace = nil, input_method = nil)
       @command_aliases = @user_aliases.merge(KEYWORD_ALIASES)
     end
 
+    private def term_interactive?
+      STDIN.tty? && ENV['TERM'] != 'dumb'
```
tompng, line 155: How about `STDIN.tty? && STDOUT.tty?`? Does this solves your problem?

When STDIN and STDOUT are both tty, it is executed in an interactive terminal.
Reline and IRB both have a tty check, but there is an inconsistency. IRB only checks `STDIN.tty?`, and on the other hand, Reline only checks `$stdout.tty?`. If `$stdout` is not a tty or TERM is dumb, Reline enters unusable test mode.

`ENV['TERM'] == 'dumb'` is used for testing IRB with Reline. Although using `ENV['TERM']` like this is a bit weird, changing the TERM=dumb behavior right now will break the test.

dgutov: Unfortunately, both `STDIN.tty?` and `STDOUT.tty?` return true when the shell is running inside Emacs. But apparently some of the terminal features it's using don't work as expected in that environment. So treating "dumb" terminals differently makes sense to me (that's the definition - terminal with limited capabilities).

Do you know any other real-world "dumb" terminals? It could be useful to test IRB in one of them as well.

dgutov: The tests are indeed currently broken by this change, any recommendations for finding a fix would be welcome.

tompng: I see. Thanks for the explanation. I agree IRB should treat dumb terminals differently :+1:

For test, could you add a way to test IRB by forcing `term_interactive?` to be true?
Let it return true if `ENV["TEST_IRB_FORCE_INTERACTIVE"]` is set, and in IRB's test which uses `PTY.spawn`, specify both TEST_IRB_FORCE_INTERACTIVE and TERM=dumb env.

dgutov:
The "latest reline" breakage is the same on master.

tompng (approved):

tompng:
Looks good! thank you :+1:

PR #908, opened 2024-03-22
Fix indent test and output sequence test for new reline

tompng:
CI is failing after https://github.com/ruby/reline/pull/614 is merged.

### Indent is changed/fixed
For example, pasting `"a\n .b\n  .c"`:
```
Reline 0.4.2
> a
>   .b
> .c█

Reline 0.5.0.pre.1
> a
>   .b
>   .c█
```
This pull request changes the test to pass with both version by using `"a\n  .b\n  .c\n.d"`.

### DIFF
--- a/test/irb/test_irb.rb
+++ b/test/irb/test_irb.rb
@@ -70,11 +70,8 @@ module TestIRB
         type "exit"
       end
 
-      # Input cramped together due to how Reline's Reline::GeneralIO works
-      assert_include(
-        output,
-        "irb(main):001> irb(main):002> irb(main):002>  irb(main):002>   => nil\r\n"
-      )
+      assert_not_match(/irb\(main\):001> (\r*\n)?=> nil/, output)
+      assert_match(/irb\(main\):002>   (\r*\n)?=> nil/, output)
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/irb/test_irb.rb
```diff
@@ -70,11 +70,10 @@ def test_empty_input_echoing_behaviour
         type "exit"
       end
 
-      # Input cramped together due to how Reline's Reline::GeneralIO works
-      assert_include(
-        output,
-        "irb(main):001> irb(main):002> irb(main):002>  irb(main):002>   => nil\r\n"
-      )
+      # Ignore ANSI color codes
```
st0012, line 73: In IRB we disable coloring when the terminal is `dumb`, which the integration tests use. I feel we should match that behaviour in Reline as well, or make coloring controllable from IRB.

tompng: This part is acutaly ignoring color reset code `"\e[0m"`, not color specifying code.
Updated to:
```ruby
# Ignore ANSI color reset escape sequences
output = output.gsub(/\e\[0?m/, '')
```
Do you think Reline should not send color reset code when TERM is dumb?

tompng: This part can be removed after https://github.com/ruby/reline/pull/661

tompng: removed

st0012 (approved):

PR #917, opened 2024-04-07
Support `IRB.conf[:BACKTRACE_FILTER]`

### LINKED ISSUE #914: Introduce IRB.conf[:backtrace_filter] for upstream projects to extend filter_backtrace
## Description

If upstream projects (like Rails) want to extend or alter the behaviour of `filter_backtrace`, they currently have to override the `filter_backtrace` method (like Rails does [here](https://github.com/rails/rails/blob/af773ae44a1ae465cc0d19b1e8bf3fc731221ef0/railties/lib/rails/commands/console/console_command.rb#L8)). Directly overriding a method of the internal `WorkSpace` component is not a great way to extend a behaviour. Instead, we could add a config that creates a more proper way for upstream projects to extend the filter.

A better way to do this would be to: 

>Introduce IRB.conf[:backtrace_filter] that’s invoked after filter_backtrace is called.
Ideally, IRB.conf[:backtrace_filter] = Rails.backtrace_cleaner should fully replace the current handle_exception patch. But we’ll see if that’s possible when we have a PR
The above means that IRB’d expect the value of backtrace_filter to implement a filter(array_of_backtrace) method

Thanks for the suggestion on how to approach this @st0012

st0012:
This config allows users to customize the backtrace of exceptions raised and displayed in IRB sessions. This is useful for filtering out library frames from the backtrace.

IRB expects the given value to response to `filter` method and return the filtered backtrace.

Closes #914 

### Notes

- Ideally backtrace filters should be chainable, like Rack middlewares, to allow multiple sources to inject their own filters. However, I haven't seen much use cases for backtrace filtering outside of [Rails'](https://github.com/rails/rails/blob/bad7ff1664fb05cc227d0386ee3cbe4c292efe05/railties/lib/rails/commands/console/console_command.rb), so I'm not sure if that'll actually be necessary.
- To make IRB compatible with older Rails versions, we need to keep `WorkSpace#filter_backtrace`'s behaviour the same at least for a few years.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1242,27 +1242,33 @@ module IRB
         irb_bug = true
       else
         irb_bug = false
-        # This is mostly to make IRB work nicely with Rails console's backtrace filtering, which patches WorkSpace#filter_backtrace
-        # In such use case, we want to filter the exception's backtrace before its displayed through Exception#full_message
-        # And we clone the exception object in order to avoid mutating the original exception
-        # TODO: introduce better API to expose exception backtrace externally
-        backtrace = exc.backtrace.map { |l| @context.workspace.filter_backtrace(l) }.compact
+        # To support backtrace filtering while utilizing Exception#full_message, we need to clone
+        # the exception to avoid modifying the original exception's backtrace.
         exc = exc.clone
-        exc.set_backtrace(backtrace)
+        filtered_backtrace = exc.backtrace.map { |l| @context.workspace.filter_backtrace(l) }.compact
+        backtrace_filter = IRB.conf[:BACKTRACE_FILTER]
+
+        if backtrace_filter
+          if backtrace_filter.respond_to?(:call)
+            filtered_backtrace = backtrace_filter.call(filtered_backtrace)
+          else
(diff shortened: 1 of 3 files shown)


lib/irb.rb
```diff
@@ -1222,27 +1222,33 @@ def handle_exception(exc)
...
-      if RUBY_VERSION < '3.0.0'
-        if STDOUT.tty?
-          message = exc.full_message(order: :bottom)
-          order = :bottom
-        else
-          message = exc.full_message(order: :top)
-          order = :top
+        if backtrace_filter
+          if backtrace_filter.respond_to?(:filter)
+            filtered_backtrace = backtrace_filter.filter(filtered_backtrace)
```
tompng, line 1233: I think `backtrace_filter.call` is better, assuming `backtrace_filter` to be a Proc. `.filter` is too specific to ActiveSupport::BacktraceCleaner's API. With `.call`, Rails can still specify backtrace_cleaner as a filter by:
```ruby
IRB.conf[:BACKTRACE_FILTER] = -> bt { backtrace_cleaner_instance.filter(bt) }
```
What do you think?

st0012: 👍 updated

tompng (approved):

PR #919, opened 2024-04-10
Add a workaround to make IRB work with debug's tests

st0012:


k0kubun:
Merging to fix ruby/ruby CI.

PR #920, opened 2024-04-11
Pass statements to Context#evaluate

st0012:
This has a few benefits:

- We can keep hiding the evaluation logic inside the Context level, which has always been the convention until #824 was merged recently.
- Although not an official API, gems like `debug` and `mission_control-jobs` patch `Context#evaluate` to wrap their own logic around it. This implicit contract was broken after #824, and this change restores it.

In addition to the refactor, I also converted some context-level evaluation tests into integration tests, which are more robust and easier to maintain.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1028,21 +1028,7 @@ module IRB
               return statement.code
             end
 
-            case statement
-            when Statement::EmptyInput
-              # Do nothing
-            when Statement::Expression
-              @context.evaluate(statement.code, line_no)
-            when Statement::Command
-              ret = statement.command_class.execute(@context, statement.arg)
-              # TODO: Remove this output once we have a better way to handle it
-              # This is to notify `debug`'s test framework that the current input has been processed
-              # We also need to have a way to restart/stop threads around command execution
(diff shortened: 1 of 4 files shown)


lib/irb.rb
```diff
@@ -1028,21 +1028,7 @@ def eval_input
...
-            when Statement::Expression
-              @context.evaluate(statement.code, line_no)
-            when Statement::Command
-              ret = statement.command_class.execute(@context, statement.arg)
-              # TODO: Remove this output once we have a better way to handle it
-              # This is to notify `debug`'s test framework that the current input has been processed
-              # We also need to have a way to restart/stop threads around command execution
-              # when being used as `debug`'s console.
-              # https://github.com/ruby/debug/blob/master/lib/debug/irb_integration.rb#L8-L13
-              puts "INTERNAL_INFO: {}" if @context.with_debugger && ENV['RUBY_DEBUG_TEST_UI'] == 'terminal'
```
st0012, line 1043: We won't need this workaround with this change because the current `debug` patch will be applied to command execution too.

tompng (approved):

PR #922, opened 2024-04-14
Remove internal-only methods from Command::Base

tompng:
I think we should hide `Command#ruby_args` and `Command#unwrap_string_literal` from custom command definition, (at least for now).

We can reduce the use of these methods. Most of the case these methods are used for backward compatibility in default commands. Hiding it, we can remove these methods in any time.

### DIFF
--- a/lib/irb/command/edit.rb
+++ b/lib/irb/command/edit.rb
@@ -8,6 +8,8 @@ module IRB
 
   module Command
     class Edit < Base
+      include IRB::Command::RubyArgsExtractor
+
       category "Misc"
       description 'Open a file or source location.'
       help_message <<~HELP_MESSAGE
--- a/lib/irb/command/edit.rb
+++ b/lib/irb/command/edit.rb
@@ -8,7 +8,7 @@ module IRB
 
   module Command
     class Edit < Base
-      include IRB::Command::RubyArgsExtractor
+      include RubyArgsExtractor
 
       category "Misc"
       description 'Open a file or source location.'
--- a/lib/irb/command/base.rb
+++ b/lib/irb/command/base.rb
@@ -50,25 +50,6 @@ module IRB
 
       attr_reader :irb_context
 
-      def unwrap_string_literal(str)
-        return if str.empty?
-
-        sexp = Ripper.sexp(str)
-        if sexp && sexp.size == 2 && sexp.last&.first&.first == :string_literal
-          @irb_context.workspace.binding.eval(str).to_s
-        else
-          str
-        end
-      end
-
-      def ruby_args(arg)
-        # Use throw and catch to handle arg that includes `;`
-        # For example: "1, kw: (2; 3); 4" will be parsed to [[1], { kw: 3 }]
-        catch(:EXTRACT_RUBY_ARGS) do
-          @irb_context.workspace.binding.eval "IRB::Command.extract_ruby_args #{arg}"
-        end || [[], {}]
-      end
-
       def execute(arg)
         #nop
       end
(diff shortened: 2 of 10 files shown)


lib/irb/command/edit.rb
```diff
@@ -8,6 +8,8 @@ module IRB
 
   module Command
     class Edit < Base
+      include IRB::Command::RubyArgsExtractor
```
st0012, line 11: I think the `IRB::Command` namespace is not necessary?

st0012 (commented):
Good idea 👍 
I think we can even move them into a separate file so the module will not be loaded automatically either:

```diff
diff --git a/lib/irb/command/base.rb b/lib/irb/command/base.rb
index 8bf494a..b078b48 100644
--- a/lib/irb/command/base.rb
+++ b/lib/irb/command/base.rb
@@ -55,28 +55,6 @@ module IRB
       end
     end
 
-    # Internal use only, for default command's backward compatibility.
-    module RubyArgsExtractor
-      def unwrap_string_literal(str)
-        return if str.empty?
-
-        sexp = Ripper.sexp(str)
-        if sexp && sexp.size == 2 && sexp.last&.first&.first == :string_literal
-          @irb_context.workspace.binding.eval(str).to_s
-        else
-          str
-        end
-      end
-
-      def ruby_args(arg)
-        # Use throw and catch to handle arg that includes `;`
-        # For example: "1, kw: (2; 3); 4" will be parsed to [[1], { kw: 3 }]
-        catch(:EXTRACT_RUBY_ARGS) do
-          @irb_context.workspace.binding.eval "IRB::Command.extract_ruby_args #{arg}"
-        end || [[], {}]
-      end
-    end
-
     Nop = Base
   end
 
diff --git a/lib/irb/command/internal_helpers.rb b/lib/irb/command/internal_helpers.rb
new file mode 100644
index 0000000..249b5cd
--- /dev/null
+++ b/lib/irb/command/internal_helpers.rb
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module IRB
+  module Command
+    # Internal use only, for default command's backward compatibility.
+    module RubyArgsExtractor # :nodoc:
+      def unwrap_string_literal(str)
+        return if str.empty?
+
+        sexp = Ripper.sexp(str)
+        if sexp && sexp.size == 2 && sexp.last&.first&.first == :string_literal
+          @irb_context.workspace.binding.eval(str).to_s
+        else
+          str
+        end
+      end
+
+      def ruby_args(arg)
+        # Use throw and catch to handle arg that includes `;`
+        # For example: "1, kw: (2; 3); 4" will be parsed to [[1], { kw: 3 }]
+        catch(:EXTRACT_RUBY_ARGS) do
+          @irb_context.workspace.binding.eval "IRB::Command.extract_ruby_args #{arg}"
+        end || [[], {}]
+      end
+    end
+  end
+end
diff --git a/lib/irb/default_commands.rb b/lib/irb/default_commands.rb
index 6025b05..d680655 100644
--- a/lib/irb/default_commands.rb
+++ b/lib/irb/default_commands.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative "command"
+require_relative "command/internal_helpers"
 require_relative "command/context"
 require_relative "command/exit"
 require_relative "command/force_exit"
```

lib/irb/command/edit.rb
st0012, line 11: I think this one was missed?

```suggestion
      include RubyArgsExtractor
```

st0012 (approved):

PR #925, opened 2024-04-17
Stop using ExtendCommandBundle internally

st0012:
This module was used to extend both commands and helpers when they're not separated. Now that they are, and we have a Command module, we should move command-related logic to the Command module and update related references.

This will make the code easier to understand and refactor in the future.

### DIFF
--- a/lib/irb/command.rb
+++ b/lib/irb/command.rb
@@ -18,12 +18,6 @@ module IRB # :nodoc:
       def register(name, command_class)
         @commands[name] = [command_class, []]
       end
-
-      # This API is for IRB's internal use only and may change at any time.
-      # Please do NOT use it.
-      def _register_with_aliases(name, command_class, *aliases)
-        @commands[name] = [command_class, aliases]
-      end
     end
   end
 end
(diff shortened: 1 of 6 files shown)


test/irb/test_command.rb
```diff
@@ -265,7 +265,7 @@ def execute(_arg)
     end
 
     def test_def_extend_command
-      IRB::Command._register_with_aliases(:foobar, FooBarCommand, [:fbalias, IRB::ExtendCommandBundle::OVERRIDE_ALL])
+      IRB::ExtendCommandBundle.def_extend_command(:foobar, FooBarCommand, nil, [:fbalias, IRB::Command::OVERRIDE_ALL])
```
st0012, line 268: I incorrectly updated the method called in this test in the command extension PR, so I restored it here.

lib/irb/command.rb
```diff
@@ -18,12 +18,6 @@ class << self
       def register(name, command_class)
         @commands[name] = [command_class, []]
       end
-
-      # This API is for IRB's internal use only and may change at any time.
-      # Please do NOT use it.
-      def _register_with_aliases(name, command_class, *aliases)
```
st0012, line 24: Since it's for internal use, I moved it to `default_commands.rb` instead.

tompng (approved):

PR #926, opened 2024-04-18
Prompt specifiers: documentation improvements

Uaitt:
While reading the docs for specifying your own custom prompt, I noticed a few issues.

First, the placeholder for specifying the indentation level and the line number are wrong. `*NN*i` does not work, nor `*NN*n` (replacing NN with digits of course). The right placeholders are actually `%NNi` and `%NNn`.

Additionally, it took me a lot of time to understand that NN meant that I had to use digits and not actual NN in the placeholder 🤣. I have added a brief line for both the placeholders that actually states that you should use digits and not the raw NN.

PR's open for discussion 😃.


tompng (approved):
Looks great! thank you

Issue #629, opened 2023-07-04
Commands that haven't been covered by tests


st0012:
We need tests for them in `test_cmd.rb`:

- [ ] `irb_require`
- [ ] `irb`
- [ ] `job`
- [ ] `fg`
- [ ] `kill`

PR #929, opened 2024-04-19
Add MultiIRB commands test

igaiga:
Added MultiIRB commands(jobs, irb, fg and kill) test. MultiIRB is a deprecated feature, so I only wrote a minimal test to see if it works.

I wrote these codes after seeing the following issue, but since it was a deprecated feature, you can discard it if you don't need it.
https://github.com/ruby/irb/issues/629


st0012 (approved):
Thank you!

PR #932, opened 2024-04-24
Command registration should take both strings and symbols as names

st0012:
This will save users some heads scratching when they try to register a command with a string name and found that it doesn't work (which I experienced when trying to port commands for Rails console).

I also rewrote converted custom command tests into integration tests to make test setup/cleanup easier.


tompng (approved):
Looks good :+1:

PR #936, opened 2024-04-27
Restore IRB.CurrentContext when binding.irb is nested

### LINKED ISSUE #650: IRB doesn't restore context after leaving the nested session
(Originally discovered by @tompng in https://github.com/ruby/irb/pull/575#discussion_r1264357018)

Because IRB overrides the global `IRB.conf[:MAIN_CONTEXT]` when a new session starts, and doesn't change it back to the previous value when the session exists, IRB doesn't support nested sessions well.

For example:

```rb
# test.rb
class Foo
  def self.bar
    binding.irb
  end
end
binding.irb
```

```
be ruby test.rb
    4:   end
    5: end
 => 6: binding.irb

irb(main):001:0> irb_context.workspace
=> #<IRB::WorkSpace:0x00000001038f6818 @binding=#<Binding:0x0000000100754760>, @main=main>
irb(main):002:0> Foo.bar

From: test.rb @ line 3 :

    1: class Foo
    2:   def self.bar
 => 3:     binding.irb
    4:   end
    5: end
    6: binding.irb

irb(Foo):001:0> irb_context.workspace # context of the nested session
=> #<IRB::WorkSpace:0x0000000104ef4160 @binding=#<Binding:0x0000000104ef5e70>, @main=Foo>
irb(Foo):002:0> exit # leaving the nested session
=> nil
irb(main):003:0> irb_context.workspace # <==== The context is not restored
=> #<IRB::WorkSpace:0x0000000104ef4160 @binding=#<Binding:0x0000000104ef5e70>, @main=Foo>
```

tompng:
Fix #650

```
$ irb
irb(main):001> 123.instance_eval{binding.irb}
irb(123):001> exit
=> nil
irb(main):002> IRB.CurrentContext.main
=> 123 # Should be `main`. `IRB.CurrentContext` is not restored
```

# Changes
Moved preparation of `conf[:MAIN_CONTEXT]` to `eval_input`.
Every `eval_input` was working before because we forgot to restore MAIN_CONTEXT, so MAIN_CONTEXT was always present.

Delete `conf` parameter from `def run` which is always identical to `IRB.conf`, so we don't have to pass that param to `eval_input`.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -898,7 +898,7 @@ module IRB
     else
       irb = Irb.new
     end
-    irb.run(@CONF)
+    irb.run
   end
 
   # Quits irb
@@ -989,10 +989,10 @@ module IRB
       input
     end
 
-    def run(conf = IRB.conf)
+    def run
+      conf = IRB.conf
       in_nested_session = !!conf[:MAIN_CONTEXT]
       conf[:IRB_RC].call(context) if conf[:IRB_RC]
-      conf[:MAIN_CONTEXT] = context
 
       save_history = !in_nested_session && conf[:SAVE_HISTORY] && context.io.support_history_saving?
 
@@ -1025,6 +1025,8 @@ module IRB
 
     # Evaluates input for this session.
     def eval_input
+      prev_context = IRB.conf[:MAIN_CONTEXT]
+      IRB.conf[:MAIN_CONTEXT] = @context
       configure_io
 
       each_top_level_statement do |statement, line_no|
@@ -1055,6 +1057,8 @@ module IRB
           end
         end
       end
+    ensure
+      IRB.conf[:MAIN_CONTEXT] = prev_context
     end
 
     def read_input(prompt)
@@ -1585,7 +1589,7 @@ class Binding
       # workspace
       binding_irb = IRB::Irb.new(workspace)
       binding_irb.context.irb_path = irb_path
-      binding_irb.run(IRB.conf)
+      binding_irb.run
       binding_irb.debug_break
     end
   end
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1109,6 +1109,7 @@ module IRB
 
     def each_top_level_statement
       loop do
+        IRB.conf[:MAIN_CONTEXT] = @context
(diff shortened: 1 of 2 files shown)


tompng:
I added the second commit just in case to reduce unexpected MAIN_CONTEXT restored to nil in multithread.
When binding.irb is executed in job-controlled multithread, binding.irb may ends with random order.
Even with the second commit, there is a chance MAIN_CONTEXT being nil or wrong because we don't use mutex or thread local variable yet. (Example: uncontrolled multithread binding.irb)

Example of job-controlled multiple binding.irb.
```ruby
def current
  [self, IRB.CurrentContext.main]
end
def pass(thread_id = nil)
  thread_id ||= next_thread_id
  raise "No such thread #{thread_id}" unless $queues[thread_id]
  $queues[thread_id].enq :pass
  $queues[$threads.index(Thread.current)].deq
  nil
end
def next_thread_id
  thread_id = $threads.index(Thread.current)
  (1..4).map { (thread_id + _1) % $queues.size }.find { $queues[_1] }
end
$queues = 4.times.map { Queue.new }
$threads = $queues.map.with_index do |queue, thread_id|
  "Thread #{thread_id}".instance_eval do
    Thread.new do
      queue.deq
      binding.irb
      $queues[thread_id] = nil
      $queues[next_thread_id].enq :pass if next_thread_id
    end
  end
end
$queues[0].enq :pass
$threads.map(&:join)
```

```ruby
$ ruby file.rb
irb(Thread 0):001> pass 1
irb(Thread 1):001> pass 2
irb(Thread 2):001> pass 3
irb(Thread 3):001> pass 0
irb(Thread 0):002> exit # This will restore MAIN_CONTEXT to nil
irb(Thread 1):002> pass 3
irb(Thread 3):002> exit
irb(Thread 1):003> pass 2
irb(Thread 2):002> exit
irb(Thread 1):004> exit
# Order: start 0, start 1, start 2, start 3, exit 0, exit 3, exit 2, exit 1
```

tompng:
Restoring to nil may crash IRB.
I'll create more simple pull request first, and re-think later for fundamental solution.

PR #942, opened 2024-05-01
Revert "Switch to StdioInputMethod when TERM is 'dumb' (#907)"

tompng:
Reverts #907 to avoid ruby/ruby CI failure and Debug compatibility test failure

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -311,9 +311,7 @@ require_relative "irb/pager"
 # ### Input Method
 #
 # The IRB input method determines how command input is to be read; by default,
-# the input method for a session is IRB::RelineInputMethod. Unless the
-# value of the TERM environment variable is 'dumb', in which case the
-# most simplistic input method is used.
+# the input method for a session is IRB::RelineInputMethod.
 #
 # You can set the input method by:
 #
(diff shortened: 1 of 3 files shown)


st0012:
@dgutov sorry that we need to revert your PR for now as it'd break Ruby CI and block other Ruby committer's work. This is because as part of its CI suites, `debug` (bundled gem) will be tested against the latest `ruby/ruby`, which includes IRB commits synced there.

And currently `debug`'s tests [rely on](https://github.com/ruby/debug/blob/master/test/support/console_test_case.rb#L222) the `TERM=dumb` behaviour before you PR. But unlike IRB, `debug` is only synced to `ruby/ruby` when a release happens. 

This means that to prevent Ruby CI from failing, we will need to:

1. patch `debug` to adjust for your change
2. And then cut a release with it

But unfortunately we don't have control over the `debug` project, so we need to coordinate with its maintainer.

st0012 (approved):

dgutov:
That's too bad, but ok.

Could you please Cc me when you file the corresponding issue or PR to the `debug` repo?

tompng:
@dgutov Thank you for fixing #907, and sorry for making confusion.

With the workaround #943, now we don't need to revert.
I've opened a pull request to debug https://github.com/ruby/debug/pull/1100 that fixes the test to pass without the workaround.

dgutov:
Thanks.

The new workaround is not ideal (adding an extra newline before the prompt is a little odd), but it's still better than Irb being entirely non-functional. Looking forward to having it reverted after all test suites are updated.

PR #947, opened 2024-05-03
Use flag instead of caller for `debug`'s binding.irb check

### LINKED ISSUE #946: Can't enter a debug session with irb 1.13
## Description

I'm getting an error message saying I can't use `debug`. Seems like version 1.12 is working fine, tho.

```
    15:
 => 16:       binding.irb
    17:       { min: min.ceil(2), max: max.floor(2) }
    
>> debug
Debugging commands are only available when IRB is started with binding.irb
=> nil
```

## Result of irb_info

```
>> irb_info
Ruby version: 3.3.1
IRB version: irb 1.13.0 (2024-05-01)
InputMethod: RelineInputMethod with Reline 0.5.4
Completion: Tab Complete, RegexpCompletor
.irbrc paths: /Users/filipe/.irbrc, /Users/filipe/src/github.com/core/.irbrc
RUBY_PLATFORM: arm64-darwin23
LC_ALL env: en_US.UTF-8
East Asian Ambiguous Width: 1
```

## Terminal Emulator

What's your terminal emulator?

## Setting Files

st0012:
Although #946 was caused by [file loading order change](https://github.com/ruby/irb/issues/946#issuecomment-2093165300), I believe it's better not to rely on caller location for the `binding.irb` check. So this PR flags `Irb` instances that are created from `binding.irb`, and use that info to inform `debug` commands.

Closes #946

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -931,8 +931,11 @@ module IRB
     # The lexer used by this irb session
     attr_accessor :scanner
 
+    attr_reader :from_binding
+
     # Creates a new irb session
-    def initialize(workspace = nil, input_method = nil)
+    def initialize(workspace = nil, input_method = nil, from_binding: false)
+      @from_binding = from_binding
       @context = Context.new(self, workspace, input_method)
       @context.workspace.load_helper_methods_to_main
       @signal_status = :IN_IRB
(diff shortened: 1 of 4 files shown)


tompng (approved):
:+1:

PR #948, opened 2024-05-03
Improve help message for no meta commands

st0012:
1. Add placeholder values for both command category and description
2. Update help command's output to give different types of categories
   more explicit ordering


tompng (approved):
Looks good :+1:

PR #949, opened 2024-05-04
Change debug test workaround only enabled when output is tty

tompng:
In https://github.com/ruby/irb/pull/943, we add a workaround `puts` but it caused IRB binding spec failure in https://github.com/ruby/ruby/pull/10712.

To minimize the impact and to avoid test failure in other project, I think it is better to restrict this workaround only to tty environment.
For other projects that runs integration test using `IO.popen` in other projects won't be affected by the workaround.
The workaround still works for debug because debug uses `PTY.open`, not `IO.popen`.

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -67,7 +67,7 @@ module IRB
     #
     # See IO#gets for more information.
     def gets
-      puts
+      puts if @stdout.tty? # workaround for debug compatibility test
       print @prompt
       line = @stdin.gets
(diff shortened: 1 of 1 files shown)


st0012 (approved):

dgutov:
@tompng FWIW, the original `puts` workaround added a superfluous newline in Emacs' inf-ruby before every prompt (to remind: StdinInputMethod is used there since #907), and this didn't fix it (`@stdout.tty?` returns true).

PR #953, opened 2024-05-06
Add a new initialization step to validate IRB.conf's values

### LINKED ISSUE #774: EVAL_HISTORY is busted
with `IRB.conf[:EVAL_HISTORY] = true` everything is busted:

```
9972 % irb
>> 1  + 2
/Users/ryan.davis/.rubies/ruby-3.2.2/lib/ruby/3.2.0/irb/ext/history.rb:127:in `>': comparison of Integer with true failed (ArgumentError)

      @contents.shift if @size != 0 && @contents.size > @size
                                                        ^^^^^
	from /Users/ryan.davis/.rubies/ruby-3.2.2/lib/ruby/3.2.0/irb/ext/history.rb:127:in `push'
	from /Users/ryan.davis/.rubies/ruby-3.2.2/lib/ruby/3.2.0/irb/ext/history.rb:26:in `set_last_value'
```

even `irb_info`

current .irbrc:
```
IRB.conf[:PROMPT_MODE]      = :SIMPLE
IRB.conf[:USE_AUTOCOMPLETE] = false
IRB.conf[:USE_COLORIZE]     = false

# maybe?

IRB.conf[:EVAL_HISTORY]     = true
```

st0012:
Currently, users can only find out that they have set a wrong value for IRB configs when the value is used, with opaque error messages like "comparison of Integer with true failed (TypeError)".

This commit adds a new initialization step to validate the values of some IRB configs, so that users can find out about the wrong values during the initialization of IRB.

Closes #774

### DIFF
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -73,11 +73,12 @@ module IRB
 
       self.prompt_mode = IRB.conf[:PROMPT_MODE]
 
-      if IRB.conf[:SINGLE_IRB] or !defined?(IRB::JobManager)
-        @irb_name = IRB.conf[:IRB_NAME]
-      else
-        @irb_name = IRB.conf[:IRB_NAME]+"#"+IRB.JobManager.n_jobs.to_s
+      @irb_name = IRB.conf[:IRB_NAME]
+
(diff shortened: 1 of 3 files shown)


tompng (approved):
Added error message looks great :+1:

PR #954, opened 2024-05-08
Enhance regexp to account for prism error messages

kddnewton:
Prism changes this error message slightly to provide more information. Currently:

```
$ ./ruby -c -e 'foo.;'
-e:1: syntax error, unexpected ';'
foo.;
./ruby: compile error (SyntaxError)
$ ./ruby --parser=prism -c -e 'foo.;'
./ruby: warning: The compiler based on the Prism parser is currently experimental and compatibility with the compiler based on parse.y is not yet complete. Please report any issues you find on the `ruby/prism` issue tracker.
./ruby: -e:1: syntax error found (SyntaxError)
> 1 | foo.;
    |     ^ unexpected ';'; expecting a message to send to the receiver
  2 | 
```

This is the only regexp that needs to change to get all of the CRuby/IRB tests passing with `--parser=prism`.

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -230,7 +230,7 @@ module IRB
           #   example:
           #     '
           return :recoverable_error
-        when /syntax error, unexpected end-of-input/
+        when /unexpected end-of-input/
           # "syntax error, unexpected end-of-input, expecting keyword_end"
           #
           #   example:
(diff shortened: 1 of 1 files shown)


lib/irb/ruby-lex.rb
```diff
@@ -250,7 +250,7 @@ def check_code_syntax(code, local_variables:)
           #   example:
           #     end
           return :unrecoverable_error
-        when /syntax error, unexpected '\.'/
+        when /(syntax error,|\^~*(\e\[m)?) unexpected '\.'/
```
st0012, line 253:
```suggestion
        # Prism's `SyntaxError#message` can contain escape sequences
        when /(syntax error,|\^~*(\e\[m)?) unexpected '\.'/
```

st0012 (approved):
Let's add a comment to explain the pattern for escape sequence?

st0012 (approved):

Issue #958, opened 2024-05-21
undefined method `extract_ruby_args=' for module IRB::Command (NoMethodError)


rebuilt:
## Description

In a rails console, running the following command fails `measure = Measure.first`
But the following succeeds `m = Measure.first`

The error message of the `measure = Measure.first` command is 
```
(eval at /home/nelson/.asdf/installs/ruby/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb/command/internal_helpers.rb:22):1:in `<main>': undefined method `extract_ruby_args=' for module IRB::Command (NoMethodError)

IRB::Command.extract_ruby_args = Measure.first

Maybe IRB bug!
```

When I revert to using irb version 1.12.0 all commands work correctly, but on versions 1.13.0 and 1.13.1, the command fails as described above.  

I have other models in my application but so far, only measure.rb fails in this way.   I'm still testing to see if the problem affects any other model.  

## Result of irb_info
```
Ruby version: 3.3.1
IRB version: irb 1.13.1 (2024-05-05)
InputMethod: RelineInputMethod with Reline 0.5.7
Completion: Autocomplete, RegexpCompletor
RUBY_PLATFORM: x86_64-linux
LANG env: en_US.UTF-8
East Asian Ambiguous Width: 1
```

## Terminal Emulator

What's your terminal emulator?
Kitty

## Setting Files

Are you using `~/.irbrc` and `~/.inputrc`?
No

rebuilt:
I assume there's a clash with the irb command 
` measure        `measure` enables the mode to measure processing time. `measure :off` disables it.`

tompng:
`measure = 1` is a measure command call with `args = "= 1"`. Related to #803.
The error message is not user friendly so it should be improved.

Earlopain:
I ran into the same thing when trying to assign some source code string to a variable named `source`. I guess this would be because of #824? Not seeing much else in the changelog.

I was curious how python handles this: its repl also has some buildins. The `help` command for example works, as long as you don't reassign it. `exit` as well, which means you can only close with ctrl+d, not with `exit()` anymore. I believe irb should do the same.

st0012:
In #824, we decided commands should always take precedence over variable names or methods (mostly my idea tbf). This is to avoid situations like having `help` in the scope of the `binding.irb` breakpoint unexpectedly disabling the command completely.

But now I think we can do a more sophisticated check to support this precedence order:

1. Local variables defined in the IRB session
[...]

PR #959, opened 2024-05-22
Suppress 'Maybe IRB bug!' message in extract_ruby_args

tompng:
Improve error message of #958

```
irb(main):001> measure = 1
undefined method `extract_ruby_args=' for module IRB::Command (NoMethodError)
IRB::Command.extract_ruby_args = 1
            ^^^^^^^^^^^^^^^^^^^^
Maybe IRB bug!
```
This is not a bug. `"= 1"` is an invalid argument for `measure` command.

```
irb(main):002> measure a + b
undefined local variable or method `a' for main (NameError)
IRB::Command.extract_ruby_args a + b
                               ^
Maybe IRB bug!
```
This is also not a bug.

This pull request improves the error message with less effort.
Note that `IRB::Command::RubyArgsExtractor` is a backward-compatibility layer of command arguments and we are going to reduce using it.

### DIFF
--- a/lib/irb/command/internal_helpers.rb
+++ b/lib/irb/command/internal_helpers.rb
@@ -4,6 +4,8 @@ module IRB
   module Command
     # Internal use only, for default command's backward compatibility.
     module RubyArgsExtractor # :nodoc:
+      class Error < StandardError; end
+
       def unwrap_string_literal(str)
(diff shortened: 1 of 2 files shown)


tompng:
#961 will solve most of the problem.

PR #964, opened 2024-06-05
Add accidentally dropped disable_irb command back

st0012:
Additionally, I sorted `default_commands.rb`'s requires by name so it's a bit easier to find missing commands.


tompng (approved):
:+1:

PR #966, opened 2024-06-09
Suppress Ruby warnings in certain backtrace filtering tests

st0012:
Since they're sensitive to the warnings, and the warnings are not relevant to the tests, we can suppress them to keep the tests simple.

This should fix the current test failures on master.

### DIFF
--- a/test/irb/test_irb.rb
+++ b/test/irb/test_irb.rb
@@ -825,6 +825,13 @@ module TestIRB
   end
 
   class BacktraceFilteringTest < TestIRB::IntegrationTestCase
+    def setup
+      super
+      # These tests are sensitive to warnings, so we disable them
+      original_rubyopt = [ENV["RUBYOPT"], @envs["RUBYOPT"]].compact.join(" ")
+      @envs["RUBYOPT"] = original_rubyopt + " -W0"
+    end
+
     def test_backtrace_filtering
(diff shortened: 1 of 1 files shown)


test/irb/test_irb.rb
```diff
@@ -825,6 +825,13 @@ def build_irb
   end
 
   class BacktraceFilteringTest < TestIRB::IntegrationTestCase
+    def setup
+      super
+      # These tests are sensitive to warnings, so we disable them
+      original_rubyopt = ENV["RUBYOPT"] || "" + (@envs["RUBYOPT"] || "")
```
tompng, line 831: Precedence of the operator will be `ENV["RUBYOPT"] || ("" + (@envs["RUBYOPT"] || ""))`

How about this? It will make it chainable in subclasses. (expecting `@envs["RUBYOPT"]` already including `ENV["RUBYOPT"]` if it is not nil)
```ruby
original_rubyopt = @envs["RUBYOPT"] || ENV["RUBYOPT"] || ""
```

st0012: No I think we need to include both values. So probably this instead:

```rb
      original_rubyopt = [ENV["RUBYOPT"], @envs["RUBYOPT"]].compact.join(" ")
```

tompng (approved):

PR #971, opened 2024-06-16
Introduce cd command

st0012:
It's essentially a combination of `pushws` and `popws` commands that are easier to use.

Help message:

```
Usage: cd ([target]|..)

IRB uses a stack of workspaces to keep track of context(s), with `pushws` and `popws` commands to manipulate the stack.
The `cd` command is an attempt to simplify the operation and will be subject to change.

When given:
- an object, cd will use that object as the new context by pushing it onto the workspace stack.
- "..", cd will leave the current context by popping the top workspace off the stack.
- no arguments, cd will move to the top workspace on the stack by popping off all workspaces.

Examples:

  cd Foo
  cd Foo.new
  cd @ivar
  cd ..
  cd
```

Please note that it's not a replica of Pry's `cd` command as it doesn't support the `/` argument, nor complicated navigation syntax.

### DIFF
--- a/lib/irb/command/cd.rb
+++ b/lib/irb/command/cd.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module IRB
+  module Command
+    class CD < Base
+      category "Workspace"
+      description "Move into the given object or leave the current context."
+
+      help_message(<<~HELP)
+        Usage: cd ([target]|..)
+
+        IRB uses a stack of workspaces to keep track of context(s), with `pushws` and `popws` commands to manipulate the stack.
+        The `cd` command is an attempt to simplify the operation and will be subject to change.
+
+        When given:
+        - an object, cd will use that object as the new context by pushing it onto the workspace stack.
+        - "..", cd will leave the current context by popping the top workspace off the stack.
+        - no arguments, cd will move to the top workspace on the stack by popping off all workspaces.
+
+        Examples:
+
+          cd Foo
+          cd Foo.new
+          cd @ivar
+          cd ..
+          cd
+      HELP
+
+      def execute(arg)
+        case arg
+        when ".."
+          irb_context.pop_workspace
+        when ""
+          # TODO: decide what workspace commands should be kept, and underlying APIs should look like,
+          # and perhaps add a new API to clear the workspace stack.
+          prev_workspace = irb_context.pop_workspace
+          while prev_workspace
+            prev_workspace = irb_context.pop_workspace
+          end
+        else
+          begin
+            obj = eval(arg, irb_context.workspace.binding)
+            irb_context.push_workspace(obj)
+          rescue StandardError => e
+            warn "Error: #{e}"
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/irb/default_commands.rb
+++ b/lib/irb/default_commands.rb
@@ -5,6 +5,7 @@ require_relative "command/internal_helpers"
 require_relative "command/backtrace"
 require_relative "command/break"
 require_relative "command/catch"
+require_relative "command/cd"
 require_relative "command/chws"
 require_relative "command/context"
 require_relative "command/continue"
@@ -240,6 +241,8 @@ module IRB
     _register_with_aliases(:irb_disable_irb, Command::DisableIrb,
       [:disable_irb, NO_OVERRIDE]
     )
+
+    register(:cd, Command::CD)
   end
 
   ExtendCommand = Command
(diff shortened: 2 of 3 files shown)


lib/irb/command/cd.rb
```diff
@@ -0,0 +1,39 @@
...
+          cd Foo.new
+          cd @ivar
+          cd ..
+          cd
+      HELP
+
+      def execute(arg)
+        case arg
+        when "..", ""
+          irb_context.pop_workspace
```
sambostock, line 27: In the terminal, `cd` (without arguments) goes to the user's home, unlike `cd ..`. Rather than combine the two, should we only support `cd ..`, and leave `cd` as a possible future feature? That was, it wouldn't be a breaking change to introduce whatever the closest behavior would be to "return home".

st0012: Good catch, I thought in Pry it goes back a level, but it goes to the top-level instead. Will update 👍

lib/irb/command/cd.rb
```diff
@@ -0,0 +1,47 @@
...
+
+          cd Foo
+          cd Foo.new
+          cd @ivar
+          cd ..
+          cd
+      HELP
+
+      def execute(arg)
+        case arg
```
kmcphillips, line 26: How about the `cd -` case? As in, go to the previous but also push the current back on the stack.

lib/irb/command/cd.rb
```diff
@@ -0,0 +1,57 @@
...
+      HELP
+
+      def execute(arg)
+        previous_workspace = irb_context.workspace
+
+        case arg.strip
+        when ".."
+          irb_context.pop_workspace
+        when "-"
+          irb_context.replace_workspace(irb_context.previous_workspace)
```
st0012, line 36: To support this, we need a way to remember popped workspaces too. And because commands are not memoized, we can't use ivars here. So I needed to add a new `previous_workspace` attribute to `Context` just for this feature. Additionally, it will not work with other commands like `pushws` or `chws` without adding further complexities.

While I understand its use cases, I personally don't use it. So either keeping or dropping it is fine for me. Thoughts? @ruby/irb-reline

tompng: I think it's worth adding, or maybe we can add it later. Not particular about keeping or dropping it at the first step.

Since `cd` has a stack that represents directory-like model, I think `cd -` should restore the whole stack, not just the current workspace.
```
cd 1  # 1
cd 2  # 1/2
cd 3  # 1/2/3
cd .. # 1/2
cd -  # 1/2/3
cd    # main
cd -  # 1/2/3
cd .. # 1/2
```

And we need to consider `irb_context.previous_workspace==nil` case.
```
$ irb
irb(main):001> cd -
```

brandondrew:
```
- "..", cd will leave the current context, moving back to the previous context.
```

I believe this is incorrect, based on both the assumption that this works similarly to `cd ..` in a shell, and based on the code shown above (in a previous comment), which shows it 'popping' contexts (which normally means moving up a level, unless the 'popping' metaphor is also very different in `irb`.)  I've copied that code here:

```ruby
        when ".."
          irb_context.pop_workspace
```
Popping does not necessarily take you to the _previous_ workspace.  You could be where you are because of using `cd -`, which case the next level up might not be the previous workspace.

alpaca-tc:
Thanks for the great PR 💛 
I hope this PR will be merged because I also used the `cd` command in my [.irbrc](https://gist.github.com/alpaca-tc/769e670eeacbe2aeb0c621d8df1316a1).

Currently, this PR doesn't support `cd ../..` like pry.
I would like that feature, do you have any plans to add this feature in the future?

lib/irb/command/cd.rb
st0012: That's a good point. I'll update my implementation 👍

st0012:
> Popping does not necessarily take you to the previous workspace. You could be where you are because of using cd -, in which case the next level up might not be the previous workspace.

It'll be the previous workspace on the same workspace stack. In this context, previous means the previous value that's been added to the stack (array). I know it's not perfectly accurate, but I can't think of better description for it and I'm open to better ideas 🙂 

>  I would like this feature. Do you have any plans to add it in the future?

Unfortunately I don't plan to introduce nested syntax atm, such as `cd @x/@y` like Pry does. In the similar sense, I won't introduce multi-level exits yet.

st0012:
I couldn't 100% replicate Prys' `cd -` behaviour without making huge changes, which is against my goal to build a nicer interface based on the **current** implementation. Even if I were allowed to cut a few corners and deliver a simpler version of it, I wouldn't know where to cut and what to keep as I personally am not a user 😅 
So I decided to drop it from the current PR and ship what we have first. If you really want to see `cd -`, please open an issue and describe the desire behaviour **in details**. Thanks 🙂

tompng (approved):

PR #973, opened 2024-06-17
Return only commands when completing help command's argument

st0012:
The command only takes command names as arguments, so we should only return command names as candidates.

This will help users find a command faster as completion will be another useful hint too.

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -33,6 +33,8 @@ module IRB
       yield
     ]
 
+    HELP_COMMAND_PREPOSING = /\Ahelp\s+/
+
     def completion_candidates(preposing, target, postposing, bind:)
       raise NotImplementedError
     end
@@ -86,8 +88,8 @@ module IRB
         )
     end
 
-    def command_completions(preposing, target)
-      if preposing.empty? && !target.empty?
+    def command_candidates(target)
+      if !target.empty?
         IRB::Command.command_names.select { _1.start_with?(target) }
       else
         []
@@ -111,8 +113,18 @@ module IRB
     end
 
     def completion_candidates(preposing, target, _postposing, bind:)
-      commands = command_completions(preposing, target)
+      # When completing the argument of `help` command, only commands should be candidates
+      return command_candidates(target) if preposing.match?(HELP_COMMAND_PREPOSING)
+
+      commands = if preposing.empty?
+        command_candidates(target)
+      # It doesn't make sense to propose commands with other preposing
+      else
+        []
+      end
+
       result = ReplTypeCompletor.analyze(preposing + target, binding: bind, filename: @context.irb_path)
+
       return commands unless result
 
       commands | result.completion_candidates.map { target + _1 }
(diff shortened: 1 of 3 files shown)


lib/irb/completion.rb
```diff
@@ -111,8 +113,17 @@ def inspect
     end
 
     def completion_candidates(preposing, target, _postposing, bind:)
-      commands = command_completions(preposing, target)
+      preposing ||= ''
```
st0012, line 116: @tompng Do you know when `proposing` will be `nil`? It looks like `ReplTypeCompletor` simply [assumes](https://github.com/ruby/repl_type_completor/blob/4e273330181bf096f71016937f52cddacfa77b7a/lib/repl_type_completor.rb#L33) it'd always be present?

tompng: I think it wont be nil anymore.
In the old code, preposing and postposing might be nil
```ruby
IRB::InputCompletor::CompletionProc = ->(target, preposing=nil, postposing=nil){}
Readline.completion_proc = IRB::InputCompletor::CompletionProc # preposing and postposing are always nil
Reline.completion_proc = IRB::InputCompletor::CompletionProc # preposing and postposing are always string
```
But now, empty string is passed to `completion_candidates` in such case.
```ruby
Readline.completion_proc = ->(target) {
  bind = IRB.conf[:MAIN_CONTEXT].workspace.binding
  @completor.completion_candidates('', target, '', bind: bind)
}
```

In the last few lines in this file, `InputCompletor::CompletionProc` is the only place where nil might be passed to `RegexpCompletor#completion_candidates`, but it is not used from IRB's code.
```ruby
module InputCompletor
  CompletionProc = ->(target, preposing = nil, postposing = nil) {
    regexp_completor.completion_candidates(preposing, target, postposing, bind: IRB.conf[:MAIN_CONTEXT].workspace.binding)
   # maybe `(preposing || '', target, postposing || '', bind: ...)` is better
  }
end
deprecate_constant :InputCompletor
```

lib/irb/completion.rb
```diff
@@ -33,6 +33,8 @@ def defined? do
       yield
     ]
 
+    HELP_COMMAND_PREPOSING = /\Ahelp\s+/
```
tompng, line 36: I think each command can have a completion and colorization for command arguments (as an internal api).
Do you think we should do it now or not or maybe later?
I'm thinking of something like this
```ruby
if command? && argument_part?
  # pass completor so that command can fallback to completor if the arg part is a ruby code
  return command_instance._internal_complete(command_arg, completor)
end
```

st0012: IMO, we should first finish completion support for signature arguments too, like `show_source` and `edit`'s, before designing any public API. This is to make sure we know what we want and don't want to expose. I also don't see too strong of a demand for such feature yet, so I think we can wait a bit.

lib/irb/completion.rb
st0012: I'll change `CompletionProc` then, thanks!

tompng (approved):

PR #983, opened 2024-07-20
Colorize command input

tompng:
Command colorization for `$`, alias of show_source was not good.
This pull request improves it.

- command name part: bold
- command arg part: colorized as ruby code

## before and after
command_color

Colorization for some command that uses `extract_ruby_args` mostly for backward compatibility gets worse, but we're going to reduce using extract_ruby_args like #962.
command_color_cons

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -270,7 +270,13 @@ module IRB
           proc do |output, complete: |
             next unless IRB::Color.colorable?
             lvars = IRB.CurrentContext&.local_variables || []
-            IRB::Color.colorize_code(output, complete: complete, local_variables: lvars)
+            if IRB.CurrentContext&.parse_command(output)
+              name, sep, arg = output.split(/(\s+)/, 2)
+              arg = IRB::Color.colorize_code(arg, complete: complete, local_variables: lvars)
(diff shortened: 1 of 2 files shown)


lib/irb/input-method.rb
```diff
@@ -270,7 +270,13 @@ def initialize(completor)
           proc do |output, complete: |
             next unless IRB::Color.colorable?
             lvars = IRB.CurrentContext&.local_variables || []
-            IRB::Color.colorize_code(output, complete: complete, local_variables: lvars)
+            if IRB.CurrentContext&.irb&.command?(output)
```
st0012, line 273: Since `command?` calls `parse_command` underneath, I think we can directly call `parse_command` to get a more constructive `Statement` object here instead?

(I also opened #993 to move `parse_command` to `Context`).

st0012: I wasn't clear (and also not completely correct) about the idea: maybe we can use `Irb#build_statement` to get a `Statement` back, and put most of the block into a method like `Statement#colorized_output`.

BUT, I think it may actually be a wrong coupling to make `Statement` handle display logic as well + it now needs to take in local variables too. So let's drop it 🙈

st0012 (approved):

PR #985, opened 2024-08-04
Remove Ruby version checks

RicardoTrindade:
Minor cleanup from test files where it's being checked if the Ruby version is lower than 2.7. Since the gem requires 2.7+ by default I'd say these can be removed


tompng (approved):
Thank you :+1: Looks good!

PR #988, opened 2024-08-14
Always use alternate sceen on alt-d

tompng:
Fixes screen corruption when `PAGER=more` is used.

When ALT+D is pressed while showing document dialog, document in shown in full screen using pager.
IRB assumes that the pager switches to alternate screen, turns off alternate screen at exit, and restore IRB's screen.

Some pager like `more` does not do that. Screen corrupts. We need to switch to alternate screen manually.

irb_with_pager_eq_more
To reproduce:
`PAGER=more irb` (make sure `more` is not aliased to `less`)
`PAGER="sh -c 'cat;sleep 1'" irb`

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -361,9 +361,15 @@ module IRB
           if show_easter_egg
             IRB.__send__(:easter_egg)
           else
+            # RDoc::RI::Driver#display_names uses pager command internally.
+            # Some pager command like `more` doesn't use alternate screen
+            # so we need to turn on and off alternate screen manually.
             begin
+              print "\e[?1049h"
               driver.display_names([name])
             rescue RDoc::RI::Driver::NotFoundError
+            ensure
+              print "\e[?1049l"
             end
           end
         end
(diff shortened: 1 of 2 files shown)


lib/irb/input-method.rb
```diff
@@ -361,9 +361,15 @@ def show_doc_dialog_proc
           if show_easter_egg
             IRB.__send__(:easter_egg)
           else
+            # RDoc::RI::Driver#display_names uses pager command internally.
+            # Some pager command like `more` doesn't use alternate screen
+            # so we need to turn on and off alternate screen manually.
             begin
+              print "\e[?1049h"
```
st0012, line 368: Is this something we want `ri` to do most of the time as well?

st0012 (commented):
Should we wrap this around IRB's pager implementation too?

tompng:
IRB's pager and in ri command, I think we don't need to use this escape sequence.
The situation that needs force enable alternate screen mode and restore screen only happens when called from show_doc_dialog_proc.

PR #993, opened 2024-08-25
Move parse_command method to Context

st0012:
Since Context dictates whether a line is a command or an expression, moving the parse_command method to Context makes the relationship more explicit.


tompng (approved):
Looks good :+1:

PR #999, opened 2024-09-05
Suppress warning when loading rdoc

Earlopain:
This actually starts to warn on Ruby 3.3.5 now because that was backported. Just rescuing the LoadError is not enough, since the warning is emitted anyways.

This is not ideal, see https://bugs.ruby-lang.org/issues/20714#note-2, but I'm not aware of a different solution.
`reline` does the same thing for `fiddle`: https://github.com/ruby/reline/blob/c90f08f7e308d2f1cdd7cfaf9939fe45ce546fd2/lib/reline/terminfo.rb#L1-L15

Without this, a simple `bundle exec irb` on Ruby 3.3.5 shows this:
> rdoc was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
> You can add rdoc to your Gemfile or gemspec to silence this warning.

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -288,9 +288,12 @@ module IRB
 
       if IRB.conf[:USE_AUTOCOMPLETE]
         begin
+          verbose, $VERBOSE = $VERBOSE, nil
           require 'rdoc'
           Reline.add_dialog_proc(:show_doc, show_doc_dialog_proc, Reline::DEFAULT_DIALOG_CONTEXT)
         rescue LoadError
+        ensure
+          $VERBOSE = verbose
         end
       end
     end
@@ -321,6 +324,7 @@ module IRB
       return @rdoc_ri_driver if defined?(@rdoc_ri_driver)
 
       begin
+        verbose, $VERBOSE = $VERBOSE, nil
         require 'rdoc'
       rescue LoadError
         @rdoc_ri_driver = nil
@@ -328,6 +332,8 @@ module IRB
         options = {}
         options[:extra_doc_dirs] = IRB.conf[:EXTRA_DOC_DIRS] unless IRB.conf[:EXTRA_DOC_DIRS].empty?
         @rdoc_ri_driver = RDoc::RI::Driver.new(options)
+      ensure
+        $VERBOSE = verbose
       end
     end
 


hsbt:
I consider to suppress that warning if target gem like irb already has dependency like rdoc explicitly.

Earlopain:
Oh, I didn't realize it already depends on rdoc. I agree, this PR doesn't make much sense then. I will close.

hsbt:
BTW: I couldn't reproduce this.

```
$ cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem "irb"
$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    io-console (0.7.2)
    irb (1.14.0)
      rdoc (>= 4.0.0)
      reline (>= 0.4.2)
    psych (5.1.2)
      stringio
    rdoc (6.7.0)
      psych (>= 4.0.0)
    reline (0.5.9)
      io-console (~> 0.5)
    stringio (3.1.1)

PLATFORMS
  arm64-darwin-24
  ruby

DEPENDENCIES
  irb

BUNDLED WITH
   2.5.18
$ bundle exec irb
>>
```

This is the intentional behavior. Because irb already has rdoc dependency. If irb is invoked under the bundler, irb will load rdoc from GEM_PATH, not ruby installation path. So, `bundled_gems.rb` will not warn that case.

Earlopain:
You need a gemfile that doesn't depend on on irb/rdoc in some way:
```
$ cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem "asciidoctor"
$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    asciidoctor (2.0.23)

PLATFORMS
  ruby
  x86_64-linux

DEPENDENCIES
  asciidoctor

BUNDLED WITH
   2.5.16

$ bundle exec irb
/home/user/.rbenv/versions/3.3.5/lib/ruby/3.3.0/reline.rb:9: warning: fiddle was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add fiddle to your Gemfile or gemspec to silence this warning.
rdoc was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add rdoc to your Gemfile or gemspec to silence this warning.
```

hsbt:
You should add `irb` into Gemfile if you want to use irb under the bundler.

Earlopain:
I guess you want to migrate irb to bundled gems too so in that regard that makes sense.

Regardless of my opinion of doing that, `irb` itself does not warn. Should I really be made to do this right now?

PR #1001, opened 2024-09-06
Make rendering test faster using updated yamatanooroti

tompng:
Makes test_rendering faster. Yamatanooroti's wait time is reduced from 0.1 sec to 0.01 sec
`test_yamatanooroti(vterm-yamatanooroti head)`: 37.58 sec → 10.72 sec

## startup_message
We don't need `puts 'start IRB'`. It does not mean IRB is ready for input. We should check the prompt with `startup_message: /irb\(main\)/`

## close
It does not only close_input but read few bytes and terminates process immediately.
Calling close before assert_screen sometimes terminates process before reading the last output.

### DIFF
--- a/test/irb/yamatanooroti/test_rendering.rb
+++ b/test/irb/yamatanooroti/test_rendering.rb
@@ -39,49 +39,44 @@ class IRB::RenderingTest < Yamatanooroti::TestCase
   end
 
   def test_launch
-    write_irbrc <<~'LINES'
-      puts 'start IRB'
-    LINES
-    start_terminal(25, 80, %W{ruby -I#{@pwd}/lib #{@pwd}/exe/irb}, startup_message: 'start IRB')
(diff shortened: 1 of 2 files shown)


test/irb/yamatanooroti/test_rendering.rb
```diff
@@ -311,154 +282,129 @@ def test_autocomplete_with_showdoc_in_gaps_on_narrow_screen_left
...
     write("'a' * 80 * 11\n")
     write("'foo' + 'bar'\n") # eval something to make sure IRB resumes
-    close
 
-    screen = result.join("\n").sub(/\n*\z/, "\n")
-    assert_match(/(a{80}\n){8}/, screen)
+    assert_screen(/(a{80}\n){8}/)
     # because pager is invoked, foobar will not be evaluated
-    assert_not_match(/foobar/, screen)
+    assert_screen(/\A(?!foobar)/)
```
tompng, line 390: Yamatanooroti does not have `assert_not_screen`. We need to use negative lookahead to use a retryable assertion.

st0012 (approved):

PR #1004, opened 2024-09-11
Remove KEYWORD_ALIASES which handled special alias name of irb_break irb_catch and irb_next command

tompng:
This comment is outdated. We don't need KEYWORD_ALIASES anymore.
```ruby
# because all input will eventually be evaluated as Ruby code,
# command names that conflict with Ruby keywords need special workaround
# we can remove them once we implemented a better command system for IRB
KEYWORD_ALIASES = { ... }
```

## [:catch, OVERRIDE_PRIVATE_ONLY]
Just like `exit` and `exit!`, `catch` is already defined as a private method of Object.
IRB's command usually sets OVERRIDE_PRIVATE_ONLY in this case.

## [:next , OVERRIDE_ALL], [:break , OVERRIDE_ALL]
both `next a` `break a` are always syntax error. These input should be always treated as a command even if a method with the same name is defined.

### DIFF
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -150,24 +150,13 @@ module IRB
       end
 
       @user_aliases = IRB.conf[:COMMAND_ALIASES].dup
-      @command_aliases = @user_aliases.merge(KEYWORD_ALIASES)
+      @command_aliases = @user_aliases.dup
     end
 
     private def term_interactive?
       return true if ENV['TEST_IRB_FORCE_INTERACTIVE']
       STDIN.tty? && ENV['TERM'] != 'dumb'
     end
-
-    # because all input will eventually be evaluated as Ruby code,
-    # command names that conflict with Ruby keywords need special workaround
-    # we can remove them once we implemented a better command system for IRB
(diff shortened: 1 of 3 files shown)


lib/irb/context.rb
```diff
@@ -150,25 +150,14 @@ def initialize(irb, workspace = nil, input_method = nil)
       end
 
       @user_aliases = IRB.conf[:COMMAND_ALIASES].dup
```
st0012, line 152:
```suggestion
```

lib/irb/context.rb
```diff
@@ -150,25 +150,14 @@ def initialize(irb, workspace = nil, input_method = nil)
...
-      @command_aliases = @user_aliases.merge(KEYWORD_ALIASES)
+      @command_aliases = @user_aliases.dup
```
st0012, line 153: I think `@user_aliases` is not referenced anymore after this change:

```suggestion
      @command_aliases = IRB.conf[:COMMAND_ALIASES].dup
```

st0012 (approved):
Good catch 👍 Just a small suggestion.

PR #1010, opened 2024-09-30
Change default completor from regexp to type-completor if RUBY_VERSION>=3.4

tompng:
In Ruby >= 3.4, ReplTypeCompletor is a bundled gem, installed by default.
This pull request changes the default completor to type-completor if `RUBY_VERSION >= 3.4`.
If ReplTypeCompletor is not available (not installed or not added to Gemfile), fallbacks to regexp-baased completion.
The default isstill :regexp if `RUBY_VERSION < 3.4`

## Run with bundler
With bundler and when repl_type_completor is not added to Gemfile, IRB will use regexp-based completion.
IRB shows warning message only If `--type-completor` or `IRB.conf[:COMPLETOR] = :type` is explicitly specified.

```
$ ruby -v
ruby 3.4.0dev

$ irb
irb(main):001> irb_info
(TypeCompletor)

$ echo "gem 'irb', path: 'path/to/irb'" > Gemfile

$ bundle exec irb
(no waarning)
irb(main):001> irb_info
(RegexpCompletor)

$ bundle exec irb --type-completor
TypeCompletor requires `gem repl_type_completor`: cannot load such file -- repl_type_completor
irb(main):001> irb_info
(RegexpCompletor)
```

## Restriction with rdbg

ReplTypeCompletor uses background thread to load RBS type information which could take time. 
Debug gem stops all other threads, so if IRB enters debug mode before loading RBS completes, type completor can't use type information.
Even without type information, type completor can still complete as good or better than regexp-based completion.

### DIFF
--- a/lib/irb/context.rb
+++ b/lib/irb/context.rb
@@ -176,11 +176,17 @@ module IRB
 
     private def build_completor
       completor_type = IRB.conf[:COMPLETOR]
+
+      # Gem repl_type_completor is added to bundled gems in Ruby 3.4.
+      # Use :type as default completor only in Ruby 3.4 or later.
+      verbose = !!completor_type
+      completor_type ||= RUBY_VERSION >= '3.4' ? :type : :regexp
+
       case completor_type
       when :regexp
         return RegexpCompletor.new
       when :type
-        completor = build_type_completor
+        completor = build_type_completor(verbose: verbose)
         return completor if completor
       else
         warn "Invalid value for IRB.conf[:COMPLETOR]: #{completor_type}"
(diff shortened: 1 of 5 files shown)


st0012:
Since `repl_type_completor` only has ~64k downloads atm, this means almost all IRB sessions after this change will trigger a `LoadError` unnecessarily. And at the same time, this change isn't actually making more people use `repl_type_completor`. It simply saves the existing users the need to configure it.

IMO the right solution here is to propose making `repl_type_completor` a bundled gem. And only until that happens, we make it a default for IRB.

tompng:
ReplTypeCompletor is added to bundled gem now :tada:  I updated the pull request description. 

> with a Ruby version check

Looks nice, added

st0012 (approved):
Let's give it a try 🎉

ima1zumi (approved):
LGTM! 🚀

PR #1012, opened 2024-10-08
Document infinite history

eval:
...as introduced in 824473e8

Setting `IRB.conf[:SAVE_HISTORY] = 0` won't store anything.


tompng (approved):
:+1:

PR #1013, opened 2024-10-08
History refactors

eval:
As part of #1012 I started looking into history some more and came up with some refactors around saving/loading history and the settings `SAVE_HISTORY` and `HISTORY_FILE`.

The main change is that this PR bundles history logic in a separate module `IRB::History`, with helpers like `save_history` (`IRB.conf[:SAVE_HISTORY]` as an int), `save_history?`, `infinite?` and `history_file`. Logic that currently is duplicated and/or not explicitly named as such.  
Furthermore, `IRB::HistorySavingAbility.save_history` is currently quite complex in that there's multiple returns - these are now all in the top of the method with more descriptive names.

### Changes in functionality

- permissions
  Anything that might prevent the file from being saved (folder doesn't exist, lacking permissions to folder or file), now results in a warning.  
  Currently, there's only a warning when the folder does not exist. Lacking folder-permissions (to write a new file) will result in an exception. Lacking folder/file permissions for an existing file currently result in silently not saving.

I did quite some git spelunking to see what issues lay behind the current code, and tested locally different scenarios to ensure that nothing gets deleted.

The PR is split into 2 commits  (save_history and history_file resp. changes) to aid reviewing. Let me know what's preferred for additional changes (additional commits or amending these commits, the latter can be noisy and hard to review). 

### (open) Issues

- should `conf.save_history`. and `conf.history_file` be 'aliases' for resp. `IRB::History.save_history` and `IRB::History.history_file`?   
  Pro: This way a user can more easily inspect what IRB will do, e.g. settings like `IRB.conf[:SAVE_HISTORY] = nil` are turned into `conf.save_history #=> 0`, and showing a clear path when `IRB.conf[:HISTORY_FILE] = nil`.
  Downside is that it might confuse as it's not what a user 'configged'.

PS I saw the hacktoberfest-label on the repo, this PR is not part of that, I'm not participating.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1004,9 +1004,10 @@ module IRB
       prev_context = conf[:MAIN_CONTEXT]
       conf[:MAIN_CONTEXT] = context
 
-      save_history = !in_nested_session && History.save_history? && context.io.support_history_saving?
+      load_history = !in_nested_session && context.io.support_history_saving?
+      save_history = load_history && History.save_history?
 
-      if save_history
+      if load_history
         context.io.load_history
       end
 
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -14,6 +14,7 @@ require_relative "irb/default_commands"
 
 require_relative "irb/ruby-lex"
 require_relative "irb/statement"
+require_relative "irb/history"
 require_relative "irb/input-method"
 require_relative "irb/locale"
 require_relative "irb/color"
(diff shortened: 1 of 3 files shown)


tompng (commented):
Thank you for the pull request. The change looks basically great!

eval:
Thanks for the feedback @tompng ! I made the changes.
I also added an extra commit that detaches loading the history from saving, allowing for readonly history - maybe it warrants a separate CLI-flag to ensure nothing happens to the history-file? I can move it to a separate issue/PR if you prefer.

Let me know if this one is finished and I'll rebase. Thanks again!

tompng (approved):
Looks Good :+1:
Thank you!

PR #1019, opened 2024-10-14
Suppress "literal string will be frozen in the future" warning

tikkss:
Before change:

```console
$ ruby -W -I lib -e 'require "irb"; IRB.setup(nil); IRB::Irb.new.build_statement("1 + 2")'
/Users/zzz/src/github.com/ruby/irb/lib/irb.rb:1135: warning: literal string will be frozen in the future
```

After change:

```console
$ ruby -W -I lib -e 'require "irb"; IRB.setup(nil); IRB::Irb.new.build_statement("1 + 2")'
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1132,7 +1132,7 @@ module IRB
         return Statement::EmptyInput.new
       end
 
-      code.dup.force_encoding(@context.io.encoding)
+      code = code.dup.force_encoding(@context.io.encoding)
       if (command, arg = @context.parse_command(code))
         command_class = Command.load_command(command)
         Statement::Command.new(code, command_class, arg)
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1132,7 +1132,7 @@ module IRB
         return Statement::EmptyInput.new
       end
 
-      code.force_encoding(@context.io.encoding)
+      code.dup.force_encoding(@context.io.encoding)
       if (command, arg = @context.parse_command(code))
         command_class = Command.load_command(command)
         Statement::Command.new(code, command_class, arg)


tompng:
Actually, warning in `IRB::Irb.new.build_statement("1 + 2")` is not a problem because literal string is not passed to build_statement (this is a internal API) in ruby/irb.

However, making build_statement not modify the given argument improves readability and code quality. :+1:

lib/irb.rb
```diff
@@ -1132,7 +1132,7 @@ def build_statement(code)
         return Statement::EmptyInput.new
       end
 
-      code.force_encoding(@context.io.encoding)
+      code.dup.force_encoding(@context.io.encoding)
```
tompng, line 1135: This statement has no effect. We need to change the encoding of `code`. 
maybe `code = code.dup.force_encoidng(...)`

tikkss: Thanks for your review and suggestion! I fixed them at 
37fe153.

By the way, I met the warning in [SciRuby/iruby's test code](https://github.com/SciRuby/iruby/blob/a19e3f3cd2e95553b5ea463fc8e4b35d72c6eef4/test/iruby/backend_test.rb#L13
), [SciRuby/iruby's code](https://github.com/SciRuby/iruby/blob/f341266ea123a0a916b17b148742bb101b6c0a41/lib/iruby/backend.rb#L60).

tompng (approved):
Thank you :+1:

tikkss:
I was deeply impressed by your keynote at RubyKaigi 2024 in person.
I'm happy to collaborate with you on code. Thanks!

PR #1020, opened 2024-10-15
Disable autocompletions by default

jedrekdomanski:
### Breakdown of the change:

- `ENV.fetch("IRB_USE_AUTOCOMPLETE", "false")`: If the environment variable is **not set**, the default value is now `"false"` instead of `"true"`.
- The comparison `!= "false"` ensures that autocomplete is enabled only when `IRB_USE_AUTOCOMPLETE` is explicitly set to something other than `"false"`.

### Behavior after the change:

- **Autocomplete is disabled by default** unless `IRB_USE_AUTOCOMPLETE` is explicitly set to `"true"` or something other than `"false"`.
  
This way, users won’t need to set the environment variable to disable autocomplete; it will be disabled by default.

### Note:
Users can still run `irb` with extra flags to control autocomplete:

- `--autocomplete`: Enables autocomplete for that session.
- `--noautocomplete`: Disables autocomplete for that session.

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -79,7 +79,7 @@ module IRB # :nodoc:
 
     @CONF[:USE_SINGLELINE] = false unless defined?(ReadlineInputMethod)
     @CONF[:USE_COLORIZE] = (nc = ENV['NO_COLOR']).nil? || nc.empty?
-    @CONF[:USE_AUTOCOMPLETE] = ENV.fetch("IRB_USE_AUTOCOMPLETE", "true") != "false"
+    @CONF[:USE_AUTOCOMPLETE] = ENV.fetch("IRB_USE_AUTOCOMPLETE", "false") != "false"
     @CONF[:COMPLETOR] = ENV["IRB_COMPLETOR"]&.to_sym
     @CONF[:INSPECT_MODE] = true
     @CONF[:USE_TRACER] = false
--- a/test/irb/test_context.rb
+++ b/test/irb/test_context.rb
@@ -215,7 +215,7 @@ module TestIRB
     end
 
     def test_default_config
-      assert_equal(true, @context.use_autocomplete?)
+      assert_equal(false, @context.use_autocomplete?)
     end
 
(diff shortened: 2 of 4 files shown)


st0012:
Why should autocomplete be disabled by default? Please list specific issues that you think are outweighing the benefits to help us understand the root problems.

jedrekdomanski:
> Why should autocomplete be disabled by default? Please list specific issues that you think are outweighing the benefits to help us understand the root problems.

The current configuration requires users to either create an .irbrc file on every server and manually disable autocomplete with:
```
IRB.conf[:USE_AUTOCOMPLETE] = false
```
in every project, or explicitly set the IRB_USE_AUTOCOMPLETE environment variable on each server across all projects. This can become cumbersome to manage, especially in environments with multiple servers or projects.

Additionally, the autocomplete feature has a significant performance impact. It tends to be slow, which negatively affects the user experience when typing commands in the console. The prompt often displays all possible methods and suggestions, which can drastically slow down typing and execution. The autocomplete UI itself is invasive and distracting, which detracts from a smooth console workflow.

There are reports of dissatisfaction with this default behavior, as seen in [IRB issue #351](https://github.com/ruby/irb/issues/351). This PR aims to address these concerns by making the behavior more user-friendly and less intrusive by default. Users who prefer autocomplete can still enable it easily, but it won't be forced on everyone, improving overall usability.

st0012:
The problems mentioned in #351 have mostly been addressed in both IRB and Reline, such as:
- The completion performance has dramatically improved
- The color is now black and white by default
- The dropdown doesn't aggressively push up the input line

Unless you can prove or convince us that most of the Ruby developers using the latest IRB and Reline are still suffering the problem and actively seeking a way to disable autocompletion, turning it off by default is a massive breaking change that'd require many users to enable it manually. Therefore, we will not make this decision lightly.

If you're seeing those issues because you're not using the latest IRB and Reline (e.g. using the default in Ruby 3.1), then this PR won't help it anyway.

PR #1022, opened 2024-10-19
Improve history test's encoding setting

monkeyWzr:
* Set stdio's encoding when changing default encoding in test_history_different_encodings 
  * This also fix the failed tests for https://github.com/ruby/reline/pull/765
* Reset `IRB.conf` after tests completed


tompng (approved):
:+1:

PR #1031, opened 2024-11-14
Don't use delegator to install helper methods to main object

tompng:
Delegator was needed to install command methods to frozen objects, but we don't do it anymore.

Dropping delegator will fix this weird behavior.
```ruby
irb(main):001> cd 1
irb(1):002> self
=> 1
irb(1):003> frozen?
=> false
irb(1):004> self.class
=> SimpleDelegator
irb(1):005> { self => 1, 1 => 1 }
=> {1=>1, 1=>1}
```

Helper methods will be unavailable on frozen main object. However, I think it makes sense that frozen object can't have helper methods.
(We can change this by adding helper methods to Kernel instead of main object maybe in the future. It will also completely fix helper method polluting main object.)

This pull request also adds minimal support for BasicObject because the original delegator code was checking BasicObject.
```ruby
case @main
when Object
  use_delegator = @main.frozen?
else
  use_delegator = true # main is BasicObject
end
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -1463,16 +1463,21 @@ module IRB
       end
     end
 
+    def basic_object_safe_main_call(method)
+      main = @context.main
+      Object === main ? main.__send__(method) : Object.instance_method(method).bind_call(main)
+    end
+
     def format_prompt(format, ltype, indent, line_no) # :nodoc:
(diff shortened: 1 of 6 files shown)


st0012 (approved):
Thanks. There are some refactors that I think can be done but I'll give them a try in a separate PR.

PR #1033, opened 2024-11-19
Store method objects in constants

st0012:
It probably won't speed up things significantly, but these are hot paths and we can save a few method calls per completion/input call.

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -137,27 +137,33 @@ module IRB
   end
 
   class RegexpCompletor < BaseCompletor # :nodoc:
+    KERNEL_METHODS = ::Kernel.instance_method(:methods)
+    KERNEL_PRIVATE_METHODS = ::Kernel.instance_method(:private_methods)
+    KERNEL_INSTANCE_VARIABLES = ::Kernel.instance_method(:instance_variables)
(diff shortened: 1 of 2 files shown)


lib/irb/completion.rb
```diff
@@ -137,27 +137,34 @@ def doc_namespace(preposing, matched, _postposing, bind:)
...
         end
 
         def eval_private_methods
-          ::Kernel.instance_method(:private_methods).bind(eval("self")).call
+          KERNEL_PRIVATE_METHODS.bind(eval("self")).call
         end
 
         def eval_instance_variables
-          ::Kernel.instance_method(:instance_variables).bind(eval("self")).call
+          KERNEL_INSTANCE_VARIABLES.bind(eval("self")).call
```
tompng, line 158: We can use `bind_call(receiver)` for the above three.

lib/irb/completion.rb
```diff
@@ -137,27 +137,34 @@ def doc_namespace(preposing, matched, _postposing, bind:)
...
         end
 
         def eval_global_variables
-          ::Kernel.instance_method(:global_variables).bind(eval("self")).call
+          KERNEL_GLOBAL_VARIABLES.bind(eval("self")).call
```
tompng, line 162: `::Kernel.global_variables` is better because this does not need to look into `self`

tompng (approved):
Looks good 👍

Issue #1032, opened 2024-11-18
TruffleRuby CI Fails When Installing `io-console`


ima1zumi:
## Description

Since November 13, TruffleRuby's CI has been failing when attempting to install the `io-console` gem.

```
An error occurred while installing io-console (0.7.2), and Bundler cannot
continue.

In Gemfile:
  irb was resolved to 1.14.1, which depends on
    reline was resolved to 0.5.11, which depends on
      io-console
``` 

https://github.com/ruby/irb/actions/runs/11879710332/job/33101967182

## Full log

```
Run ruby/setup-ruby@v1
Modifying PATH
Downloading Ruby
Extracting  Ruby
Print Ruby version
Installing Bundler
> bundle install
/home/runner/.rubies/truffleruby-head/bin/bundle config --local path /home/runner/work/irb/irb/vendor/bundle
/home/runner/.rubies/truffleruby-head/bin/bundle lock
Fetching https://github.com/ruby/debug.git
Fetching https://github.com/ruby/reline.git
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Writing lockfile to /home/runner/work/irb/irb/Gemfile.lock
Cache key: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-truffleruby-head-wd-/home/runner/work/irb/irb-with--without--only--ABI-3.3.5.2-Gemfile.lock-4b[7](https://github.com/ruby/irb/actions/runs/11879710332/job/33101967182#step:3:8)fce7ba9fe1d364c7716a8a7d4a89352945980e[10](https://github.com/ruby/irb/actions/runs/11879710332/job/33101967182#step:3:12)847840e65da9e01cbcf24
/home/runner/.rubies/truffleruby-head/bin/bundle install --jobs 4
Fetching gem metadata from https://rubygems.org/........
Fetching rake [13](https://github.com/ruby/irb/actions/runs/11879710332/job/33101967182#step:3:16).2.1
Installing rake 13.2.1
Fetching io-console 0.7.2
Fetching stringio 3.1.2
Fetching ast 2.4.2
Fetching json 2.8.2
Installing io-console 0.7.2 with native extensions
Installing stringio 3.1.2 with native extensions
Installing ast 2.4.2
Installing json 2.8.2 with native extensions
Fetching language_server-protocol 3.17.0.3
Installing language_server-protocol 3.17.0.3
Fetching parallel 1.26.3
Installing parallel 1.26.3
[...]

PR #1035, opened 2024-11-19
Extract truffleruby workflow

st0012:
Background: #1032

I understand that constantly testing IRB with TruffleRuby-head is important to help the project cache issues early, but the failed build leads to a lot of noise in the CI workflow.

So in this PR, I hope to achieve 2 things by having a dedicated build configuration for TruffleRuby head:

1. We can very simply disable the CI runs through GH Actions' UI when there's a known issue, and re-enable it easily later, all without changing code.
2. Since TruffleRuby doesn't run most of the steps in the original `build` configuration, extracting it out makes the steps it DOES run more obvious.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
   ruby-versions:
     uses: ruby/actions/.github/workflows/ruby_versions.yml@master
     with:
-      engine: cruby-truffleruby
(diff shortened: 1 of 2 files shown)


st0012:
cc @andrykonchin @eregon

ima1zumi (approved):

PR #1039, opened 2024-11-27
Don't show 'Maybe IRB bug!' in show_source and ls command

tompng:
`ls` command shows `Maybe IRB bug!` when eval fails.

```
irb(main):001> ls foo
undefined local variable or method `foo' for main (NameError)
[backtraces]
Maybe IRB bug!

irb(main):001> ls raise(Exception)
Exception (Exception)
[backtraces]
Maybe IRB bug!

irb(main):001> ls (1.."a")
bad value for range (ArgumentError)
[backtraces]
Maybe IRB bug!
```

Same for show_source
```
irb(main):001> show_source (1.."a").itself
bad value for range (ArgumentError)
Maybe IRB bug!

irb(main):001> show_source raise(Exception).itself
Exception (Exception)
Maybe IRB bug!
```

This pull request fixes it.
```
irb(main):001> ls foo
NameError: undefined local variable or method `foo' for main
irb(main):002> ls raise(Exception,'aa')
Exception: aa
irb(main):003> ls Integer, grep: /#{raise Exception,'bb'}/
Exception: bb
irb(main):004> show_source raise(Exception,'aa').itself
Error: Couldn't locate a definition for raise(Exception,'aa').itself
```

The use of `args, kwargs = ruby_args(arg)` (in legacy ls format) is also removed.

### DIFF
--- a/lib/irb/command/ls.rb
+++ b/lib/irb/command/ls.rb
@@ -11,7 +11,7 @@ module IRB
 
   module Command
     class Ls < Base
-      include RubyArgsExtractor
+      class EvaluationError < StandardError; end
 
       category "Context"
       description "Show methods, constants, and variables."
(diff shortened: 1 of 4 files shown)


st0012 (approved):
Thx 👍

PR #1040, opened 2024-11-29
Show a quick preview of inspect result before pager launch

tompng:
IRB takes about 2 seconds to show the result of `200000.times.to_a`.
```ruby
irb(main):001> 200000.times.to_a
```

In this case, creating the full pretty_print content takes time. Fortunately, there is a way to get the first WINDOW_HEIGHT lines of the content quickly.
This pull request will change IRB to show the first page of the pretty_print result quickly even when pretty_print is slow.

preview_before_pager
First page is displayed 0.1sec after pressing ENTER key.
After two seconds, the whole pretty_print content is ready and pager launches (`:` on the bottom of the line indicates that).

### newline_before_multiline_output
If the inspect takes more than 0.1 sec and result exceeds screen height, IRB will print the first page of the content immediately.
If the inspect of assignment expression result exceeds 1 line, generating rest of the content will be skipped.
In this case, output will be treated as multiline even if the un-generated whole inspect result does not contain newline.

### Considering pretty_print prints SQL log in Rails console
In rails console, pretty printing ActiveRecord::Relation will print a SQL log.

Here is an example of the terminal buffer with terminal height = 5

```
irb(main):001> [User.where(name: 'foo'), User.where(name: 'bar')]
User Load(9.9ms) SELECT * from users where name="foo"
(and other SQL log while retrieving the first page)
=>
[#<User id: 1>,
 #<User id: 2>,
 #<User id: 3>
User Load(9.9ms) SELECT * from users where name="bar"
(and other SQL log while retrieving the rest of the pretty_print content)
=>
[#<User id: 1>,
 #<User id: 2>,
 #<User id: 3>
:
```

There is a duplicated output in the terminal buffer if pretty_print takes more than 0.1sec. This is the downside of this pull request.

It is possible to erase it by switching to alternate screen with escape sequence `"\e[?1049h", "\e[?1049l"`.
This pull request does not do that because it will also erase the SQL log from the terminal buffer.

### Another simple implementation

This works pretty well in most cases, but I didn't use it in this pull request.
```ruby
IRB::Pager.page(retain_content: true) do |out|
  out.puts '=>'
  @context.inspect_last_value(out)
end
```
When evaluating a code like `[100.times.to_a, User.where(condition)]` or `ClassThatStoresRelationAsInstanceVariable.new(User.where(condition))`, SQL log messes up the pager output.

### DIFF
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -56,10 +56,11 @@ module IRB
       def page_with_preview(width, height, formatter_proc)
         overflow_callback = ->(lines) do
           modified_output = formatter_proc.call(lines.join, true)
-          content, = take_first_page(width, height - 1) {|o| o.write modified_output }
+          content, = take_first_page(width, [height - 2, 0].max) {|o| o.write modified_output }
           content = content.chomp
           content = "#{content}\e[0m" if Color.colorable?
           $stdout.puts content
+          $stdout.puts 'Inspecting...'
         end
         out = PageOverflowIO.new(width, height, overflow_callback)
         yield out
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -62,7 +62,7 @@ module IRB
           $stdout.puts content
           $stdout.puts 'Preparing full inspection value...'
         end
-        out = PageOverflowIO.new(width, height, overflow_callback)
+        out = PageOverflowIO.new(width, height, overflow_callback, delay: 0.1)
         yield out
         content = formatter_proc.call(out.string, out.multipage?)
         if out.multipage?
@@ -131,15 +131,16 @@ module IRB
 
     # Writable IO that has page overflow callback
     class PageOverflowIO
-      attr_reader :string
+      attr_reader :string, :first_page_lines
 
       # Maximum size of a single cell in terminal
       # Assumed worst case: "\e[1;3;4;9;38;2;255;128;128;48;2;128;128;255mA\e[0m"
       # bold, italic, underline, crossed_out, RGB forgound, RGB background
       MAX_CHAR_PER_CELL = 50
 
-      def initialize(width, height, overflow_callback)
+      def initialize(width, height, overflow_callback, delay: nil)
         @lines = []
+        @first_page_lines = nil
         @width = width
         @height = height
         @buffer = +''
@@ -147,6 +148,7 @@ module IRB
         @col = 0
         @string = +''
         @multipage = false
+        @delay_until = (Time.now + delay if delay)
       end
 
       def puts(text = '')
@@ -156,7 +158,13 @@ module IRB
 
       def write(text)
         @string << text
-        return if @multipage
+        if @multipage
+          if @delay_until && Time.now > @delay_until
+            @overflow_callback.call(@first_page_lines)
+            @delay_until = nil
+          end
+          return
+        end
 
         overflow_size = (@width * (@height - @lines.size) + @width - @col) * MAX_CHAR_PER_CELL
         if text.size >= overflow_size
@@ -182,7 +190,11 @@ module IRB
           @col = Reline::Unicode.calculate_width(@buffer)
         end
         if overflow || @lines.size > @height || (@lines.size == @height && @col > 0)
-          @overflow_callback.call(@lines.take(@height))
+          @first_page_lines = @lines.take(@height)
+          if !@delay_until || Time.now > @delay_until
+            @overflow_callback.call(@first_page_lines)
+            @delay_until = nil
+          end
           @multipage = true
         end
       end
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -60,7 +60,7 @@ module IRB
           content = content.chomp
           content = "#{content}\e[0m" if Color.colorable?
           $stdout.puts content
-          $stdout.puts 'Inspecting...'
+          $stdout.puts 'Preparing full inspection value...'
         end
         out = PageOverflowIO.new(width, height, overflow_callback)
         yield out
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'reline'
+
 module IRB
   # The implementation of this class is borrowed from RDoc's lib/rdoc/ri/driver.rb.
   # Please do NOT use this class directly outside of IRB.
@@ -47,11 +49,40 @@ module IRB
       rescue Errno::EPIPE
       end
 
-      private
-
       def should_page?
         IRB.conf[:USE_PAGER] && STDIN.tty? && (ENV.key?("TERM") && ENV["TERM"] != "dumb")
       end
+
+      def page_with_preview(width, height, formatter_proc)
+        overflow_callback = ->(lines) do
+          modified_output = formatter_proc.call(lines.join, true)
+          content, = take_first_page(width, height - 1) {|o| o.write modified_output }
+          content = content.chomp
+          content = "#{content}\e[0m" if Color.colorable?
+          $stdout.puts content
+        end
+        out = PageOverflowIO.new(width, height, overflow_callback)
+        yield out
+        content = formatter_proc.call(out.string, out.multipage?)
+        if out.multipage?
+          page(retain_content: true) do |io|
+            io.puts content
+          end
+        else
+          $stdout.puts content
+        end
+      end
+
+      def take_first_page(width, height)
+        overflow_callback = proc do |lines|
+          return lines.join, true
+        end
+        out = Pager::PageOverflowIO.new(width, height, overflow_callback)
+        yield out
+        [out.string, false]
+      end
+
+      private
 
       def content_exceeds_screen_height?(content)
         screen_height, screen_width = begin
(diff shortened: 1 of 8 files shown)


lib/irb/pager.rb
```diff
@@ -47,12 +49,39 @@ def page(retain_content: false)
...
       end
 
+      def page_with_preview(width, height, modifier_proc)
+        out = PageOverflowIO.new(width, height) do |lines|
+          modified_output = modifier_proc.call(lines.join, true)
+          content, = take_first_page(width, height - 1) {|o| o.write modified_output }
+          content = content.chomp
+          content = "#{content}\e[0m" if Color.colorable?
+          $stdout.puts content
+        end
```
st0012, line 63: nit: can we define the `overflow_callback` as a local? Like:

```rb
overflow_callback = proc do |lines|
  # ...
end
out = PageOverflowIO.new(width, height, overflow_callback)
```

Otherwise, I'd assume this block is called during the `PageOverflowIO`'s inititalization.

lib/irb/pager.rb
```diff
@@ -47,12 +49,39 @@ def page(retain_content: false)
       rescue Errno::EPIPE
       end
 
-      private
-
       def should_page?
         IRB.conf[:USE_PAGER] && STDIN.tty? && (ENV.key?("TERM") && ENV["TERM"] != "dumb")
       end
 
+      def page_with_preview(width, height, modifier_proc)
```
st0012, line 56: The block passing around `modifier_proc`, `yield`, and `overflow_callback` is quite confusing imo.
I think if we keep this method's logic inside `Irb` class, the implementation could be simpler as we don't need to package the context as block as much. WDYT?

tompng: Even if we move this part to Irb, the code is still complicated. We still need modifier_proc because it is used in two place.

```diff
+ modifier_proc = ->(content, multipage) do
+   content = content.chomp
+   content = "\n#{content}" if @context.newline_before_multiline_output? && (multipage || content.include?("\n"))
+   format(@context.return_format, content)
+ end
  ...
- yield out
+ @context.inspect_last_value(out)
  ...
```

`Irb#output_value` will be long, we will want to extract it into a method. This implementation is doing it.
I don't have a good idea to make it more simple.

The usage of this method is similar to `Pager.page do |io| end` and IMO it is not so bad.
```ruby
Pager.page_with_preview(winwidth, winheight, modifier_proc) do |out|
  @context.inspect_last_value(out)
end
```

lib/irb/pager.rb
```diff
@@ -47,12 +49,41 @@ def page(retain_content: false)
...
+        if out.multipage?
+          page(retain_content: true) do |io|
+            io.puts content
+          end
+        else
+          $stdout.puts content
+        end
+      end
+
+      def take_first_page(width, height)
```
st0012, line 76: I hope I didn't miss anything, but can we simply take `content` as an argument instead of yielding the overlow io?

```suggestion
      def take_first_page(content, width, height)
```

I think it'll reduce the code complexity a little bit.

lib/irb/pager.rb
```diff
@@ -47,12 +49,41 @@ def page(retain_content: false)
...
+        overflow_callback = ->(lines) do
+          modified_output = modifier_proc.call(lines.join, true)
+          content, = take_first_page(width, height - 1) {|o| o.write modified_output }
+          content = content.chomp
+          content = "#{content}\e[0m" if Color.colorable?
+          $stdout.puts content
+        end
+        out = PageOverflowIO.new(width, height, overflow_callback)
+        yield out
+        content = modifier_proc.call(out.string, out.multipage?)
```
st0012, line 66: I'm a bit lost on this part. Why do we need to call modifier proc in the overflow callback, and why here too?

lib/irb/pager.rb
```diff
@@ -47,12 +49,41 @@ def page(retain_content: false)
...
+        end
+      end
+
+      def take_first_page(width, height)
+        overflow_callback = proc do |lines|
+          return lines.join, true
+        end
+        out = Pager::PageOverflowIO.new(width, height, overflow_callback)
+        yield out
+        [out.string, false]
```
st0012, line 82: Is this second return value the inverted `out.multipage?`

If that's the case, can we simplify the method to just:


```rb
        out = Pager::PageOverflowIO.new(width, height, proc {})
        out.write(content)
        [out.string, !out.multipage?]
```

tompng: The core part of this method is `proc { return value }` that immediately exits from this method.
It will abort out.write even in this case.
```ruby
take_first_page(w, h) do |out|
  # Infinite loop
  # but take_first_page will return the content when overflowed.
  loop do
    write 'line'
  end
end
```
Example that pretty_print takes almost infinite time.
```
irb(main):001> a = [100000.times.to_a]*100000
=> 
[[0,
...
irb(main):002> 
```

When `yield out` is done, (that means `overflow_proc=proc{return value}` was not called) `out.multipage?` is always false.
```ruby
# Returns [first_page_content, overflowed_and_stripped]
if output.multipage?
  # Does not reach here
  return [take_first_page_of(out.string), true]
else
  return [out.string, false]
end
```

lib/irb/pager.rb
tompng: modifier_proc formats the content. (Maybe we should rename it to formatter_proc)

The format is typically `"(prefix)%s(postfix)"`.
We can do this instead of specifying modifier_proc if prefix and postfix is a constant, and the format is using `%s`.
```ruby
Pager.page_with_preview(winwidth, winheight) do |out|
  out.write(prefix)
  @context.inspect_last_value(out)
  out.write(postfix)
end
```

But there are two difficulties:
- Prefix will change depending on `content.lines.size` when newline_before_multiline_output is set to true
- Format can contain `%p` or `%.10s` instead of `%s`

So this method will

1. Calculates overflow without applying format
2. Prints the formatted preview just after IRB knows that the content overflows screen
3. Continue preparing the whole content
4. Prints the full formatted content with pager

Step 2 calls modifier_proc with partial content.
Step 4 calls modifier_proc with full content.

st0012:
I'm used to using `j` to move down the pager whenever it opens. Currently, I'd wait for the pager to open before scrolling down. It's not super convenient but easy to understand.

With this feature, I can get a preview fast, which is great 👍 

But to users:
1. It's unclear that they're seeing a preview, not the pager with the full result
2. It's unclear that when IRB switches to full content pager

So if I keep my old habit to press `j` when a pager is rendered, it would feel like the pager renders very fast, then stuck in the middle, and then resume, like demonstrated below:

https://github.com/user-attachments/assets/21cb14b4-c425-4b62-a047-b38ae0cb55b7

Is there a way we can make this a smoother experience or easier to understand through UI hints....etc.?

tompng:
> Is there a way we can make this a smoother experience or easier to understand through UI hints....etc.?

That's a good point. I updated to show a message like this.

Before pager
```
=>
[0,
 1,
 2,
 3,
 4,
Inspecting...
█
```

After pager launch
```
=>
[0,
 1,
 2,
 3,
 4,
 5,
:█
```
Considering pretty_print prints something to stdout(ex: SQL log in rails console), cursor should be at the last line and the line should be blank.

lib/irb/pager.rb
st0012: I see, thanks for the explanation.

lib/irb/pager.rb
st0012: Does this suggestion make sense? @tompng

lib/irb/pager.rb
```diff
@@ -47,12 +49,42 @@ def page(retain_content: false)
...
       end
 
+      def page_with_preview(width, height, formatter_proc)
+        overflow_callback = ->(lines) do
+          modified_output = formatter_proc.call(lines.join, true)
+          content, = take_first_page(width, [height - 2, 0].max) {|o| o.write modified_output }
+          content = content.chomp
+          content = "#{content}\e[0m" if Color.colorable?
+          $stdout.puts content
+          $stdout.puts 'Inspecting...'
```
st0012, line 63:
```suggestion
          $stdout.puts 'Preparing full inspection value...'
```

lib/irb/pager.rb
tompng: This part needs the block. Other two part that uses `take_first_page` does not need the block though.
```ruby
if omit
  content, overflow = Pager.take_first_page(winwidth, 1) do |out|
    @context.inspect_last_value(out)
  end
  ...
end
```
To show this quickly.
```
irb(main):001> a = 1000000.times.to_a
=> 
[0,
...
irb(main):002>
```
We need to take the first page in a case that full content is not available yet.

tompng:
The last commit adds 0.1s delay before calling overflow_callback.
When pretty_print completes within 0.1s, preview is not shown. This will eliminate the downside of duplicated output (preview and pager output) in terminal buffer when pretty_print is not slow.

lib/irb/pager.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'reline'
```
st0012, line 3: Q: do you think pager should be part of Reline?
If we position Reline as Ruby's default terminal application utility library, it kinda makes sense for it to have a pager class too 🤔

lib/irb/pager.rb
```diff
@@ -96,5 +128,80 @@ def setup_pager(retain_content:)
...
+        @string << text
+        if @multipage
+          if @delay_until && Time.now > @delay_until
+            @overflow_callback.call(@first_page_lines)
+            @delay_until = nil
+          end
+          return
+        end
+
+        overflow_size = (@width * (@height - @lines.size) + @width - @col )* MAX_CHAR_PER_CELL
```
st0012, line 169:
```suggestion
        overflow_size = (@width * (@height - @lines.size) + @width - @col)* MAX_CHAR_PER_CELL
```

st0012 (approved):
👍

lib/irb/pager.rb
tompng: I don't think the basic pager should be in Reline, because it's mainly about command process names(less, more) and options(-R, -X).
The added preview part is IRB specific that we can't assume pretty_print not printing a log.
If we can assume no log while pritty-printing, there is an easy way: just stream the output to pager io.

Making a pure-ruby pager library (key handling, rendering, scrolling, etc): I think it is very interesting.

PR #1041, opened 2024-11-30
Prevent cursor flickering in dancing ruby

ima1zumi:
Before

The cursor appears in the bottom right corner and occasionally flickers.

https://github.com/user-attachments/assets/ba82dbaa-f477-4950-9403-6f2e378a965b

After

https://github.com/user-attachments/assets/9f5ccf72-ff98-4995-8952-8613ea633836

### DIFF
--- a/lib/irb/easter-egg.rb
+++ b/lib/irb/easter-egg.rb
@@ -125,6 +125,7 @@ module IRB
             canvas = Canvas.new(Reline.get_screen_size)
           end
           ruby_model = RubyModel.new
+          print "\e[?25l" # hide cursor
           0.step do |i| # TODO (0..).each needs Ruby 2.6 or later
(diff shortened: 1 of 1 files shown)


lib/irb/easter-egg.rb
```diff
@@ -142,6 +143,7 @@ def render_frame(i)
         end
       end
     ensure
+      print "\e[?25h" # show cursor
```
tompng, line 146: `print "\e[?25l"` is written inside `STDOUT.cooked do end` how about move this to the ensure statement of `STDOUT.cooked do ensure end`?

ima1zumi: Good point, I've fixed it.

tompng (approved):
💎

PR #1043, opened 2024-12-10
Page the output in irb:rdbg sessions too

st0012:
IRB started to page its evaluation output and it became a useful feature for users. However, in `irb:rdbg` sessions, the output is not paged so the sudden change in behavior is surprising and inconvenient.

This commit makes `irb:rdbg` sessions page the output of the debugger too.

Notes:

- I didn't add a test for it as we currently don't have a way to test pager output.
- When printing results from `ThreadClient`, `Debug::UI#puts` will [always receive string](https://github.com/ruby/debug/blob/master/lib/debug/session.rb#L278). So we only need to activate pager in that case. In fact, `debug` most of the time passes string to this method, except for when [printing exception backtrace](https://github.com/ruby/debug/blob/master/lib/debug/session.rb#L1942).

### DIFF
--- a/lib/irb/debug/ui.rb
+++ b/lib/irb/debug/ui.rb
@@ -45,9 +45,7 @@ module IRB
             $stdout.puts line.chomp
           }
         when String
-          str.each_line{|line|
-            $stdout.puts line.chomp
-          }
+          Pager.page_content(str, retain_content: true)
         when nil
           $stdout.puts
         end


tompng:
This input will launch the pager 3 times.
```ruby
irb:rdbg(main):025> outline Object.new.tap{|o|['a','b','c'].each{o.define_singleton_method(_1*10000){}}}
```

This part calls `Debug::UI#puts` several times to print a single output.
https://github.com/ruby/debug/blob/9c4279598e9e0376657b3813bf76a206f1f97beb/lib/debug/session.rb#L278C7-L278C13
```ruby
# in DEBUGGER__::Session#process_event
output.each{|str| @ui.puts str} if ev != :suspend
```

I think this is just an acceptable edge case. Maybe we can fix it in debug.
What do you think?

st0012:
Thanks for catching that. Since `outline` is just a ported version of `ls` (https://github.com/ruby/debug/pull/173), an alternative could be that we somehow alias `outline` to IRB's `ls` command in `irb:rdbg` mode 🤔

tompng (approved):

st0012:
> an alternative could be that we somehow alias outline to IRB's ls command in irb:rdbg mode

Actually not cause we need to have an `outline` command class for that to work. And I don't think it's worth introducing a new command (even just as an alias) just to work around this edge case. I'll submit a PR to `debug` if people does hit this case.

technicalpickles:
Thanks for fixing this! I asked about this on the ruby-dx slack and was not effecting a solution this fast for it 😁

PR #1044, opened 2024-12-11
Add copy command

### LINKED ISSUE #753: Explore the possibility to add a command to copy evaluation result
It could be very helpful to have a command to copy IRB's output without manually selecting a (potentially huge) chunk of text. For example:

```
copy User.all.to_a # copies and echos all users
copy User.all.to_a; # copies and doesn't echo all users
copy _ # copies the previous output value
```

Things to consider:

- We cannot rely on any non-default Ruby gems
- We may use system commands if they're widely available, at least among UNIX-based systems

Prajjwal:
Closes #753

### DIFF
--- a/lib/irb/color_printer.rb
+++ b/lib/irb/color_printer.rb
@@ -5,8 +5,8 @@ require_relative 'color'
 module IRB
   class ColorPrinter < ::PP
     class << self
-      def pp(obj, out = $>, width = screen_width)
-        q = ColorPrinter.new(out, width)
+      def pp(obj, out = $>, width = screen_width, colorize: true)
+        q = ColorPrinter.new(out, width, colorize: colorize)
         q.guard_inspect_key {q.pp obj}
         q.flush
         out << "\n"
@@ -19,6 +19,12 @@ module IRB
       rescue Errno::EINVAL # in `winsize': Invalid argument - <STDIN>
         79
       end
+    end
+
+    def initialize(out, width, colorize: true)
+      @colorize = colorize
+
+      super(out, width)
     end
 
     def pp(obj)
@@ -41,9 +47,9 @@ module IRB
       when ',', '=>', '[', ']', '{', '}', '..', '...', /\A@\w+\z/
         super(str, width)
       when /\A#</, '=', '>'
-        super(Color.colorize(str, [:GREEN]), width)
+        super(@colorize ? Color.colorize(str, [:GREEN]) : str, width)
       else
-        super(Color.colorize_code(str, ignore_error: true), width)
+        super(@colorize ? Color.colorize_code(str, ignore_error: true) : str, width)
       end
     end
   end
--- a/lib/irb/command/copy.rb
+++ b/lib/irb/command/copy.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+module IRB
+  module Command
+    class Copy < Base
+      category "Workspace"
+      description "Copy command output to clipboard"
+
+      help_message(<<~HELP)
+        Usage: copy [command]
+      HELP
+
+      def execute(arg)
+        # Copy last value if no expression was supplied
+        arg = '_' if arg.to_s.strip.empty?
+
+        value = irb_context.workspace.binding.eval(arg)
+        output = irb_context.inspect_method.inspect_value(value, colorize: false)
+
+        if clipboard_available?
+          copy_to_clipboard(output)
+        else
+          warn "System clipboard not found"
+        end
+      rescue StandardError => e
+        warn "Error: #{e}"
+      end
+
+      private
+
+      def copy_to_clipboard(text)
+        IO.popen(clipboard_program, 'w') do |io|
+          io.write(text)
+        end
+
+        raise IOError.new("Copying to clipboard failed") unless $? == 0
+
+        puts "Copied to system clipboard"
+      rescue Errno::ENOENT => e
+        warn e.message
+        warn "Is IRB.conf[:COPY_COMMAND] set to a bad value?"
+      end
+
+      def clipboard_program
+        @clipboard_program ||= if IRB.conf[:COPY_COMMAND]
+                                 IRB.conf[:COPY_COMMAND]
+                               elsif executable?("pbcopy")
+                                 "pbcopy"
+                               elsif executable?("xclip")
+                                 "xclip -selection clipboard"
+                               end
+      end
+
+      def executable?(command)
+        system("which #{command} > /dev/null 2>&1")
+      end
+
+      def clipboard_available?
+        !!clipboard_program
+      end
+    end
+  end
+end
--- a/lib/irb/default_commands.rb
+++ b/lib/irb/default_commands.rb
@@ -9,6 +9,7 @@ require_relative "command/cd"
 require_relative "command/chws"
 require_relative "command/context"
 require_relative "command/continue"
+require_relative "command/copy"
 require_relative "command/debug"
 require_relative "command/delete"
 require_relative "command/disable_irb"
@@ -250,6 +251,7 @@ module IRB
     )
 
     register(:cd, Command::CD)
+    register(:copy, Command::Copy)
   end
 
   ExtendCommand = Command
--- a/man/irb.1
+++ b/man/irb.1
@@ -227,6 +227,8 @@ or
 .Sy type
 .
 .Pp
+.It Ev IRB_COPY_COMMAND
+Overrides the default program used to interface with the system clipboard.
 .El
 .Pp
 Also
--- a/test/irb/test_init.rb
+++ b/test/irb/test_init.rb
@@ -163,6 +163,26 @@ module TestIRB
       IRB.conf[:USE_AUTOCOMPLETE] = orig_use_autocomplete_conf
     end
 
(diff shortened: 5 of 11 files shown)


lib/irb/command/copy.rb
```diff
@@ -0,0 +1,68 @@
...
+    class Copy < Base
+      category "TODO"
+      description "TODO"
+
+      help_message(<<~HELP)
+        Usage: copy (input)
+      HELP
+
+      def execute(arg)
+        output = irb_context.workspace.binding.eval(arg)
```
st0012, line 16: I think we need to apply the [output inspection method](https://github.com/ruby/irb/blob/master/lib/irb/context.rb#L676-L678) here so the copied output will be closer to what the users may expect. 

For example, the `output` here by default will be the `to_s` result of the evaluation result, but IRB's default output inspector [pretty prints](https://github.com/ruby/irb/blob/master/lib/irb/inspector.rb#L116-L118) the result inspect.

Additionally, we need to make sure colorizing is NOT applied to the copied value.

Prajjwal: The problem with using the pretty printer is that colorizing output is [hardcoded into it](https://github.com/Shopify/irb/blob/sin.add-copy-command/lib/irb/color_printer.rb#L37-L53), and I don't think [we want it in our clipboard](https://screenshot.click/16-29-kmq50-u4g6w.jpg).

Unless there's a pretty printer that I've missed, the options are:

1. Grab colorized output, strip color before copying. Least invasive, and will respect the currently configured inspector.
- Install a new pretty printer analogous to [the default color printer](https://github.com/Shopify/irb/blob/sin.add-copy-command/lib/irb/color_printer.rb) that does basically the same thing but doesn't colorize output. Don't like this one because we're conflating two concepts and hardcoding a command to use a specific kind of inspection method.
- Like above, `color_printer.rb` to accept a parameter that turns colorizing on / off. Most invasive.

I like 1, but will defer.

lib/irb/default_commands.rb
```diff
@@ -200,6 +201,9 @@ def load_command(command)
     _register_with_aliases(:irb_continue, Command::Continue,
       [:continue, NO_OVERRIDE]
     )
+    _register_with_aliases(:irb_copy, Command::Copy,
```
st0012, line 204: We don't need to register new commands with alias. Just this should be good:

```rb
  # line 252
  register(:cd, Command::CD)
  register(:copy, Command::Copy)
```

lib/irb/color_printer.rb
```diff
@@ -46,10 +52,14 @@ def text(str, width = nil)
...
+        super(colorize ? Color.colorize(str, [:GREEN]) : str, width)
       else
-        super(Color.colorize_code(str, ignore_error: true), width)
+        super(colorize ? Color.colorize_code(str, ignore_error: true) : str, width)
       end
     end
+
+    private
+
+    attr_reader :colorize
```
st0012, line 63: Let's just use `@colorize` without adding a new attr_reader.

lib/irb/command/copy.rb
```diff
@@ -0,0 +1,59 @@
...
+      def execute(arg)
+        value = irb_context.workspace.binding.eval(arg)
+        output = irb_context.inspect_method.inspect_value(value, colorize: false)
+
+        if clipboard_available?
+          copy_to_clipboard(output)
+          puts "Copied to system clipboard"
+        else
+          temp_file = write_to_tempfile(output)
+          puts "Wrote: #{temp_file}"
```
st0012, line 22: I kinda prefer just printing something like `System clipboard not found` for now instead of writing to a tempfile. I am not against the idea but I want to wait until users without clipboard to propose solutions that could work for them the best. WDYT?

lib/irb/command/copy.rb
```diff
@@ -0,0 +1,59 @@
...
+      def write_to_tempfile(text)
+        file = Tempfile.new
+        file.write(text)
+        file.close
+        file.path
+      end
+
+      def clipboard_program
+        return "pbcopy" if executable?("pbcopy")
+        return "xclip -selection clipboard" if executable?("xclip")
```
st0012, line 47: A potential future improvement: having a `IRB.conf` or env var to allow users specify the clipboard command to use.

lib/irb/command/copy.rb
Prajjwal: Aligned, it seems like a sensible default to me but might not apply to everyone.

lib/irb/command/copy.rb
```diff
@@ -0,0 +1,53 @@
...
+          io.write(text)
+        end
+
+        raise IOError.new("Copying to clipboard failed") unless $? == 0
+
+        puts "Copied to system clipboard"
+      end
+
+      def clipboard_program
+        @clipboard_program = "pbcopy" if executable?("pbcopy")
```
st0012, line 39: Should we also have `return if defined?(@clipboard_program)` so we don't need to reevaluate the presence of the executables?

test/irb/test_init.rb
```diff
@@ -163,6 +163,26 @@ def test_use_autocomplete_environment_variable
...
+      orig_copy_command_env = ENV['IRB_COPY_COMMAND']
+      orig_copy_command_conf = IRB.conf[:COPY_COMMAND]
+
+      ENV['IRB_COPY_COMMAND'] = nil
+      IRB.setup(__FILE__)
+      refute IRB.conf[:COPY_COMMAND]
+
+      ENV['IRB_COPY_COMMAND'] = ''
+      IRB.setup(__FILE__)
+      assert_equal('', IRB.conf[:COPY_COMMAND])
```
Prajjwal, line 176: Unsure what the behavior here should be. Setting this env var overrides the other clipboard programs we use, so right now this would cause `copy` to stop working.

However, we could just as easily interpret `''` to mean "fall back to `pbcopy`".

man/irb.1
```diff
@@ -227,6 +227,8 @@ or
 .Sy type
 .
 .Pp
+.It Ev IRB_COPY_COMMAND
+Overrides the default program used to interface with the system clipboard.
```
Prajjwal, line 231: screenshot

README.md
```diff
@@ -358,6 +358,7 @@ irb(main):002> a.first. # Completes Integer methods
 - `NO_COLOR`: Assigning a value to it disables IRB's colorization.
 - `IRB_USE_AUTOCOMPLETE`: Setting it to `false` disables IRB's autocompletion.
 - `IRB_COMPLETOR`: Configures IRB's auto-completion behavior, allowing settings for either `regexp` or `type`.
+- `IRB_COPY_COMMAND`: Overrides the default program used to interface with the system clipboard.
```
Prajjwal, line 361: Does the documentation need to be updated anywhere other than here and the man page?

lib/irb/command/copy.rb
```diff
@@ -0,0 +1,58 @@
...
+    class Copy < Base
+      category "Workspace"
+      description "Copy command output to clipboard"
+
+      help_message(<<~HELP)
+        Usage: copy (command)
+      HELP
+
+      def execute(arg)
+        value = irb_context.workspace.binding.eval(arg)
```
tompng, line 14: When this command is executed without argument, `arg` will be an empty string, and the value `nil` will be copied.
I think we can show a usage message for this command, or just copy the previous result just like executing `copy _`.

lib/irb/command/copy.rb
```diff
@@ -0,0 +1,58 @@
...
+        output = irb_context.inspect_method.inspect_value(value, colorize: false)
+
+        if clipboard_available?
+          copy_to_clipboard(output)
+        else
+          warn "System clipboard not found"
+        end
+      rescue StandardError => e
+        warn "Error: #{e}"
+        warn "Is IRB.conf[:COPY_COMMAND] set to a bad value?"
```
tompng, line 24: Currently, `copy typo` will show this COPY_COMMAND warning message.
I think this message should be placed inside `def copy_to_clipboard`

lib/irb/command/copy.rb
Prajjwal: Good catch, I updated it to do the latter.

lib/irb/command/copy.rb
Prajjwal: I updated it to be `@clipboard_program ||=`, but it looks like the command is re-instantiated each time so it doesn't matter. I could cache the value elsewhere (say, the class), but that has a subtle bug where changes to `IRB.conf[:CLIPBOARD_PROGRAM]` won't reflect once this has been computed.

lib/irb/command/copy.rb
Prajjwal: Updated! Here's how it works now:

```
λ  irb → λ git sin.add-copy-command → ./bin/console
irb(main):001> copy typo
Error: undefined local variable or method `typo' for main
irb(main):002> copy 5
Copied to system clipboard
irb(main):003> IRB.conf[:COPY_COMMAND] = 'lulz'
=> "lulz"
irb(main):004> copy 5
No such file or directory - lulz
Is IRB.conf[:COPY_COMMAND] set to a bad value?
irb(main):005>
```

st0012 (approved):
This is great, thank you ❤️

Issue #975, opened 2024-06-18
HISTORY is unavailable when using rails console with RUBY_DEBUG_IRB_CONSOLE=1


tompng:
## Description

Rails console HISTORY is unavailable with `RUBY_DEBUG_IRB_CONSOLE=1`

### Preparation
Create new rails app with `rails new appname`.
Add `gem 'spring'` to Gemfile and run `spring binstub --all`. (This step is optional)

versions are:
```
debug (1.9.2)
reline (0.5.9)
irb (1.13.2)
rails (7.1.3.4)
spring (4.2.1)
```

### Reproduce
Without spring
```
$ rails c
irb(main):001> Up arrow shows history
$ RUBY_DEBUG_IRB_CONSOLE=1 rails c
irb(main):001> Up arrow does not show history
```

With spring
```
$ rails c
irb(main):001> HISTORY available
$ RUBY_DEBUG_IRB_CONSOLE=1 rails c
irb(main):001> HISTORY available (until `spring stop`)
$ spring stop
$ RUBY_DEBUG_IRB_CONSOLE=1 rails c
irb(main):001> HISTORY unavailable
$ rails c
irb(main):001> HISTORY unavailable (until `spring stop`)
$ spring stop
$ rails c
irb(main):001> HISTORY available
```

## Result of irb_info

```
irb(main):001> irb_info
Ruby version: 3.3.1
IRB version: irb 1.13.2 (2024-06-15)
InputMethod: RelineInputMethod with Reline 0.5.9 and /Users/tomoya.ishida/.inputrc
Completion: Autocomplete, RegexpCompletor
RUBY_PLATFORM: arm64-darwin22
LANG env: ja_JP.UTF-8
East Asian Ambiguous Width: 1
```

st0012:
I think the issue is just between `debug` and IRB. It shows up in Rails console likely because it requires and activates `debug` by default and thus turning Rails console's IRB into a `irb:rdbg` session.

PR #1046, opened 2024-12-12
Load history when starting a direct debug session

sunblaze:
When starting a debug session directly with RUBY_DEBUG_IRB_CONSOLE=1 and `require 'debug'; debugger`, IRB's history wasn't loaded. This commit ensures history is loaded in this case by calling `load_history` when configuring IRB for the debugger.

Fixes ruby/irb#975

### DIFF
--- a/lib/irb/debug.rb
+++ b/lib/irb/debug.rb
@@ -81,6 +81,7 @@ module IRB
         IRB.instance_variable_set(:@debugger_irb, irb)
         irb.context.with_debugger = true
         irb.context.irb_name += ":rdbg"
+        irb.context.io.load_history
       end
 
       module SkipPathHelperForIRB
--- a/lib/irb/debug.rb
+++ b/lib/irb/debug.rb
@@ -81,7 +81,7 @@ module IRB
         IRB.instance_variable_set(:@debugger_irb, irb)
         irb.context.with_debugger = true
         irb.context.irb_name += ":rdbg"
-        irb.context.io.load_history
+        irb.context.io.load_history if irb.context.io.class < HistorySavingAbility
       end
 
       module SkipPathHelperForIRB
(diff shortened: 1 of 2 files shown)


test/irb/test_history.rb
```diff
@@ -488,6 +488,28 @@ def foo
...
+        debugger
+        binding.irb # needed to satisfy run_ruby_file
+      RUBY
+
+      output = run_ruby_file do
+        type "history"
+        type "exit!"
+      end
+
+      assert_include(output, "old_history_3")
```
st0012, line 510: The current test already captures what we want to cover, but I think we can make expectations even clearer to help future readers understand the context:

```suggestion
      output = run_ruby_file do
        type "history"
        type "puts 'foo'"
        type "history"
        type "exit!"
      end

      assert_include(output, "irb:rdbg(main):002") # assert that we're in an irb:rdbg session
      assert_include(output, "5: history")
      assert_include(output, "4: puts 'foo'")
      assert_include(output, "3: history")
      assert_include(output, "2: old_history_3")
      assert_include(output, "1: old_history_2")
      assert_include(output, "0: old_history_1")
```

st0012 (approved):
Can confirm the patch fixes the issue, thank you!

lib/irb/debug.rb
```diff
@@ -81,6 +81,7 @@ def configure_irb_for_debugger(irb)
         IRB.instance_variable_set(:@debugger_irb, irb)
         irb.context.with_debugger = true
         irb.context.irb_name += ":rdbg"
+        irb.context.io.load_history
```
st0012, line 84: It looks like we need some safeguarding here as `context.io` doesn't always include `HistorySavingAbility`, especially in some non-tty environments.

```suggestion
        irb.context.io.load_history if irb.context.io.class < HistorySavingAbility
```

test/irb/test_history.rb
sunblaze: Nice additions!

PR #1047, opened 2024-12-12
Avoid generating documentation pages for internal components

st0012:
Currently, the documentation website lists most of IRB's internal classes/modules...etc.

Screenshot 2024-12-12 at 23 00 38

They're not helpful to end users and are mostly undocumented anyway. Therefore, we should avoid generating documentation for them.

Result:
Screenshot 2024-12-12 at 23 02 18

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -46,8 +46,6 @@ task :default => :test
 
 RDoc::Task.new do |rdoc|
   rdoc.title = "IRB"
-  rdoc.rdoc_files.include("*.md", "lib/**/*.rb")
-  rdoc.rdoc_files.exclude("lib/irb/xmp.rb")
   rdoc.rdoc_dir = "docs"
   rdoc.main = "README.md"
   rdoc.options.push("--copy-files", "LICENSE.txt")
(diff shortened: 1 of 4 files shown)


Rakefile
```diff
@@ -46,8 +46,6 @@ task :default => :test
 
 RDoc::Task.new do |rdoc|
   rdoc.title = "IRB"
-  rdoc.rdoc_files.include("*.md", "lib/**/*.rb")
-  rdoc.rdoc_files.exclude("lib/irb/xmp.rb")
```
st0012, line 50: If we use these options, `.document` will be ignored. Since `.document` will be applied to `ruby/ruby`'s documentation generation too, we should prioritize using it.

st0012: Actually, this is not correct. I think it'd be better if that's the case though. Will see how I can make it happen.

lib/irb.rb
```diff
@@ -1504,7 +1504,7 @@ def format_prompt(format, ltype, indent, line_no) # :nodoc:
   end
 end
 
-class Binding
+class Binding # :nodoc:
```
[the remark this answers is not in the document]
st0012: Yeah good point. I've removed this change.

Rakefile
st0012: @tompng I opened https://github.com/ruby/ruby/pull/12325 for this. Can you review it there too?

PR #1051, opened 2024-12-18
Drop ColorPrinter's workaround for BasicObject

st0012:
`pp` 0.6.0+ includes https://github.com/ruby/pp/pull/26 to handle BasicObject, so we can drop the workaround.

(`pp` supports Ruby 2.7+ too)

### DIFF
--- a/irb.gemspec
+++ b/irb.gemspec
@@ -42,4 +42,5 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "reline", ">= 0.4.2"
   spec.add_dependency "rdoc", ">= 4.0.0"
+  spec.add_dependency "pp", ">= 0.6.0"
 end
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,6 +42,8 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          # Added to make Ruby 2.7 correctly require installed default gems, like `pp`.
+          rubygems: latest
       - name: Run tests
         run: bundle exec rake test
       - name: Run tests in isolation
--- a/lib/irb/color_printer.rb
+++ b/lib/irb/color_printer.rb
@@ -4,9 +4,6 @@ require_relative 'color'
 
 module IRB
   class ColorPrinter < ::PP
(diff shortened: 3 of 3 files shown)


st0012:
Based on the debugging run's message: 

```
Error: test_object_inspection_handles_basic_object(TestIRB::ContextTest): NameError: uninitialized constant PP::VERSION
/home/runner/work/irb/irb/test/irb/test_context.rb:161:in `test_object_inspection_handles_basic_object'
     15[8](https://github.com/ruby/irb/actions/runs/12399361528/job/34614111414?pr=1051#step:4:9):         irb.eval_input
     159:       end
     160:       assert_empty err
  => 161:       puts PP::VERSION
     162:       assert_not_match(/NoMethodError/, out)
     163:       assert_match(/#<BasicObject:.*>/, out)
     164:     ensure
Error: NameError: uninitialized constant PP::VERSION
```

It looks like Ruby 2.7's bundler failed to use the `pp` gem and still uses the one that's bundled with Ruby.

irb.gemspec
```diff
@@ -42,4 +42,5 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "reline", ">= 0.4.2"
   spec.add_dependency "rdoc", ">= 4.0.0"
+  spec.add_dependency "pp", ">= 0.6.0"
```
tompng, line 45: Do we need to specify this instead of adding it just to Gemfile for testing?
User can install bug fixed version of pp if they want (but I know user usually don't update pp nor add pp to Gemfile)
It is beneficial, but adding it is not a usual case. We don't usually bump reline or rdoc requirements for each bug fixes.
What do you think?

st0012: We need to specify it otherwise with the workaround removed, users with newer IRB but older pp will get a regression when dealing with basic objects.

> It is beneficial, but adding it is not a usual case. We don't usually bump reline or rdoc requirements for each bug fixes.

We do when we have to make corresponding changes on IRB side?

It's like when `repl_type_completor` adopts new Prism APIs/structures, it bumps its requirement for `prism`.

We can of course make the patch conditional based on the version of `pp` it uses, but IMO with the size of the current patch it's not worth it.

tompng (approved):

Issue #630, opened 2023-07-04
Configs that haven't been covered by tests


st0012:
These are the configs (`IRB.conf[config_name]`) referenced in `Context#initialize` but not covered by any test cases:

- [ ] `:AP_NAME`
- [ ] `:LOAD_MODULES`
- [ ] `:USE_TRACER` - This could be broken with `ruby/tracer` version. May need to drop it directly
- [ ] `:USE_LOADER`
- [ ] `:IGNORE_SIGINT`
- [ ] `:IGNORE_EOF`
- [ ] `:BACK_TRACE_LIMIT`
- [ ] `:IRB_NAME`
- [ ] `:EXTRA_DOC_DIRS`

PR #1052, opened 2024-12-20
add context.ap_name test

QWYNG:
Thank you for the great gem.
This partially resolves issue #630 .
If this test seems good, I would like to tackle other items as well.


tompng (approved):
:+1:

PR #1053, opened 2024-12-22
Restructure IRB's documentation

st0012:
Currently, part of the documentation lives in `lib/irb.rb` as RDoc comments, and the rest lives in `README.md`.

The RDoc comments are not for viewing on GitHub, while README.md is. So this means users need to read part of the documentation on https://ruby.github.io/irb/, and the rest on GitHub. This is not a great user experience.

This PR restructures the documentation so all major documentation is in `doc/index.md` that will become the main documentation for IRB on https://ruby.github.io/irb/.

This solves a 2 main problems:

1. Users can read IRB documentation all in one place, on https://ruby.github.io/irb/.
2. It makes updating the documentation easier, especially the content that currently lives in `lib/irb.rb`.

## Next Steps

- Once this PR is merged and deployed, we can replace `README.md`'s content with links to relevant sections.
- `EXTEND_IRB.md` needs to be updated to link to the new page. The file can't be deleted because it was advertised in multiple places.
- https://github.com/ruby/rdoc/pull/1250 will allow https://ruby.github.io/irb/ to have Table of Contents for the index page too.
- We should also add guidance on contributing to IRB's documentation.

## What This PR Does NOT Do

This PR is a foundation for future documentation changes. It does not include fact-checking existing documentation or correcting existing errors...etc.

## Demo (generated with https://github.com/ruby/rdoc/pull/1250)

https://github.com/user-attachments/assets/4274ca54-5e01-40e9-8c0e-1000cc13a37f


.rdoc_options
```diff
@@ -0,0 +1,2 @@
+page_dir: doc
+warn_missing_rdoc_ref: true
```
st0012, line 2: These settings can't be specified in RDoc's rake task so we need this file.

Rakefile
```diff
@@ -45,8 +45,8 @@ end
...
 
 RDoc::Task.new do |rdoc|
-  rdoc.title = "IRB"
-  rdoc.rdoc_dir = "docs"
-  rdoc.main = "README.md"
-  rdoc.options.push("--copy-files", "LICENSE.txt")
+  rdoc.title = "IRB Documentation"
+  rdoc.main = "index.md"
+  rdoc.rdoc_dir = "_site"
+  rdoc.options.push("lib")
```
st0012, line 51: We also need this task because we now need to use `lib` as the target folder, not the root folder.

st0012 (commented):
RDoc setting related changes are all for achieving the followings all at once:

- All doc files should live under `doc` folder
- When RDoc generates these files' pages, they should NOT be put under `doc` folder, but treated as top-level pages in the sidebar
- `doc/index.md` should be the main page
- Documentation under `lib` should be generated as usual

The current implementation is the only set of configurations I could make this work. RDoc really needs to be a better tool 😞

tompng (approved):
Looks good 👍  🚀

PR #1057, opened 2024-12-29
Fix broken `history` command with `-g`

sanfrecce-osaka:
The `-g` option of `history` command was not working, so I fixed it. Local variable `grep` was always nil because the regular expression parsing options contained an unnecessary `\n`. `test_history_grep` did not detect this bug because it only asserted what was included in the output.


st0012 (approved):
Great catch! Thanks for fixing it.

PR #1060, opened 2025-01-06
Simplify vterm-yamatanooroti's steps

st0012:


.github/workflows/test.yml
```diff
@@ -98,8 +94,10 @@ jobs:
...
           make
           sudo make install
-      - name: Install dependencies
-        run: |
-          WITH_VTERM=1 bundle install
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
```
st0012, line 101: This automatically runs bundled install and caches the result, so we don't need a separate step for it. Putting it after libvterm installation to make sure we have all the dependencies needed for gems.

PR #1069, opened 2025-01-21
Fix pager preview with escape sequence and newlines

tompng:
Fixes this bug
```
irb(main):001> 100.times.map{[1]*it}
An error occurred when inspecting the object: #<TypeError: no implicit conversion of nil into String>
Result of Kernel#inspect: #<Array:0x0000000129d94778>
=> 
[[],
 [1],
 [1, 1],
 [1, 1, 1],
 [1, 1, 1, 1],
 [1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
 [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
irb(main):002> 
```

### Non-string
PP doesn't seem to write non-string object, but fix it in case.
```ruby
text = text.to_s unless text.is_a?(String)
```

### Escape sequence
Need to set the second parameter `allow_escape_code` to `Reline::Unicode.calculate_width`
```ruby
Reline::Unicode.calculate_width("\e[31mA\e[m") #=> 11 because length of "^[[31mA^[[m"(escaped for printing) is 11
Reline::Unicode.calculate_width("\e[31mA\e[m", true) #=> 1
```

### Newline
`Reline::Unicode.split_by_width(line.chomp, @width)`
`wrapped_lines` does not include `"\n"` because line is chomp-ed. (`split_by_width` does not work if string contains newline character)
So we need to add `"\n"` if `line` ends with `"\n"`. And the logic had a bug.
PP doesn't seem to write continuous newlines, but found while adding more tests.

### DIFF
--- a/lib/irb/pager.rb
+++ b/lib/irb/pager.rb
@@ -152,11 +152,13 @@ module IRB
       end
 
       def puts(text = '')
+        text = text.to_s unless text.is_a?(String)
         write(text)
         write("\n") unless text.end_with?("\n")
       end
 
       def write(text)
+        text = text.to_s unless text.is_a?(String)
         @string << text
         if @multipage
           if @delay_until && Time.now > @delay_until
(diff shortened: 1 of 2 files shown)


ima1zumi (approved):
LGTM!

PR #1075, opened 2025-01-25
Remove ruby-core workflow

st0012:
Since IRB is not a default gem anymore, we don't need to run CI against Ruby core anymore, at least not in the current form.


tompng (approved):
:+1:
Test failing because readline is changed to a bundled gem. Not related to this change.

PR #1076, opened 2025-01-26
Fallback to Reline when `require 'readline'` fails

tompng:
Require readline may fail because it is a bundled gem in 3.5.0.dev.

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -175,10 +175,15 @@ module IRB
   class ReadlineInputMethod < StdioInputMethod
     class << self
       def initialize_readline
-        require "readline"
-      rescue LoadError
-      else
-        include ::Readline
+        return if defined?(self::Readline)
+
+        begin
+          require 'readline'
+          const_set(:Readline, ::Readline)
+        rescue LoadError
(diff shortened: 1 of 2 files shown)


lib/irb/input-method.rb
```diff
@@ -175,10 +175,15 @@ def close
...
-      else
-        include ::Readline
+        return if defined?(self::Readline)
+
+        begin
+          require 'readline'
+          const_set(:Readline, ::Readline)
+        rescue LoadError
+          const_set(:Readline, ::Reline)
+        end
```
tompng, line 185: Adds `ReadlineInputMethod::Readline = ::Readline or ::Reline`
because we don't want to pollute the top level by adding constant `::Readline` when `require 'readline'` fails.

st0012:
Is there any feature that readline does better than reline? Does it make sense to still allow both types of input methods going forward 🤔

tompng:
I didn't thought of that. Maybe it's time to use Reline also in `"--singleline", "--readline", "--legacy"` mode.

(Off topic: `irb --readline` was useful for me to check the difference between Reline and Readline while developing Reline)

PR #1078, opened 2025-01-27
bundled gems migration

hsbt:
We also need https://github.com/ruby/irb/pull/1076 and https://github.com/ruby/irb/pull/1074


tompng (approved):
Thank you :+1:

PR #1079, opened 2025-01-28
Ignore to contain directory to Gem::Specification#files

hsbt:
Fixes https://github.com/ruby/ruby/actions/runs/13000759399/job/36258746468?pr=12648#step:8:160

`Dir.glob` contained directory like `lib/irb`. But `rbinstall.rb` don't handle directory with `Gem::Specification#files`. We should ignore that.


hsbt:
ref. https://github.com/ruby/ruby/pull/12648

PR #1082, opened 2025-02-05
(docs) Document the keys for completion

andyw8:
(it's a little unintuitive that the cursor keys can't be used for selecting a suggestion).


st0012 (approved):
Thank you!

PR #1085, opened 2025-02-13
Handle keyword local variables correctly

tompng:
Local variable can be a keyword. Fix a bug that IRB doesn't work in this case.
```ruby
irb(main):001> {if:1} => {if:}
=> nil
irb(main):002> 1+2
irb(main):003> 
irb(main):004> (Can't submit input)
```

IRB prepends local variable assignment code `a=_=nil;` to the input code. (Because Ruby code is parsed differently depending on local variable existence)
Prepending `if=_=nil;` cause bug in this case.


lib/irb/completion.rb
```diff
@@ -459,12 +439,12 @@ def retrieve_completion_data(input, bind:, doc_namespace:)
             eval("#{perfect_match_var}.class.name", bind) rescue nil
           else
             candidates = (bind.eval_methods | bind.eval_private_methods | bind.local_variables | bind.eval_instance_variables | bind.eval_class_constants).collect{|m| m.to_s}
-            candidates |= ReservedWords
+            candidates |= RubyLex::RESERVED_WORDS.map(&:to_s)
```
st0012, line 442: I think using `name` will prevent recreating string objects every time this is called:

```
irb(main):002> :foo.name.object_id
=> 11536
irb(main):003> :foo.name.object_id
=> 11536
irb(main):004> :foo.name.object_id
=> 11536
irb(main):005> :foo.to_s.object_id
=> 15608
irb(main):006> :foo.to_s.object_id
=> 16096
```

lib/irb/completion.rb
```diff
@@ -459,12 +439,12 @@ def retrieve_completion_data(input, bind:, doc_namespace:)
...
             candidates.find{ |i| i == input }
           end
         else
           candidates = (bind.eval_methods | bind.eval_private_methods | bind.local_variables | bind.eval_instance_variables | bind.eval_class_constants).collect{|m| m.to_s}
-          candidates |= ReservedWords
+          candidates |= RubyLex::RESERVED_WORDS.map(&:to_s)
```
st0012, line 447: Same as the above.

lib/irb/completion.rb
tompng: `Symbol#name` is added in Ruby 3.0 and IRB supports >= 2.7

st0012 (approved):

PR #1086, opened 2025-02-18
Document `USE_PAGER` config

artur-intech:
Enhances https://github.com/ruby/irb/pull/783.

### DIFF
--- a/doc/Configurations.md
+++ b/doc/Configurations.md
@@ -156,6 +156,8 @@ You can see the hash by typing `IRB.conf`. Below are the primary entries:
 - `:USE_TRACER`: Whether to use the
   {IRB tracer}[rdoc-ref:IRB@Tracer];
   initial value: `false`.
+- `:USE_PAGER`: Controls whether "pagination" is enabled.
+  initial value: `true`.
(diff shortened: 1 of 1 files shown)


doc/Configurations.md
```diff
@@ -156,6 +156,8 @@ You can see the hash by typing `IRB.conf`. Below are the primary entries:
 - `:USE_TRACER`: Whether to use the
   {IRB tracer}[rdoc-ref:IRB@Tracer];
   initial value: `false`.
+- `:USE_PAGER`: Controls whether "pagination" is enabled.
```
tompng, line 159: How about using the word `pager` instead of quoted `"pagination"`?
This document file already use the word `pager`
```md
- `RI_PAGER` / `PAGER`: Specifies the pager for documentation.
```

I think the word pagination is mostly used for dividing content into multiple numbered pages (https://en.wikipedia.org/wiki/Pagination)
https://en.wikipedia.org/wiki/Terminal_pager does not use the word pagination.

artur-intech: That's a great idea! I didn't know the terminal pager is a thing. Thank you for enlightening!

tompng (approved):
Thank you :+1:

Issue #1024, opened 2024-10-28
Replacing Ripper by Prism


eregon:
## Description

It would be great to replace the usages of Ripper in IRB by Prism.
There are many advantages:
* Prism is faster than Ripper, and the difference is even more visible on TruffleRuby, where `irb` is visibly less reactive because of Ripper.
* Prism provides a much better API
* Prism works well on all Ruby implementations, Ripper is a big hack which is very hard to support on all Ruby implementations (at least it has proven difficult to maintain on CRuby, JRuby and TruffleRuby, right now TruffleRuby is updating to Ruby 3.3 and Ripper is proving to be again a pain to get working, it even has a copy of `st.c` now to give an idea, on top of depending on many internal CRuby files).
* Prism is actually used in CRuby now as the default parser since Ruby 3.4 preview 2 and is the only parser in TruffleRuby, so it is the source of truth when it comes to parsing Ruby code.

Could this be considered? What are the challenges?

st0012:
I do consider this but not prioritizing it before Ruby 3.4. Its priority will depend on when IRB becomes a bundled gem too, which will make having `prism` gem as its dependency easier.

@tompng what do you think?

hsbt:
The boilerplate of Rails 8 contained `irb`. And the minimum support version of Rails 8 is Ruby 3.2.

If we replace `ripper` to `prism`, we need to add `prism` to dependency of `irb` for Ruby 3.2 and Rails 8. It means everyone compile `prism` gem for their Rails application.

I wonder if this is really beneficial to irb and Rails users.

eregon:
With the transition to [more bundled gems](https://github.com/ruby/ruby/blob/master/lib/bundled_gems.rb) it seems Rails already has many native dependencies.
For instance:
```
$ ruby -v
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-linux]
$ gem i rails
...
Building native extensions. This could take a while...
Successfully installed racc-1.8.1
...
Successfully installed websocket-extensions-0.1.5
Building native extensions. This could take a while...
Successfully installed websocket-driver-0.7.6
Building native extensions. This could take a while...
...
Successfully installed rails-7.2.2
46 gems installed
```

Also:
```
$ rails new rails8
$ cd rails8
$ rm -rf vendor/bundle
$ bundle install
...
Installing stringio 3.1.1 with native extensions
Installing io-console 0.7.2 with native extensions
Installing bindex 0.8.1 with native extensions
[...]

PR #1091, opened 2025-04-11
Syntax Highlight using Prism

tompng:
#1024
Migrate syntax highlight from Ripper to Prism

# Syntax highlighting

Colorize from these three information:

- Basically colorize from Prism token types
- Error coloring from error location
- Exceptional case from Prism syntax tree

Mix tokens and pseudo tokens created from error location and syntax tree which is prior than normal tokens.

## Exceptional case

Symbol name can by IDENTIFIER, operator-token or keyword-token. It may not be placed after SYMBOL_BEGIN
```ruby
:symbol; :<=>; :if

<<A; :\
A
symbol
```

Symbol and String colorize
Sometimes it's hard to distinguish symbol and string because both consists of same token types: `STRING_CONTENT`, `EMBEXPR_BEGIN`, `EMBEXPR_END` and `STRING_END`.
Traversing syntax tree is the easiest way to colorize it correctly.
```ruby
%w[a b c]
%i[a b c]
:"symbol#{embed}symbol"
"string#{embed}string"
```

IRB colors defined method name blue. Doing this is really easy from syntax tree but hard from token.
```ruby
def method_name_colored_blue_in_irb; end
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,6 +50,26 @@ jobs:
         run: bundle exec rake test
       - name: Run tests in isolation
         run: bundle exec rake test_in_isolation
+  prism-version-test:
+    needs: ruby-versions
+    name: rake test prism-${{ matrix.prism_version }}
+    strategy:
+      matrix:
+        prism_version: ['1.3.0', '1.7.0', 'latest']
+      fail-fast: false
+    runs-on: ubuntu-latest
+    env:
+      PRISM_VERSION: ${{matrix.prism_version}}
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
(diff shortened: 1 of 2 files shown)


lib/irb/color.rb
```diff
@@ -129,135 +161,132 @@ def colorize(text, seq, colorable: colorable?)
       def colorize_code(code, complete: true, ignore_error: false, colorable: colorable?, local_variables: [])
         return code unless colorable
 
-        symbol_state = SymbolState.new
-        colored = +''
-        lvars_code = RubyLex.generate_local_variables_assign_code(local_variables)
-        code_with_lvars = lvars_code ? "#{lvars_code}\n#{code}" : code
+        result = Prism.parse_lex(code, scopes: [local_variables])
```
Earlopain: Prism by default parses code as the latest version it currently supports. In irb you probably want to parse as the version that the user is currently running as.

You can do that by passing in `version: RUBY_VERSION` (or `version: "current"` on more recent prism for better error messages)

PR #1093, opened 2025-04-26
Avoid intermediate array from split

Maumagnaguagno:
Match first sequence of non-whitespace characters.


lib/irb/debug.rb
```diff
@@ -60,7 +60,7 @@ def DEBUGGER__.capture_frames(*args)
         if !DEBUGGER__::CONFIG[:no_hint] && irb.context.io.is_a?(RelineInputMethod)
           Reline.output_modifier_proc = proc do |input, complete:|
             unless input.strip.empty?
-              cmd = input.split(/\s/, 2).first
+              cmd = input[/\S+/]
```
tompng, line 63: The behavior is changed (example input: `'  info  '`) but the changed behavior matches better to debug command recognization rule. :+1:

tompng (approved):
:+1:

PR #1094, opened 2025-05-03
Prefer filter_map and map+grep instead of map+compact and select+map

Maumagnaguagno:
``Enumerable#filter_map`` is available since Ruby 2.7.

### DIFF
--- a/test/irb/test_irb.rb
+++ b/test/irb/test_irb.rb
@@ -932,7 +932,7 @@ module TestIRB
       end
 
       assert_match(/irbtest-.*\.rb:2:in (`|'Object#)foo': error \(RuntimeError\)/, output)
-      frame_traces = output.split("\n").filter_map { |line| line.strip if line.strip.match?(/from /) }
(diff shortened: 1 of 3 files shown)


test/irb/test_irb.rb
```diff
@@ -877,7 +877,7 @@ def bar
       end
 
       assert_match(/irbtest-.*\.rb:2:in (`|'Object#)foo': error \(RuntimeError\)/, output)
-      frame_traces = output.split("\n").select { |line| line.strip.match?(/from /) }.map(&:strip)
+      frame_traces = output.split("\n").filter_map { |line| line.strip if line.strip.match?(/from /) }
```
tompng, line 880: Other part of this pull request looks good.
This part, I', not sure if this change is obviously better than before.

How about changing it to `.map(&:strip).grep(/from /)` ?

Maumagnaguagno: Yes, that is much better!

tompng (approved):
Thank you :+1:

PR #1095, opened 2025-05-04
Replace gsub with rstrip

Maumagnaguagno:
The regular expression ``/\s*\z/`` matches any whitespace characters at end of string, which means only one match is possible with ``gsub``, replacing it with an empty string. The same can be achieved by ``String#rstrip``.


tompng:
Looks good, thank you!

Issue #1089, opened 2025-04-07
Exception when using `irb:rdbg` where object`to_s` might return `nil`


adam12:
The `to_s` method returning `nil` is obviously an issue, but we should guard against that anyways.

```ruby
class SomeObject
  def to_s
    nil
  end

  def do_something
    debugger
  end
end

SomeObject.new.do_something
```

```
$ rdbg -e 'config set irb_console 1 ;; c' -c ruby test.rb
```

```
/Users/adam/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/irb-1.15.2/lib/irb.rb:601:in 'IRB::Irb#truncate_prompt_main': undefined method 'tr' for nil (NoMethodError)

      str = str.tr(CONTROL_CHARACTERS_PATTERN, ' ')
               ^^^
        from /Users/adam/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/irb-1.15.2/lib/irb.rb:616:in 'block in IRB::Irb#format_prompt'
        from /Users/adam/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/irb-1.15.2/lib/irb.rb:610:in 'String#gsub'
        from /Users/adam/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/irb-1.15.2/lib/irb.rb:610:in 'IRB::Irb#format_prompt'
        from /Users/adam/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/irb-1.15.2/lib/irb.rb:585:in 'IRB::Irb#generate_prompt'
        from /Users/adam/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/irb-1.15.2/lib/irb.rb:242:in 'IRB::Irb#readmultiline'
        from /Users/adam/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/irb-1.15.2/lib/irb.rb:270:in 'block in IRB::Irb#each_top_level_statement'
        from <internal:kernel>:168:in 'Kernel#loop'
        from /Users/adam/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/irb-1.15.2/lib/irb.rb:269:in 'IRB::Irb#each_top_level_statement'
[...]

PR #1097, opened 2025-05-07
Fix nil error on debugger prompt

muno92:
Fix ruby/irb#1089

This change ensures that when a class's to_s method returns nil, it is handled in the same way as if to_s had raised an error.

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -613,9 +613,11 @@ module IRB
           @context.irb_name
         when "m"
           main_str = @context.safe_method_call_on_main(:to_s) rescue "!#{$!.class}"
+          main_str = main_str || "!#{@context.safe_method_call_on_main(:class)}"
           truncate_prompt_main(main_str)
         when "M"
           main_str = @context.safe_method_call_on_main(:inspect) rescue "!#{$!.class}"
+          main_str = main_str || "!#{@context.safe_method_call_on_main(:class)}"
           truncate_prompt_main(main_str)
         when "l"
           ltype
--- a/test/irb/test_debugger_integration.rb
+++ b/test/irb/test_debugger_integration.rb
@@ -51,6 +51,56 @@ module TestIRB
       assert_match(/=>   2\| puts "hello"/, output)
     end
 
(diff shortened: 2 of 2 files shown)


lib/irb.rb
```diff
@@ -613,9 +613,11 @@ def format_prompt(format, ltype, indent, line_no) # :nodoc:
           @context.irb_name
         when "m"
           main_str = @context.safe_method_call_on_main(:to_s) rescue "!#{$!.class}"
+          main_str = main_str || "!#{@context.safe_method_call_on_main(:class)}"
```
tompng, line 616: I think `!` in `"!RuntimeError"` is expressing that stringifying main raised `RuntimeError`, not that main is a `RuntimeError`.
So we don't need `!` prefixed to a non-error class name.

I think main_str can be just an empty string. Just like when `to_s` returned an empty string.
How about `main_str = "#{@context.safe_method_call_on_main(:to_s)}" rescue "!#{$!.class}"`
This way, it can handle other erroneous case such as `to_s` returned `Object.new`

test/irb/test_debugger_integration.rb
```diff
@@ -51,6 +51,56 @@ def test_debug
       assert_match(/=>   2\| puts "hello"/, output)
     end
 
+    def test_debug_class_to_s_return_string
```
tompng, line 54: This test case is not needed.
It is testing a common normal case that to_s is defined and it returns string. I think this case is already tested in another test case.

lib/irb.rb
tompng: Or this one
```ruby
main_str = "!TypeError" unless String === main_str
```

muno92: Thank you for reviewing.

I misunderstood meaning of prompt.

And `"#{@context.safe_method_call_on_main(:to_s)}"` looks better!
So I changed.

test/irb/test_debugger_integration.rb
muno92: I removed this test case.

st0012 (approved):

PR #1098, opened 2025-05-20
[DOC] Include document files in the generated gem file

nobu:


irb.gemspec
```diff
@@ -22,19 +22,12 @@ Gem::Specification.new do |spec|
...
-    "bin/console",
-    "bin/setup",
-    "doc/irb/irb-tools.rd.ja",
-    "doc/irb/irb.rd.ja",
     "exe/irb",
-    "irb.gemspec",
     "man/irb.1",
   ] + Dir.chdir(File.expand_path('..', __FILE__)) do
-    Dir.glob("lib/**/*").map {|f| f unless File.directory?(f) }.compact
+    Dir.glob(%w[*.txt *.md bin/* doc/**/* lib/**/*]).select {|f| !File.directory?(f) }
```
st0012, line 30: I think files in `bin/` are for development only (at least in this repo). So maybe we don't need to include it?

st0012 (approved):

irb.gemspec
nobu: Removed.

Issue #900, opened 2024-03-13
Completion crashes on a method having invalid name as UTF-8


pocke:
## Description

IRB crashes with the following code:

```
irb(main):001> define_method("a\xff".force_encoding(Encoding::ASCII_8BIT)){}
=> :"a\xFF"
irb(main):002> a/Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/gems/3.4.0+0/gems/irb-1.11.2/lib/irb/completion.rb:184:in `encode': "\xFF" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/gems/3.4.0+0/gems/irb-1.11.2/lib/irb/completion.rb:184:in `block in completion_candidates'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/gems/3.4.0+0/gems/irb-1.11.2/lib/irb/completion.rb:184:in `map'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/gems/3.4.0+0/gems/irb-1.11.2/lib/irb/completion.rb:184:in `completion_candidates'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/gems/3.4.0+0/gems/irb-1.11.2/lib/irb/input-method.rb:251:in `block in initialize'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/3.4.0+0/reline/line_editor.rb:1634:in `call_completion_proc_with_checking_args'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/3.4.0+0/reline/line_editor.rb:533:in `call_completion_proc_with_checking_args'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/3.4.0+0/reline.rb:237:in `block in <class:Core>'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/3.4.0+0/reline/line_editor.rb:588:in `instance_exec'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/3.4.0+0/reline/line_editor.rb:588:in `call'
	from /Users/kuwabara.masataka/.rbenv/versions/trunk/lib/ruby/3.4.0+0/reline/line_editor.rb:623:in `call'
[...]

PR #1101, opened 2025-06-03
Fix methods defined with invalid encoding are not displayed in completion

ksaito422:
## Summary

This PR fixes a crash in IRB's method completion when a completion candidate contains characters that cannot be converted to `Encoding.default_external`.（#900）

## Problem

Currently, if a method or variable name includes characters incompatible with `Encoding.default_external`, triggering method completion results in an `Encoding::UndefinedConversionError`, which crashes the entire IRB session. This leads to a poor user experience, as the session terminates unexpectedly and work can be lost.

## Solution

Instead of allowing the exception to crash the session, this change catches the encoding error and prints a warning message to stderr. To avoid flooding the console, this warning is displayed only once per session upon the first occurrence of the error.

This approach is preferable because it informs the user about the underlying encoding issue without abruptly ending their workflow.

### Behavior Change

-   **Before:** IRB crashes when method completion is triggered with an incompatible character.
-   **After:** A warning is printed once, and the IRB session continues to run.

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -193,13 +193,11 @@ module IRB
 
       # It doesn't make sense to propose commands with other preposing
       commands = [] unless preposing.empty?
-      @encoding_warning_shown ||= false
 
       completion_data = retrieve_completion_data(target, bind: bind, doc_namespace: false).compact.map do |i|
         i.encode(Encoding.default_external)
       rescue Encoding::UndefinedConversionError
-        warn "Warning: Invalid encoding in method name '#{i}'. can't be converted to the locale #{Encoding.default_external}." unless @encoding_warning_shown
-        @encoding_warning_shown = true
+        # If the string cannot be converted, we just ignore it
         nil
       end.compact
       commands | completion_data
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -193,8 +193,15 @@ module IRB
 
       # It doesn't make sense to propose commands with other preposing
       commands = [] unless preposing.empty?
+      @encoding_warning_shown ||= false
 
-      completion_data = retrieve_completion_data(target, bind: bind, doc_namespace: false).compact.map{ |i| i.encode(Encoding.default_external) }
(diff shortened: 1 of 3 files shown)


lib/irb/completion.rb
```diff
@@ -193,8 +193,15 @@ def completion_candidates(preposing, target, postposing, bind:)
 
       # It doesn't make sense to propose commands with other preposing
       commands = [] unless preposing.empty?
-
-      completion_data = retrieve_completion_data(target, bind: bind, doc_namespace: false).compact.map{ |i| i.encode(Encoding.default_external) }
+      @encoding_warning_shown ||= false
+
+      completion_data = retrieve_completion_data(target, bind: bind, doc_namespace: false).compact.map do |i|
+        i.encode(Encoding.default_external)
+      rescue Encoding::UndefinedConversionError
```
tompng, line 200: Can you add this `encode` and `rescue` to `IRB::TypeCompletor#completion_candidates`?
It is the default completor when `RUBY_VERSION >= '3.4'`

lib/irb/completion.rb
```diff
@@ -193,8 +193,15 @@ def completion_candidates(preposing, target, postposing, bind:)
...
+        warn "Warning: Invalid encoding in method name '#{i}'. can't be converted to the locale #{Encoding.default_external}." unless @encoding_warning_shown
+        @encoding_warning_shown = true
```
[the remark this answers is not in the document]
ksaito422: Fixed to ignore invalid encoding.
[f8ed14a](https://github.com/ruby/irb/pull/1101/commits/f8ed14af5c0a9984e21df23337056fddedbd3b1b)

lib/irb/completion.rb
ksaito422: Applied the same fix to IRB::TypeCompletor#completion_candidates.
[4897ccf](https://github.com/ruby/irb/pull/1101/commits/4897ccf15bca749f2ae943af95fc75efc662e572)

ksaito422:
The CI failures don't seem related.

tompng (approved):
Looks good :+1:

PR #1104, opened 2025-06-14
fix typos and wording on sigint section of docs

Stevo-S:
- fix typos where SIGINT is misspelt as SIGING
- fix wording on the docs to reflect that the
  session will, by default, not be exited when
  Ctrl-C is input


tompng (approved):
Thank you 😄

PR #1106, opened 2025-06-18
Remove all internal frames from a backtrace

mame:
In Ruby 3.5, the debug inspector API returns a raw backtrace, so irb/debug needs to remove not only `<internal:prelude>` but also all internal frames.


mame:
The CI fails because https://github.com/ruby/ruby/pull/13238 is not merged. But to merge that PR to ruby/ruby, this PR must be merged to ruby/irb...

Let me merge this now. Then I will merge that PR to ruby/ruby, and trigger ruby-dev-builder so that setup-ruby can fetch the new ruby master.

PR #1107, opened 2025-06-26
whereami: support -f to show the full file like pry

mddelk:
Support passing `-f` to `whereami` to show the full file, like `pry` does.

> [!NOTE]
> I added support for `-f`, but not for `--file`, since the only other IRB command I see that has flags only supports short forms (`history`).
>
> I haven't personally ever used `--file`.

Here's `pry`'s help page for comparison:

```
% pry
[1] pry(main)> help whereami                                                                                     
Usage: whereami [-qn] [LINES]

Describe the current location. If you use `binding.pry` inside a method then
whereami will print out the source for that method.

If a number is passed, then LINES lines before and after the current line will be
shown instead of the method itself.

The `-q` flag can be used to suppress error messages in the case that there's
no code to show. This is used by pry in the default before_session hook to show
you when you arrive at a `binding.pry`.

The `-n` flag can be used to hide line numbers so that code can be copy/pasted
effectively.

When pry was started on an Object and there is no associated method, whereami
will instead output a brief description of the current object.

    -q, --quiet                Don't display anything in case of an error
    -n, --no-line-numbers      Do not display line numbers
    -m, --method               Show the complete source for the current method.
    -c, --class                Show the complete source for the current class or module.
    -f, --file                 Show the complete source for the current file.
    -h, --help                 Show this message.
[2] pry(main)> Pry::VERSION                                                                                      
=> "0.14.2"
```

---

Thanks for IRB, btw!

### DIFF
--- a/lib/irb/command/whereami.rb
+++ b/lib/irb/command/whereami.rb
@@ -6,10 +6,13 @@ module IRB
   module Command
     class Whereami < Base
       category "Context"
-      description "Show the source code around binding.irb again."
+      description "Show the source code around binding.irb again. `-f` shows the full file."
 
-      def execute(_arg)
-        code = irb_context.workspace.code_around_binding
+      def execute(arg)
+        code = irb_context.workspace.code_around_binding(
+          show_full_file: arg.split.first == "-f"
+        )
+
         if code
           puts code
         else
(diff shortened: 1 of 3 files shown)


jethrodaniel:
CI failures seem unrelated.

PR #1110, opened 2025-07-21
Add tsort to dependencies

st0012:
tsort is becoming a bundled gem in Ruby 3.6 and Ruby 3.5 is already printing warnings about it. Since tsort supports Ruby 2.3+, it should be safe to add it to the dependencies.

Fix CI

### DIFF
--- a/irb.gemspec
+++ b/irb.gemspec
@@ -45,4 +45,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency "reline", ">= 0.4.2"
   spec.add_dependency "rdoc", ">= 4.0.0"
   spec.add_dependency "pp", ">= 0.6.0"
+  spec.add_dependency "tsort"
 end


ima1zumi:
IRB does not depend on `tsort`; this warning is coming from `rbs`. Therefore, I don't think `tsort` should be added to IRB's dependencies.

This will be resolved once https://github.com/ruby/rbs/pull/2601 is released.

```
<"/home/runner/work/irb/irb/vendor/bundle/ruby/3.5.0+2/gems/rbs-3.9.4/lib/rbs.rb:1\n" +
"1: warning: tsort was loaded from the standard library, but will no longer be pa\n" +
"rt of the default gems starting from Ruby 3.6.0\n" +
"You can add tsort to your Gemfile or gemspec to silence this warning.\n" +
```

st0012:
Ah good call. Should’ve checked more carefully. Thanks!

PR #1115, opened 2025-08-02
Handle SyntaxError#message with invalid encoding

alpaca-tc:
related: https://bugs.ruby-lang.org/issues/21528

This patch fixes an issue in IRB where it fails to handle a `SyntaxError#message` that has invalid encoding.

The root cause is that prism may generate a `SyntaxError#message` with broken encoding when the source contains multibyte characters. This patch ensures irb can handle such cases gracefully.

However, if this is considered a bug in Ruby itself, this pull request does not need to be merged.

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -73,6 +73,48 @@ module IRB
       yield
     ]
 
+    SYNTAX_ERROR_PATTERNS = {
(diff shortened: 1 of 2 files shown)


alpaca-tc:
Ruby is fixed
https://bugs.ruby-lang.org/issues/21528#change-114573

PR #1117, opened 2025-08-13
Gemfile: add github dependency on rbs

olleolleolle:
...until https://github.com/ruby/rbs/commit/14fd824ebde3a5c2403126fa2df11394d3b9b98f is in a release.

"Add explicit dependency on the tsort gem"

See https://github.com/ruby/reline/pull/847

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,5 +25,6 @@ gem "debug", github: "ruby/debug", platforms: [:mri, :mswin]
 gem "rdoc", ">= 6.11.0"
 
 if RUBY_VERSION >= "3.0.0" && !is_truffleruby
+  gem "rbs", github: "ruby/rbs" if RUBY_VERSION >= "3.2"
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -25,5 +25,6 @@ gem "debug", github: "ruby/debug", platforms: [:mri, :mswin]
 gem "rdoc", ">= 6.11.0"
 
 if RUBY_VERSION >= "3.0.0" && !is_truffleruby
+  gem "rbs", github: "ruby/rbs" if RUBY_VERSION >= "3.2"
```
st0012, line 28:
```suggestion
  # TODO: Remove this after rbs is released with tsort in its dependencies
  gem "rbs", github: "ruby/rbs" if RUBY_VERSION >= "3.2"
```

st0012 (approved):
Thanks!

PR #1127, opened 2025-10-17
Improve prompt generating performance by caching prompt parts(%m, %M)

### LINKED ISSUE #1126: Bug: `%m` in default IRB prompt can make the console unusably slow for objects with huge `to_s`/`inspect`
## Versions
irb: 1.15.2
debug: 1.11.0
Ruby: 3.3.6
OS: MacOS

## Summary
Using IRB’s default prompt:

```ruby
{
  PROMPT_I: "%N(%m):%03n> ",
  PROMPT_S: "%N(%m):%03n%l ",
  PROMPT_C: "%N(%m):%03n* ",
  RETURN:   "=> %s\n"
}
```

the `%m` expansion becomes extremely slow when the current context’s to_s/inspect is very large. In a real app with a “massive” object (its string representation is YAML-like and huge), every prompt render freezes for a long time.

## Why this happens

%m appears to derive its text from the current context receiver (ultimately calling to_s/inspect). If that method returns a very large string, printing it each prompt render overwhelms the console and makes IRB feel hung.

## Expected behavior

The default prompt should remain responsive even if self#to_s or self#inspect are slow or produce large output.

## Minimal reproduction

This script simulates a large to_s:
```ruby
# script.rb
require "irb"

class Liquidation
  def dummy_method
    binding.irb
  end

  def to_s
    "VeryLongClass\n" * 100_000_000 # large string
  end
end

Liquidation.new.dummy_method
```
or, if you prefer the debug console:
`bundle exec ruby script.rb`

You’ll notice the prompt becomes immediately sluggish, as %m tries to print the gigantic to_s on every render.

Video: 

https://github.com/user-attachments/assets/62d720d1-a22f-4558-bcff-80c75bfe1286

## Workarounds

Avoid %m in the prompt:
```ruby
IRB.conf[:PROMPT][:FAST] = {
  PROMPT_I: "%N:%03n> ",
  PROMPT_S: "%N:%03n%l ",
  PROMPT_C: "%N:%03n* ",
  RETURN:   "=> %s\n"
}
IRB.conf[:PROMPT_MODE] = :FAST
```

## Proposal

Consider changing the default %m behavior to a cheap, bounded representation, for example:

If self is a Module/Class: `self.name` (or fallback to `self.class.name`)

Otherwise: `self.class.name#object_id`

That keeps the default prompt fast while still being informative.

## Notes / Real-world context

In our application, the default prompt tries to render a massive YAML-like string for the current context, which makes the console “go crazy slow.” A screenshot for reference:

Image

tompng:
Fixes #1126

In prompt calculation, `main.to_s` was called on every keystroke and every line in multiline input.
Cache prompt parts(`%m` and `%M`) so that `main.to_s` is only called once per read-eval cycle.

Also fixes this problem.
```ruby
irb(main):001> def self.to_s = 'a' * rand(1..10)
=> :to_s
# before
irb(aaa):002* def f 
irb(aaaaaaa):003*   puts 'Hello'
irb(aa):004*   puts 'World'
irb(aaaa):005> end

# after
irb(aaaa):002* def f
irb(aaaa):003*   puts 'Hello'
irb(aaaa):004*   puts 'World'
irb(aaaa):005> end
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -621,6 +621,7 @@ module IRB
     end
 
     def format_prompt(format, ltype, indent, line_no) # :nodoc:
+      # @prompt_part_cache could be nil in unit tests
       part_cache = @prompt_part_cache || {}
       format.gsub(/%([0-9]+)?([a-zA-Z%])/) do
         case $2


lib/irb.rb
```diff
@@ -598,25 +602,28 @@ def generate_prompt(opens, continue, line_offset)
...
-      else
-        str[0, PROMPT_MAIN_TRUNCATE_LENGTH - PROMPT_MAIN_TRUNCATE_OMISSION.size] + PROMPT_MAIN_TRUNCATE_OMISSION
+      if str.size > PROMPT_MAIN_TRUNCATE_LENGTH
+        str = str[0, PROMPT_MAIN_TRUNCATE_LENGTH - PROMPT_MAIN_TRUNCATE_OMISSION.size] + PROMPT_MAIN_TRUNCATE_OMISSION
       end
+      str.tr(CONTROL_CHARACTERS_PATTERN, ' ')
     end
 
     def format_prompt(format, ltype, indent, line_no) # :nodoc:
+      part_cache = @prompt_part_cache || {}
```
st0012: Since `format_prompt` is only called in `generate_prompt`, is it possible to be called outside of `readmultiline` context? (perhaps in the `dyamic_prompt` callback?)

If it can be, let's also wrap the caller inside the wrapper mentioned above to provide caching?

lib/irb.rb
```diff
@@ -239,6 +240,7 @@ def read_input(prompt)
     end
 
     def readmultiline
+      @prompt_part_cache = {}
```
st0012: I think we can create a wrapper to ensure wherever we can generate prompt we have this set and cleaned. Like

```rb
with_prompt_part_cached do
  ...
end
```

lib/irb.rb
st0012: I see. Then let's at least add comments to explain why it could be nil, like

```suggestion
      # @prompt_part_cache could be nil in unit tests
      part_cache = @prompt_part_cache || {}
```

PR #1129, opened 2025-10-21
Fix UTF-16 autocompletion

### LINKED ISSUE #52: A method what is named with an encoding what is different from `Encoding.default_external` causes crash
1. input `define_method "a".encode(Encoding::UTF_16) do end` and Enter on UTF-8 terminal env
1. press Tab
1. crash

alexanderadam:
fixes #52

**PS:** I'm looking for a new adventure in case anybody is looking to hire or work with a Ruby/Rails/Crystal dev

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -275,7 +275,7 @@ module IRB
           nil
         else
           sym = $1
-          candidates = Symbol.all_symbols.collect do |s|
+          candidates = Symbol.all_symbols.filter_map do |s|
             s.inspect
           rescue EncodingError
             # ignore
@@ -453,6 +453,13 @@ module IRB
         else
           candidates = (bind.eval_methods | bind.eval_private_methods | bind.local_variables | bind.eval_instance_variables | bind.eval_class_constants).collect{|m| m.to_s}
           candidates |= RubyLex::RESERVED_WORDS.map(&:to_s)
+
+          target_encoding = Encoding.default_external
+          candidates = candidates.compact.filter_map do |candidate|
+            candidate.encoding == target_encoding ? candidate : candidate.encode(target_encoding)
+          rescue EncodingError
+            nil
+          end
           candidates.grep(/^#{Regexp.quote(input)}/).sort
         end
       end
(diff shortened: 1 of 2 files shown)


lib/irb/completion.rb
```diff
@@ -453,16 +453,30 @@ def retrieve_completion_data(input, bind:, doc_namespace:)
...
       end
     end
 
     # Set of available operators in Ruby
     Operators = %w[% & * ** + - / < << <= <=> == === =~ > >= >> [] []= ^ ! != !~]
 
+    def safe_grep(candidates, pattern)
+      target_encoding = Encoding.default_external
+      candidates.filter_map do |candidate|
+        next unless candidate
```
tompng: I think this part is not needed because all candidate passed to `safe_grep` are converted to string with `.collect{|m| m.to_s}`

lib/irb/completion.rb
```diff
@@ -453,16 +453,30 @@ def retrieve_completion_data(input, bind:, doc_namespace:)
...
+
+        converted = candidate.encoding == target_encoding ? candidate : candidate.encode(target_encoding)
```
tompng: Encoding conversion is more than safe-grep. I think this method should only do safe grep (`match?` and `rescue`) for simplicity.

Encoding conversion is already done in line 205.
```ruby
def completion_candidates(preposing, target, postposing, bind:)
  ...
  completion_data = retrieve_completion_data(target, bind: bind, doc_namespace: false).compact.filter_map do |i|
    i.encode(Encoding.default_external)
  rescue Encoding::UndefinedConversionError
    # If the string cannot be converted, we just ignore it
    nil
  end
  ...
end
```

lib/irb/completion.rb
alexanderadam: I was surprised as well but without this  the tests fail on TruffleRuby (compare the builds on the two commits in this PR).

It's working on every other platform though. 🤔
Or was I missing a bit here or misunderstanding something?

/CC @eregon

lib/irb/completion.rb
```diff
@@ -453,16 +453,30 @@ def retrieve_completion_data(input, bind:, doc_namespace:)
...
+
+        converted if pattern.match?(converted)
+      rescue Encoding::UndefinedConversionError, Encoding::InvalidByteSequenceError, Encoding::CompatibilityError, EncodingError
```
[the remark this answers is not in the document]
alexanderadam: Ah, that absolutely makes sense! Thank you so much! :pray:

PR #1132, opened 2025-10-28
update test to check for UTF16LE/BE

alexanderadam:
This PR just extends the test for UTF-16 LE/BE as well.

refs #52 and https://github.com/ruby/irb/pull/1129#discussion_r2469918480

**PS:** I'm looking for a new adventure in case anybody is looking to hire or work with a Ruby/Rails/Crystal dev

### DIFF
--- a/test/irb/test_completion.rb
+++ b/test/irb/test_completion.rb
@@ -345,13 +345,10 @@ module TestIRB
     end
 
     def test_utf16_method_name_does_not_crash
-      if RUBY_ENGINE == 'truffleruby'
-        omit "TruffleRuby does not support UTF-16 methods."
-      end
       # Reproduces issue #52: https://github.com/ruby/irb/issues/52
-      method_name = "test_utf16_method".encode(Encoding::UTF_16)
       test_obj = Object.new
-      test_obj.define_singleton_method(method_name) {}
+
+      test_obj.define_singleton_method("test_utf16le_method".encode(Encoding::UTF_16LE)) {}
       test_bind = test_obj.instance_eval { binding }
 
       completor = IRB::RegexpCompletor.new
@@ -360,7 +357,7 @@ module TestIRB
         result = completor.completion_candidates('', 'test', '', bind: test_bind)
       end
 
-      assert_include result, "test_utf16_method"
+      assert_include result, "test_utf16le_method"
     end
   end
 end


test/irb/test_completion.rb
```diff
@@ -345,13 +345,15 @@ def test_regexp_completor_handles_encoding_errors_gracefully
...
-      method_name = "test_utf16_method".encode(Encoding::UTF_16)
+      # Tests UTF-16, UTF-16LE, and UTF-16BE encodings
       test_obj = Object.new
-      test_obj.define_singleton_method(method_name) {}
+
+      if RUBY_ENGINE != 'truffleruby'
+        test_obj.define_singleton_method("test_utf16_method".encode(Encoding::UTF_16)) {}
+      end
+      test_obj.define_singleton_method("test_utf16le_method".encode(Encoding::UTF_16LE)) {}
+      test_obj.define_singleton_method("test_utf16be_method".encode(Encoding::UTF_16BE)) {}
```
tompng: Only testing one of `UTF_16LE` or `UTF_16BE` is enough.

As mentioned in https://github.com/ruby/irb/pull/1129#discussion_r2468591736, `UTF_16` is a dummy encoding. Testing UTF_16 is not so important. There's a benefit removing UTF_16 test: removing truffleruby condition.

Here's a list of encodings that cause error. It is not restricted to UTF_16 family.
```
UTF-16BE	UTF-16LE	UTF-32BE	UTF-32LE	UTF-16	UTF-32	CP949	Emacs-Mule
EUC-TW	Windows-1258	GB1988	macCentEuro	macThai	IBM037	stateless-ISO-2022-JISO-2022-JP
CP50220	CP50221	MacJapanese	ISO-2022-JP-KDDI	stateless-ISO-2022-JP-KDDI
```
Found by this script
```ruby
Encoding.list.each do |encoding|
  obj = Object.new
  (128..0xffff).each do |c|
    obj.define_singleton_method('a'.encode(encoding) + c.chr(encoding)){}
  rescue EncodingError, RangeError
  end
  IRB::RegexpCompletor.new.completion_candidates('', 'obj.a', '', bind: binding) rescue puts encoding.name
end
```
I don't think it's worth testing all of these encodings.
So I think selecting one of them (`UTF_16LE` or `UTF_16BE`) is enough.

Issue #960, opened 2024-05-24
IRB crashes when using `define_method('class') {}`


koic:
## Description

I was expecting no errors. However, the following error occurs:

```console
$ irb
irb(main):001> define_method('class') { puts 'hi' }
=> :class
hi
hi
hi
/Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb/history.rb:38:in `save_history': nil is not a class/module (TypeError)

      history = self.class::HISTORY.to_a
                          ^^^^^^^^^
        from /Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb.rb:1031:in `ensure in run'
        from /Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb.rb:1032:in `run'
        from /Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb.rb:904:in `start'
        from /Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/exe/irb:9:in `<top (required)>'
        from /Users/koic/.rbenv/versions/3.3.1/bin/irb:25:in `load'
        from /Users/koic/.rbenv/versions/3.3.1/bin/irb:25:in `<main>'
/Users/koic/.rbenv/versions/3.3.1/lib/ruby/3.3.0/ripper/lexer.rb:80:in `&': undefined method `new' for nil (NoMethodError)

      def &(i) self.class.new(to_int & i) end
                         ^^^^
        from /Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb/color.rb:220:in `dispatch_seq'
        from /Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb/color.rb:151:in `block (2 levels) in colorize_code'
        from /Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb/color.rb:149:in `each_line'
        from /Users/koic/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/irb-1.13.1/lib/irb/color.rb:149:in `block in colorize_code'
[...]

Issue #143, opened 2020-11-30
irb should not require standard libraries by default (if possible)


JunichiIto:
In Ruby 2.6.6, you cannot use Etc constant without requring etc library:

```
irb(main):001:0> RUBY_DESCRIPTION
=> "ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-darwin19]"
irb(main):002:0> Etc
Traceback (most recent call last):
        4: from /Users/jnito/.rbenv/versions/2.6.6/bin/irb:23:in `<main>'
        3: from /Users/jnito/.rbenv/versions/2.6.6/bin/irb:23:in `load'
        2: from /Users/jnito/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        1: from (irb):2
NameError (uninitialized constant Etc)
```

But in Ruby 2.7.2, you can do it:

```
irb(main):001:0> RUBY_DESCRIPTION
=> "ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]"
irb(main):002:0> Etc
=> Etc
```

However, you cannot use Etc constant in plain script context:

```
$ ruby -e "puts RUBY_DESCRIPTION; Etc"
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]
Traceback (most recent call last):
-e:1:in `<main>': uninitialized constant Etc (NameError)
```

Etc is defined not in core API but in standard library. So Ruby does not require it by default, but irb does in Ruby 2.7. This inconsistent behavior confuses newbies. Actually I was asked such a question today.

As far as I investigated, irb requires some other libraries like fileutils, set and so on. But I think irb should run in the same context as plain script, if possible.

rubyFeedback:
I sort of agree with JunichiIto; IRB should ideally be as close as possible to "default" ruby behaviour, as-is.

People can customize the behaviour via irbrc, which I do too - my irbrc contains lots of load-ruby-code
behaviour. :)

sos4nt:
`Set` seems to be required by RubyGems, or more specifically, by its Molinillo dependency resolver:

[...]

PR #1140, opened 2025-11-18
[lib/irb] Explicitly join on "" so $, does not change error message outputs

sampersand:
This PR does a minor change to `lib/irb.rb` so that the `.join` done on error messages doesn't use `$,`. I use `$,` occasionally when debugging some older scripts, and having IRB's error messages use it is a bit annoying.

I've made [another PR](https://github.com/ruby/reline/pull/866/files) to `reline` which solves a much more egregious problem (it does this joining when printing chars out in prompts!)

# Before
```
% irb
irb(main):001> def (Reline::Unicode).escape_for_print(str) # required until my PR for reline is merged
irb(main):002*   str.chars.map! { |gr|
irb(main):003*     case gr
irb(main):004*     when -"\n"
irb(main):005*       gr
irb(main):006*     when -"\t"
irb(main):007*       -'  '
irb(main):008*     else
irb(main):009*       Reline::Unicode::EscapedPairs[gr.ord] || gr
irb(main):010*     end
irb(main):011*   }.join('')
irb(main):012* end
=> :escape_for_print
irb(main):013> $, = '|'
=> "|"
irb(main):014> fail "oops"
(irb):14:in '<main>': oops (RuntimeError)
  from <internal:kernel>:168:in 'Kernel#loop'
| from /Users/me/.rbenv/versions/3.5.0-preview1/lib/ruby/gems/3.5.0+0/gems/irb-1.15.2/exe/irb:9:in '<top (required)>'
| from /Users/me/.rbenv/versions/3.5.0-preview1/bin/irb:25:in 'Kernel#load'
| from /Users/me/.rbenv/versions/3.5.0-preview1/bin/irb:25:in '<main>'
```

# After
```
irb(main):001> def (Reline::Unicode).escape_for_print(str) # required until my PR for reline is merged
irb(main):002*   str.chars.map! { |gr|
irb(main):003*     case gr
irb(main):004*     when -"\n"
irb(main):005*       gr
irb(main):006*     when -"\t"
irb(main):007*       -'  '
irb(main):008*     else
irb(main):009*       Reline::Unicode::EscapedPairs[gr.ord] || gr
irb(main):010*     end
irb(main):011*   }.join('')
irb(main):012* end
=> :escape_for_print
irb(main):013> $, = '|'
=> "|"
irb(main):014> fail "oops"
(irb):14:in '<main>': oops (RuntimeError)
  from <internal:kernel>:168:in 'Kernel#loop'
  from /Users/me/.rbenv/versions/3.5.0-preview1/lib/ruby/gems/3.5.0+0/gems/irb-1.15.2/exe/irb:9:in '<top (required)>'
  from /Users/me/.rbenv/versions/3.5.0-preview1/bin/irb:25:in 'Kernel#load'
  from /Users/me/rbenv/versions/3.5.0-preview1/bin/irb:25:in '<main>'
```

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -430,7 +430,7 @@ module IRB
           end
         end
         lines = lines.reverse if order == :bottom
-        lines.map{ |l| l + "\n" }.join
+        lines.map{ |l| l + "\n" }.join('')
       }
       # The "<top (required)>" in "(irb)" may be the top level of IRB so imitate the main object.
       message = message.gsub(/\(irb\):(?<num>\d+):in (?<open_quote>[`'])<(?<frame>top \(required\))>'/) { "(irb):#{$~[:num]}:in #{$~[:open_quote]}<main>'" }


sampersand:
cf https://github.com/ruby/reline/pull/866 for the reline one

tompng:
There are more `.join`, `print(*args)` (that uses `$,` for joiner character) to be fixed.
IRB depends on many other standard libraries (rdoc, reline, pp, repl_type_completor, rbs, json, and more) and it may not work if `$,` is overridden.
If you think this is a problem to be solved, all standard libraries should be `$,` override safe, please open an issue to [bugs.ruby-lang.org](https://bugs.ruby-lang.org/)

One way to fully support overriding `$,` is to set `$,` to `nil` after `eval`.
```ruby
module IRB
  class WorkSpace
    def evaluate(statements, file = __FILE__, line = __LINE__)
      $, = somewhere.stored_dollar_comma
      eval(statements, @binding, file, line)
      somewhere.stored_dollar_comma = $,
      $, = nil
    end
  end
end
```

But I'm not confident if we should really support it. Changing `$,` is deprecated and has a serious side effect.
Rubyists can do it in their own risk and IRB gets unusable is one of the risk.
Because of this reason, some related issues are closed as not planned https://github.com/ruby/irb/issues/143 https://github.com/ruby/irb/issues/235 https://github.com/ruby/irb/issues/960

For what is worth, Ruby 4.0 ships with Ruby::Box that can make a sandbox-like environment.
Here's a pull request that adds an command line option to eval inside Ruby::Box https://github.com/ruby/irb/pull/1142

PR #1141, opened 2025-11-19
Correct usage for -w. It turns ON warnings, not OFF.

zenspider:


zenspider:
Gemfile commit dropped. Rebased clean

PR #1142, opened 2025-11-22
Add `--context-mode 5`: evaluates inside Ruby::Box

tompng:
A playground to try monkey patching core methods.
```
% RUBY_BOX=1 irb --context-mode 5          
ruby: warning: Ruby::Box is experimental, and the behavior may change in the future!
See doc/box.md for known issues, etc.
Context-mode-5 (binding in new Ruby::Box) is experimental. It may be removed or changed without notice.
irb(main):001> defined? IRB
=> nil
irb(main):002* class Integer
irb(main):003*   def +(*) = 42
irb(main):004> end
=> :+
irb(main):005> 1 + 2
=> 42
irb(main):006> 
```

Related to: #143 #235 #960

### DIFF
--- a/test/irb/test_irb.rb
+++ b/test/irb/test_irb.rb
@@ -963,4 +963,34 @@ module TestIRB
       end
     end
   end
+
+  class ContextModeTest < TestIRB::IntegrationTestCase
+    # RUBY_BOX=1 may need more time to start IRB
+    TIMEOUT_SEC = TestIRB::IntegrationTestCase::TIMEOUT_SEC + 5
+
+    def test_context_mode_ruby_box
+      omit if RUBY_VERSION < "4.0.0"
+      @envs['RUBY_BOX'] = '1'
+
+      write_rc <<~'RUBY'
+        IRB.conf[:CONTEXT_MODE] = 5
+      RUBY
+
+      write_ruby <<~'RUBY'
+        require 'irb'
+        puts 'binding.irb' # Mark for breakpoint trigger
+        IRB.start
+        p answer2: 1 + 2
+      RUBY
+
+      output = run_ruby_file(timeout: TIMEOUT_SEC) do
+        type 'class Integer; def +(_other) = 42; end'
+        type 'p answer1: 1 + 2'
+        type 'exit'
+      end
+
+      assert_include(output, '{answer1: 42}')
+      assert_include(output, '{answer2: 3}')
+    end
+  end
 end
--- a/test/irb/helper.rb
+++ b/test/irb/helper.rb
@@ -139,7 +139,7 @@ module TestIRB
       end
     end
 
-    def run_ruby_file(&block)
+    def run_ruby_file(timeout: TIMEOUT_SEC, &block)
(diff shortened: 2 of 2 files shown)


st0012:
Can we have a few simple test cases running with this mode on? It can be used to detect `Ruby::Box`'s regression and we can also proactively disable it if it becomes noisy.

lib/irb/workspace.rb
```diff
@@ -55,6 +55,9 @@ def initialize(*main)
           # Note that this will typically be IRB::TOPLEVEL_BINDING
           # This is to avoid RubyGems' local variables (see issue #17623)
           @binding = TOPLEVEL_BINDING.dup
+        when 5
+          puts 'Context-mode-5 (binding in new Ruby::Box) is experimental. It may be removed or changed without notice.'
+          @binding = Ruby::Box.new.eval('Kernel.binding')
```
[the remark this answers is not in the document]
tompng: Add a `defined?(Ruby::Box)` check.
I think if Ruby::Box is unavailable, it's OK or even safe to fail instead of fallback to default.

There is one more error case. If Box is not enabled by `ENV['RUBY_BOX']`, `class Ruby::Box` is defined but calling `new` throws RuntimeError.
```
'Ruby::Box#initialize': Ruby Box is disabled. Set RUBY_BOX=1 environment variable to use Ruby::Box. (RuntimeError)
```

lib/irb/workspace.rb
```diff
@@ -55,6 +55,15 @@ def initialize(*main)
           # Note that this will typically be IRB::TOPLEVEL_BINDING
           # This is to avoid RubyGems' local variables (see issue #17623)
           @binding = TOPLEVEL_BINDING.dup
+
+        when 5  # binding in Ruby::Box
+          unless defined?(Ruby::Box)
+            puts 'Context-mode 5 (binding in Ruby::Box) requires Ruby 4.0 or later.'
+            raise NameError, 'Ruby::Box not defined'
```
tompng: Immediate `exit` will make `IRB.start rescue do_something` not work as expected

test/irb/test_irb.rb
```diff
@@ -963,4 +963,34 @@ def bar
       end
     end
   end
+
+  class ContextModeTest < TestIRB::IntegrationTestCase
+    # RUBY_BOX=1 may need more time to start IRB
+    TIMEOUT_SEC = TestIRB::IntegrationTestCase::TIMEOUT_SEC + 5
```
st0012: I think this can be 
```suggestion
    TIMEOUT_SEC = self::TIMEOUT_SEC + 5
```

```rb
class Foo
  TIMEOUT = 10
end

class Bar < Foo
  TIMEOUT = self::TIMEOUT + 10
end

puts Bar::TIMEOUT # 20
```

tompng: Actually, it can be even shorter
```ruby
TIMEOUT = TIMEOUT + 5
```
or 
```ruby
TIMEOUT += 5
```
but the visual of the code, especially the last one, looks like a constant reassignment. (It's not, though)
So my code is... a bit redundant but I'd like to use the redundant form here.

PR #1155, opened 2025-12-28
Remove deprecated toplevel ::RubyLex

### LINKED ISSUE #1154: Fix RubyLex deprecation warning
## Why

Hello. :wave: I'm running into deprecation warnings when using IRB where I see the following in my console:

``` 
warning: constant ::RubyLex is deprecated
```

## How

The fastest way to reproduce this is by launching IRB from your console and running the following (ensure Ruby warnings are enabled too):

``` 
Object.constants.find { Object.const_get(it) == Ruby }
```

## Notes

I'm using the following environment:

- **Ruby**: ruby 4.0.0 (2025-12-25 revision 553f1675f3) +YJIT +PRISM [arm64-darwin24.6.0]
- **IRB**: 1.16.0

tompng:
It was deprecated in `v1.8.0`.
Closes #1154

### DIFF
--- a/lib/irb/ruby-lex.rb
+++ b/lib/irb/ruby-lex.rb
@@ -496,6 +496,3 @@ module IRB
   end
   # :startdoc:
 end
-
-RubyLex = IRB::RubyLex
-Object.deprecate_constant(:RubyLex)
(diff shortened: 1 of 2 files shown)


st0012:
Technically speaking, removing a deprecated constant is still a breaking change. And if we were to follow semver, we need to release this change in v2.0.0.

Since there are [other things](https://github.com/search?q=repo%3Aruby%2Firb%20deprecated&type=code) being deprecated in IRB now. Maybe we can do a cleanup and cut v2.0? I've created an issue to have more discussion: https://github.com/ruby/irb/issues/1156

PR #1157, opened 2026-01-13
Add reload command to re-load files loaded via require

ahogappa:
# Summary

  Add reload command to IRB that reloads files loaded via `require`, `require_relative`, or `autoload` during an IRB session.

# Motivation

  When developing gems and debugging/testing with IRB, it's tedious to restart IRB every time you modify a file. This feature allows you to reload modified files without restarting the session.

# Implementation

  Uses Ruby::Box's sandboxed `$LOADED_FEATURES` to detect which files are loaded by require. The basic flow is:

  1. When `require` is called from the IRB prompt, execute it inside a new `Ruby::Box`
  2. Compare `$LOADED_FEATURES` before and after to detect newly loaded files
  3. Track the detected Ruby files for later reloading
  4. Load the files into the main environment using `load` (native extensions use `require`)
  5. When reload command is executed, re-load all tracked files

# Usage
## Prerequisites

  - Ruby 4.0+ with `RUBY_BOX=1` environment variable
  - `IRB.conf[:RELOADABLE_REQUIRE] = true` in your .irbrc

## Example
```rb
# .irbrc
IRB.conf[:RELOADABLE_REQUIRE] = true

# In IRB session
irb> require 'mylib'
irb> MyLib.hello  # => "original"
# ... edit mylib.rb ...
irb> reload
Reloaded: /path/to/mylib.rb
irb> MyLib.hello  # => "modified"
```
## Limitations

  - Constant redefinition warnings will appear on reload (uses load internally)
  - Native extensions (.so, .bundle) cannot be reloaded
  - Files loaded via `Ruby::Box#require` directly are not tracked
  - Context mode 5 (IRB running inside a `Ruby::Box`) is not supported
    - These could be implemented in future PRs if there's demand

### DIFF
--- a/lib/irb/command/reload.rb
+++ b/lib/irb/command/reload.rb
@@ -4,17 +4,42 @@ module IRB
   # :stopdoc:
 
   module Command
-    class Reload < Base
+    class BoxReload < Base
       category "IRB"
-      description "Reload files that were loaded via require in IRB session."
+      description "[Experimental] Reload files that were loaded via require in IRB session (requires Ruby::Box)."
+
+      help_message <<~HELP
+        Usage: box_reload
+
+        Reloads all Ruby files that were loaded via `require` or `require_relative`
+        during the current IRB session. This allows you to pick up changes made to
+        source files without restarting IRB.
+
+        Setup:
+          1. Start Ruby with RUBY_BOX=1 environment variable
+          2. Set IRB.conf[:RELOADABLE_REQUIRE] = true in your .irbrc
+
+        Example:
+          # In .irbrc:
+          IRB.conf[:RELOADABLE_REQUIRE] = true
+
+          # In IRB session:
+          require 'my_lib'  # loaded and tracked
+          # ... edit my_lib.rb ...
+          box_reload         # reloads the file
+
+        Note: This feature is experimental and requires Ruby::Box (Ruby 4.0+).
+        Native extensions (.so/.bundle) cannot be reloaded.
+      HELP
 
       def execute(_arg)
         unless reloadable_require_available?
-          warn "The reload command requires IRB.conf[:RELOADABLE_REQUIRE] = true and Ruby::Box (Ruby 4.0+) with RUBY_BOX=1 environment variable."
+          warn "box_reload requires IRB.conf[:RELOADABLE_REQUIRE] = true and Ruby::Box (Ruby 4.0+) with RUBY_BOX=1 environment variable."
           return
         end
 
-        files = IRB.conf[:__RELOADABLE_FILES__]
+        ReloadableRequire.collect_autoloaded_files
+        files = ReloadableRequire.reloadable_files
         if files.empty?
           puts "No files to reload. Use require to load files first."
           return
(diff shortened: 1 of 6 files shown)


lib/irb/default_commands.rb
```diff
@@ -252,6 +253,7 @@ def load_command(command)
 
     register(:cd, Command::CD)
     register(:copy, Command::Copy)
+    register(:reload, Command::Reload)
```
ahogappa: The command name `reload` might be confused with Rails console's `reload!`. Consider alternatives like `reload_requires` or `refresh`.

lib/irb/command/reload.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module IRB
+  # :stopdoc:
+
+  module Command
+    class Reload < Base
+      category "IRB"
+      description "Reload files that were loaded via require in IRB session."
```
st0012: We should mark this as experimental.
And this command should have detailed `help_mesasge` to explain what it's for and how to set things up to run it.

lib/irb/default_commands.rb
st0012: I think the name should have `box` in it so users understand it's closely associated with that language feature.

lib/irb/init.rb
```diff
@@ -197,6 +199,10 @@ def IRB.init_config(ap_path)
     }
 
     @CONF[:COPY_COMMAND] = ENV.fetch("IRB_COPY_COMMAND", nil)
+
+    @CONF[:RELOADABLE_REQUIRE] = false
+    @CONF[:__RELOADABLE_FILES__] = Set.new
+    @CONF[:__AUTOLOAD_FILES__] = Set.new
```
st0012: Can we not put these in CONF? It's already bloated with things unrelated configs. We shouldn't use this as a cache storage.

lib/irb/reloadable_require.rb
```diff
@@ -0,0 +1,115 @@
...
+      ruby_files.each { |path| current_box.load(path) }
+
+      result
+    end
+  end
+
+  unless Ruby::Box.method_defined?(:__irb_original_require__)
+    Ruby::Box.class_eval do
+      alias_method :__irb_original_require__, :require
+      alias_method :__irb_original_require_relative__, :require_relative
```
st0012: Can we avoid this old alias_method_chain style?

st0012 (changes requested):

lib/irb/reloadable_require.rb
ahogappa: fix 798a78c

lib/irb/init.rb
ahogappa: Thank you. I wasn't sure where to put them, so I placed them in CONF. Fixed in b1e5c44.

lib/irb/command/reload.rb
ahogappa: fix 3fcbb5c

lib/irb/default_commands.rb
ahogappa: Thank you. I simply renamed it to `box_reload`.

PR #1163, opened 2026-01-20
Add missing keyword to colorize

tompng:
Add missing `KEYWORD_*` token need to be colored green.


tompng:
I forgot to include these keyword tokens in #1091. Bug only exist in unreleased master branch.

PR #1174, opened 2026-02-18
Add IRB `fix` command to rerun with corrected spelling (references ruby/did_you_mean#179)

krsinghshubham:
# Add IRB `fix` command to rerun with corrected spelling (references ruby/did_you_mean#179)

## Summary

Adds a native `fix` command to IRB that reruns the previous command with corrected spelling when a "Did you mean?" error occurs. This implements the feature requested in [ruby/did_you_mean#179](https://github.com/ruby/did_you_mean/issues/179).

## Example

```ruby
irb(main):001> ac
# => NameError (undefined local variable or method 'ac' for main)
#    Did you mean?  acc
#    Type `fix` to rerun with the correction.

irb(main):002> fix
# Rerunning with: acc
# => ...
```

```ruby
irb(main):001> 1.zeor?
# => NoMethodError (undefined method `zeor?' for 1:Integer)
#    Did you mean?  zero?
#    Type `fix` to rerun with the correction.

irb(main):002> fix
# Rerunning with: 1.zero?
# => true
```

### The Problem

The `fix` command was originally implemented as an extension in the `did_you_mean` gem via PR ruby/did_you_mean#204 ([https://github.com/ruby/did_you_mean](https://github.com/ruby/did_you_mean)). However, this implementation introduced a load order problem.

- `did_you_mean` is a default gem—Ruby loads it at boot, before any application code
- IRB loads later (e.g., when running `rails console` or `irb`)
- When `did_you_mean` loads, IRB isn't defined yet, so the extension never loads
- Result: The fix command didn't work in Rails console or when IRB loads after app initialization

### Why IRB Should Own This Feature

1. **It's a REPL feature** – The fix command is an IRB command (like `exit`, `load`, `help`). It belongs in the REPL layer.

2. **IRB has the context** – IRB already has the eval loop, exception handling, and access to the last executed code. It's the natural place for this logic.

3. **did_you_mean provides the API** – The [did_you_mean](https://github.com/ruby/did_you_mean) gem adds `#corrections` to exceptions (NameError, NoMethodError, KeyError, etc.). IRB simply uses this public API—no patching or extension loading required.

4. **No load order issues** – When IRB runs, both IRB and did_you_mean are loaded. IRB can check `exception.respond_to?(:corrections)` and use the fix command natively.

5. **Clean separation of concerns** – did_you_mean = spell-checking. IRB = REPL behavior. Each gem does one thing well.

### What IRB Gets from did_you_mean

IRB uses only the **public API** that did_you_mean provides:

- `exception.respond_to?(:corrections)` – Check if the exception has suggestions
- `exception.corrections` – Get the suggested corrections (e.g., `["zero?"]`)
- `DidYouMean::Levenshtein.distance` – Optional; used for edit-distance filtering when available. If did_you_mean isn't loaded (e.g., `--disable-did_you_mean`), the fix command still works but skips the distance filter.

No internal APIs, no patching, no extension loading.

### Files Changed

- **`lib/irb/command/fix.rb`** (new) – Fix command and LastError storage
- **`lib/irb.rb`** – Store last correctable exception, show discoverability hint
- **`lib/irb/default_commands.rb`** – Register `fix` and `dym` commands

### Supported Error Types

NoMethodError, NameError, KeyError, NoMatchingPatternKeyError, LoadError

### Correction Logic

- Single suggestion only (edit distance ≤ 2 when did_you_mean is available)
- Replacement patterns: identifiers, symbols, strings
- `dym` alias for users who use `fix` as a variable

### Discoverability

When a fix is available, IRB shows: `Type \`fix\` to rerun with the correction.`

## References

- [ruby/did_you_mean#179](https://github.com/ruby/did_you_mean/issues/179) – Original feature request
- [thefuck](https://github.com/nvbn/thefuck) – Similar concept for shell commands

### DIFF
--- a/lib/irb.rb
+++ b/lib/irb.rb
@@ -234,7 +234,8 @@ module IRB
             end
             handle_exception(exc)
             if Command::Fix.fixable?
-              puts "\e[2mType `fix` to rerun with the correction.\e[0m"
+              hint = Command::Fix::HINT
+              puts Color.colorable? ? Color.colorize(hint, [:BOLD]) : hint
             end
             @context.workspace.local_variable_set(:_, exc)
           end
--- a/lib/irb/command/fix.rb
+++ b/lib/irb/command/fix.rb
@@ -9,7 +9,13 @@
 module IRB
   module Command
     class Fix < Base
+      # Maximum Levenshtein distance for accepting a correction (did_you_mean default).
       MAX_EDIT_DISTANCE = 2
+
+      HINT = "Type `fix` to rerun with the correction."
+      MSG_NO_PREVIOUS_ERROR = "No previous error with Did you mean? suggestions. Try making a typo first, e.g. 1.zeor?"
+      MSG_NOT_CORRECTABLE = "Last error is not correctable. The fix command only works with NoMethodError, NameError, KeyError, etc."
+      MSG_RERUNNING = "Rerunning with: %s"
 
       category "did_you_mean"
       description "Rerun the previous command with corrected spelling from Did you mean?"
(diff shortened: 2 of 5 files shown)


krsinghshubham:
> Please remove `PR_DESCRIPTION.md` at least. That file is unrelated for this repository.

Done

PR #1177, opened 2026-02-25
Make ls command work for BasicObjects

eikes:
When working with the `builder` gem I noticed that the irb inspection of the `Builder::XmlMarkup` instances was not working as expected. Upon further investigation it turns out, that the reason is that this class inherits from `BasicObject` directly and not from the more common `Object`. 

You can see the problem here:
```
4.0.1 :001 > ls BasicObject.new
/gems/irb-1.17.0/lib/irb/command/ls.rb:58:in 'IRB::Command::Ls#execute': undefined method 'class' for an instance of BasicObject (NoMethodError)

        klass  = (obj.class == Class || obj.class == Module ? obj : obj.class)
                     ^^^^^^
```

I have also created a new `ls` command test file containing a failing test which you can see below: https://github.com/ruby/irb/actions/runs/22390651625/job/64811518726 

The test is fixed with the second commit of this PR by using a slightly different way to inspect the `class`. `singleton_class` and `instance_variables` of an object. 

PS: The prism-latest test failure is because the syntax of ruby 4.1 allows trailing commas in method definitions and this change is addressed here: https://github.com/ruby/irb/pull/1178

### DIFF
--- a/lib/irb/command/ls.rb
+++ b/lib/irb/command/ls.rb
@@ -55,11 +55,13 @@ module IRB
 
         o = Output.new(grep: grep)
 
-        klass  = (obj.class == Class || obj.class == Module ? obj : obj.class)
+        klass = Kernel.instance_method(:class).bind(obj).call
+        is_class_or_module = klass == Class || klass == Module
+        klass = is_class_or_module ? obj : klass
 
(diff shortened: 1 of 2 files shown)


tompng (approved):
Thank you :+1:

PR #1199, opened 2026-04-09
Support syntax highlighting customization

shugo:
For example, put the following code in ~/.irbrc:

```ruby
Reline::Face.config(:syntax_highlighting) do |conf|
  conf.define :pseudo_variable,  foreground: "#4dacfd",  style: :bold
  conf.define :global_variable,  foreground: "#4dacfd",  style: :bold
  conf.define :constant,         foreground: "#ffa657",  style: :bold
  conf.define :comment,          foreground: "#abaeb1",  style: :bold
  conf.define :string_edge,      foreground: "#c1daec",  style: :bold
  conf.define :string_body,      foreground: "#c1daec"
  conf.define :symbol,           foreground: "#c1daec"
  conf.define :number,           foreground: "#4dacfd",  style: :bold
  conf.define :float,            foreground: "#4dacfd",  style: :bold
  conf.define :keyword,          foreground: "#f16636"
  conf.define :label,            foreground: "#4dacfd"
  conf.define :method_name,      foreground: "#a887e6",  style: :bold
  conf.define :message_name,     foreground: "#a887e6"
  conf.define :error,            foreground: "#f16636",  style: :negative
end
```

### DIFF
--- a/lib/irb/color.rb
+++ b/lib/irb/color.rb
@@ -18,103 +18,114 @@ module IRB # :nodoc:
     CYAN      = 36
     WHITE     = 37
 
-    # Following pry's colors where possible
-    TOKEN_SEQS = {
-      KEYWORD_NIL:        [CYAN, BOLD],
-      KEYWORD_SELF:       [CYAN, BOLD],
-      KEYWORD_TRUE:       [CYAN, BOLD],
-      KEYWORD_FALSE:      [CYAN, BOLD],
-      KEYWORD___FILE__:   [CYAN, BOLD],
-      KEYWORD___LINE__:   [CYAN, BOLD],
-      KEYWORD___ENCODING__: [CYAN, BOLD],
-      CHARACTER_LITERAL:  [BLUE, BOLD],
-      BACK_REFERENCE:     [GREEN, BOLD],
-      BACKTICK:           [RED, BOLD],
-      COMMENT:            [BLUE, BOLD],
-      EMBDOC_BEGIN:       [BLUE, BOLD],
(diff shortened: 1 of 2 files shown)


lib/irb/color.rb
```diff
@@ -326,6 +337,15 @@ def dispatch_alias_method_name(node)
...
       ]
 
+      # Normalize SGR sequences for existing test cases
+      def normalize_sgr(seq)
+        s = seq.sub(/\A\e\[0m/, "")
+        return s if s.match?(/\e\[(38|48|58);/) # Do not normalize extended colors
+        s.gsub(/\e\[([0-9;]+)m/) {
+          $1.split(/;/).map { |i| "\e[#{i}m" }.join
+        }
+      end
```
shugo, line 353: Instead, it may be better to fix test cases.

lib/irb/color.rb
```diff
@@ -348,3 +368,23 @@ def without_circular_ref(obj, seen:, &block)
...
+  conf.define :number,           foreground: :blue,     style: :bold
+  conf.define :float,            foreground: :magenta,  style: :bold
+  conf.define :keyword,          foreground: :green
+  conf.define :label,            foreground: :magenta
+  conf.define :method_name,      foreground: :cyan,     style: :bold
+  conf.define :message_name,     foreground: :cyan
+  conf.define :error,            foreground: :red,      style: :negative
+end
+
+IRB::Color.init
```
st0012, line 396: Why do we call this here too? Are we initializing twice?

shugo: It's called here just in case irb/color is required from applications other than irb.

PR #1200, opened 2026-04-12
Add --nobanner option to suppress startup banner

### LINKED ISSUE #1196: Add a command-line option to suppress the IRB startup banner
In #1183, a feature was added that displays a banner when IRB starts.

However, there are cases where avoiding extra output is desirable, so I would like to request a command-line option to suppress these messages.

For example, Python provides a "quiet mode" with `python -q`:

```
% python3
Python 3.11.15 (main, Mar  5 2026, 01:05:40) [Clang 19.1.7 (https://github.com/llvm/llvm-project.git llvmorg-19.1.7-0-gcd7080 on freebsd15
Type "help", "copyright", "credits" or "license" for more information.
>>> exit()
% python3 -q
>>> exit()
%
```

It would be helpful if IRB also had an option such as `--nobanner` to suppress the startup banner.

The actual option name is, of course, up to the maintainers; the examples above are just suggestions.

Thank you for your consideration.

st0012:
Add a `--nobanner` CLI option that suppresses the startup banner introduced in #1183.

Users can already disable it via `IRB.conf[:SHOW_BANNER] = false` in `.irbrc`, but this provides a command-line alternative for scripting and quick one-off usage.

Closes #1196

### DIFF
--- a/lib/irb/init.rb
+++ b/lib/irb/init.rb
@@ -346,6 +346,8 @@ module IRB # :nodoc:
         opt = $1 || argv.shift
         prompt_mode = opt.upcase.tr("-", "_").intern
         @CONF[:PROMPT_MODE] = prompt_mode
+      when "--nobanner"
+        @CONF[:SHOW_BANNER] = false
       when "--noprompt"
         @CONF[:PROMPT_MODE] = :NULL
       when "--script"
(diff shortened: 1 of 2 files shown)


ima1zumi (approved):
LGTM!

elfham:
Thank you for this change.

PR #1229, opened 2026-07-15
Keep completion alive when RDoc document retrieval fails

pocke:
## Summary

The autocompletion documentation dialog could crash the whole IRB session when loading RDoc documentation failed. This makes RDoc document retrieval failable: the error is shown in the dialog and completion (and the session) keeps working.

## Steps to reproduce

With `Ruby::Box` enabled, IRB crashes during completion (see the details below). The steps below reproduce it that way, but the crash can happen in any case where loading RDoc documentation fails, not only under `Ruby::Box`.

1. Use a Ruby build that provides `Ruby::Box` (e.g. a ruby-dev build) and start IRB with the box enabled:

   ```
   RUBY_BOX=1 irb
   ```

2. Type `1.` and press <Tab> to trigger method-name completion.

Any receiver reproduces it; the trigger is rendering the documentation dialog for a completion candidate, which loads RDoc data from disk.

## Expected behavior

The completion candidates and their documentation dialog are shown, and IRB keeps running even when the documentation cannot be loaded.

## Actual behavior

IRB terminates with an uncaught exception:

```
.../rdoc/store.rb:1190:in 'Marshal.load': undefined class/module RDoc:: (ArgumentError)
	from .../rdoc/store.rb:1190:in 'block in RDoc::Store#marshal_load'
	from .../rdoc/store.rb:835:in 'RDoc::Store#load_method'
	...
	from .../irb/input-method.rb:in 'IRB::RelineInputMethod#rdoc_dialog_contents'
	from .../reline/line_editor.rb:in 'Reline::LineEditor::Dialog#call'
	...
```

## Screen records

before:

https://github.com/user-attachments/assets/e5ae8e16-ba13-4a12-bd7d-98803b775bdb

after:

https://github.com/user-attachments/assets/598e7ba3-02e2-4d4a-af7d-bf918145a436

## Environment

```
irb(main):001> irb_info
Ruby version: 4.0.6
IRB version: irb 1.18.0 (2026-04-21)
InputMethod: RelineInputMethod with Reline 0.6.3
Completion: Autocomplete, ReplTypeCompletor: 0.1.12, Prism: 1.8.1, RBS: 3.10.0
.irbrc paths: /home/pocke/.irbrc
RUBY_PLATFORM: x86_64-linux
LANG env: C.UTF-8
East Asian Ambiguous Width: 1
```

## Root cause

The dialog loads RDoc `.ri` data via `Marshal.load`, and `rdoc_dialog_contents` only rescued `RDoc::RI::Driver::NotFoundError`. Any other error raised during retrieval propagated through Reline's dialog proc and terminated the REPL.

Here it is triggered by a Ruby core bug: under `Ruby::Box` (`RUBY_BOX=1`), `Marshal.load` fails to resolve class references and raises `ArgumentError: undefined class/module ...` (https://bugs.ruby-lang.org/issues/22090). That bug lives in Ruby itself and cannot be fixed here, but RDoc document retrieval is a fragile external dependency that IRB should treat as failable regardless of the specific cause.

## Fix

Treat only the fragile part -- fetching the document from the RDoc driver -- as failable, and leave rendering untouched.

- Extract the RDoc driver calls into `retrieve_rdoc_document`, and rescue around that call alone in `rdoc_dialog_contents`. `NotFoundError` still means "no document" and stays silent; any other `StandardError` is turned into an error document by `rdoc_error_document`.
- `RDoc::Markup::ToAnsi` rendering and the dialog layout stay outside the rescue on purpose: a failure there is IRB/Reline's own bug and should surface rather than be swallowed.
- The error document shows the exception class and message where the document would appear, so the user learns why documentation is unavailable while completion keeps working. The "Press Alt+d to read the full document" hint is omitted on error, since there is no full document to open.
- The narrow dialog does not show the backtrace. To investigate such a failure, run IRB with `-d` (which sets `$DEBUG`): the error is then re-raised with its full backtrace instead of being swallowed. The dialog points to this with a "Restart IRB with -d to see the backtrace." hint.

## Out of scope

The Alt+D full-document path (`display_document`) is left unchanged. In the error state there is no document to open, so pressing Alt+D behaves as before.

### DIFF
--- a/lib/irb/input-method.rb
+++ b/lib/irb/input-method.rb
@@ -404,12 +404,29 @@ module IRB
     end
 
     def rdoc_dialog_contents(name, width)
+      formatter = RDoc::Markup::ToAnsi.new
+      formatter.width = width
+
+      begin
+        document = retrieve_rdoc_document(name)
+      rescue RDoc::RI::Driver::NotFoundError
+        return
+      rescue => e
+        raise if $DEBUG
+        return rdoc_error_document(e).accept(formatter).split("\n")
+      end
+      return unless document
+
+      [PRESS_ALT_D_TO_READ_FULL_DOC] + document.accept(formatter).split("\n")
+    end
+
+    def retrieve_rdoc_document(name)
       driver = rdoc_ri_driver
       return unless driver
 
       name = driver.expand_name(name)
 
-      doc = if name =~ /#|\./
+      if name =~ /#|\./
         d = RDoc::Markup::Document.new
         driver.add_method(d, name)
         d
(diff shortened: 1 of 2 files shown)


tompng (approved):
Looks good.
This will also fixes RDoc's Marshal load related problem such as http://github.com/ruby/rdoc/pull/1549
Thank you :+1:

Issue #349, opened 2022-02-23
Completion: Suggestions seem to be in a strange order


rjschof:
## Description
When starting IRB (1.4.1) in Ruby 2.7+, the tab completion suggestions seem to be in a strange order: 

<img alt="Screen Shot 2022-01-08 at 12 22 23 PM" width="246" src="https://user-images.githubusercontent.com/3303032/148653612-e3dff786-1a10-4923-a0eb-3975cae10a7f.png">

The way this used to work, `require` used to appear before `require_relative` in the list of completion suggestions.

## Result of irb_info

```
irb(main):001:0> irb_info
Ruby version: 2.7.5                            
IRB version: irb 1.4.1 (2021-12-25)            
InputMethod: ReidlineInputMethod with Reline 0.3.1
RUBY_PLATFORM: arm64-darwin21                  
LANG env: en_US.UTF-8                          
East Asian Ambiguous Width: 1 
```

## Terminal Emulator
iTerm2

## Setting Files
None.

st0012:
I think this may have been fixed by https://github.com/ruby/irb/pull/379

<img width="252" alt="截圖 2022-10-09 16 22 48" src="https://user-images.githubusercontent.com/5079556/194765090-50af8312-a0ca-4722-906a-2b031c84f38b.png">

rjschof:
> I think this may have been fixed by https://github.com/ruby/irb/pull/379
> 
> 
> 
> <img width="252" alt="截圖 2022-10-09 16 22 48" src="https://user-images.githubusercontent.com/5079556/194765090-50af8312-a0ca-4722-906a-2b031c84f38b.png">
> 
> 

Thank you for the heads up!

PR #1230, opened 2026-07-15
Sort remaining completion candidates in RegexpCompletor

kryzhovnik:
#379 sorted local variables and string method candidates (fixing #349), but `receiver.method` completion — the most common case — still returns candidates in method table order, which looks random. Same for symbols, global variables and `::` constants. With an ActiveRecord model, `u.email_address` is listed below all its generated variants:

```
u.email_address_before_last_save
u.email_address_change_to_be_saved
u.email_address=
...
u.email_address
```

This sorts candidates in `select_message` (making the two pre-sorted call sites redundant) and in the symbol/gvar/absolute constant branches.

One detail: candidates starting with `_` are sorted last, otherwise completing `u.` would begin with `__id__`, `__send__`, `_run_commit_callbacks` and similar internals instead of regular methods. Same change for TypeCompletor is proposed in ruby/repl_type_completor#78.

### DIFF
--- a/lib/irb/completion.rb
+++ b/lib/irb/completion.rb
@@ -302,7 +302,7 @@ module IRB
           rescue EncodingError
             # ignore
           end
-          candidates.grep(/^#{Regexp.quote(sym)}/)
+          candidates.grep(/^#{Regexp.quote(sym)}/).sort
         end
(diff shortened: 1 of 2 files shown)


tompng (approved):
Looks good. Thank you :+1:

PR #1232, opened 2026-08-05
Accomodate new prism token

Earlopain:
Changed in https://github.com/ruby/prism/pull/4199.
Other token changes are irrelevant since they are already not used.

cc @kddnewton


tompng (approved):
Thank you :+1:

PR #1238, opened 2026-08-22
Show the startup banner only when stdin is a tty

simi:
The banner was printed for any `exe/irb` run without a script argument, so piped sessions like `echo '1+1' | irb` got it on stdout. The tip line is picked at random, which also made that output differ between runs.


tompng (approved):
:+1:

PR #1239, opened 2026-08-29
Honor explicit AUTO_INDENT false in prompt modes

OskarEichler:
## Summary

Read the prompt mode's `AUTO_INDENT` value after checking key presence, so an explicit `false` disables indentation instead of becoming `true`.

## Reproduction

A prompt configuration containing `AUTO_INDENT: false` currently stores the result of `Hash#include?`, which is `true`. The candidate stores the configured value.

## Verification

- RDoc and Rails 8.1.3.1 loading pass

## Compatibility

Only explicit false prompt configuration changes, to match its documented intent. Prepared with AI-assisted source review; no repository tests were changed.


tompng (approved):
:+1:

PR #1240, opened 2026-08-29
Remove temporary eval history configuration

OskarEichler:
## Summary

Delete the same temporary configuration key that eval-history initialization inserts.

## Reproduction

Initialization writes `IRB.conf[:__TMP__EHV__]` but deletes the mistyped `:__TMP_EHV__`, leaving the history object in global configuration after the workspace binding is established.

## Verification

- RDoc and Rails 8.1.3.1 loading pass

## Compatibility

Removes only an internal temporary key after its intended one-shot use. Prepared with AI-assisted source review; no repository tests were changed.


tompng (approved):
Nice catch :+1:

PR #1241, opened 2026-08-29
Resize partial evaluation histories safely

OskarEichler:
## Summary

Shrink evaluation history from the entries actually present rather than slicing with the configured maximum as an array index.

## Reproduction

When a history configured for 100 entries currently contains only three, resizing it to two slices from index 98 and replaces the contents with `nil`. The candidate retains the last two entries and does nothing when the populated history is already within the new limit.

## Verification

- focused partial-history grow/shrink model
- RDoc, candidate packaging and Rails 8.1.3.1 loading pass

## Compatibility

Unlimited and non-shrinking sizes remain unchanged; partial histories now follow the configured limit safely. Prepared with AI-assisted source review; no repository tests were changed.

### DIFF
--- a/lib/irb/ext/eval_history.rb
+++ b/lib/irb/ext/eval_history.rb
@@ -97,8 +97,8 @@ module IRB # :nodoc:
     end
 
     def size(size) # :nodoc:
-      if size != 0 && size < @size
-        @contents = @contents[@size - size .. @size]
+      if size.positive? && @contents.size > size
+        @contents = @contents.last(size)
       end
       @size = size
     end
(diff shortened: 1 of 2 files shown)


lib/irb/ext/eval_history.rb
```diff
@@ -97,8 +97,8 @@ def initialize(size = 16)  # :nodoc:
     end
 
     def size(size) # :nodoc:
-      if size != 0 && size < @size
-        @contents = @contents[@size - size .. @size]
+      if size.positive? && @contents.size > size
+        @contents = @contents.last(size)
```
tompng, line 101: Looks good :+1:
Can you add a test for this?
Adding a test that does the below in `test_eval_history.rb` would be enough
```ruby
out, err = execute_lines(
  "IRB.CurrentContext.eval_history = 2",
  "__",
  conf: { EVAL_HISTORY: 5 }
)
```

OskarEichler:
Added the suggested execute_lines regression for shrinking eval_history to 2 before the configured history of 5 is full. The full test_eval_history.rb file passes: 4 tests, 8 assertions.

tompng (approved):
Thank you

PR #1244, opened 2026-09-04
fix: remove debug message

### LINKED ISSUE #1227: Startup message about "debug" command is partially wrong
## Problem

One of the startup messages says: `'"debug" to start integration with the "debug" gem'`
Actually, `debug` command only works when IRB was started with `binding.irb`. The evidence is https://github.com/ruby/irb/blob/35b87cf8933f67cf6e9e0edd1ca7d6395595a758/lib/irb/command/debug.rb#L32-L34
So, technically speaking, that startup message could be wrong.

## Solution

One simple solution is just deleting that message.
Another solution could be mentioning the fact in that message, but I personally don't like it because the massage will be too long.

GabrielLidenor:
closes #1227 

First, I considered updating the startup message to `"debug" to start integration with the "debug" gem if irb started with binding.irb.`
However, as noted in the issue description, adding this text makes the startup banner overly verbose. I went with [removing the message / conditionally displaying it only when binding.irb is present] instead to keep the output clean.

### DIFF
--- a/lib/irb/startup_message.rb
+++ b/lib/irb/startup_message.rb
@@ -14,7 +14,6 @@ module IRB
       '"cd object" to navigate into an object',
       '"show_source method" to view source code',
       '"copy expr" to copy the output to clipboard',
-      '"debug" to start integration with the "debug" gem',
       '"history -g pattern" to search history',
     ].freeze
 


st0012 (approved):
Yeah I agree `debug` related commands are not ideal to show in the banner. Thanks for the fix.

PR #1246, opened 2026-09-05
Refactor pushws command

### LINKED ISSUE #1245: Refactor pushws command
### Motivation

- Fix the typo in the workspace command output collection: `inspection_resuls` should be `inspection_results`.
- Replace the magic number in `truncated_inspect` with a named constant so the truncation limit is clear and self-documenting.

GabrielLidenor:
closes #1245 

### Why
- Fix the typo in `inspection_results`
- Replace the magic number `20` in the workspace truncation logic with a named constant for clarity


lib/irb/command/pushws.rb
```diff
@@ -13,21 +13,22 @@ module Command
...
       end
 
       private
 
       def truncated_inspect(obj)
         obj_inspection = obj.inspect
 
-        if obj_inspection.size > 20
+        if obj_inspection.size > MAX_WORKSPACE_INSPECT_LENGTH
           obj_inspection = obj_inspection[0, 19] + "...>"
```
tompng: `MAX_WORKSPACE_INSPECT_LENGTH` needs to be used here to calculate `19`

lib/irb/command/pushws.rb
```diff
@@ -13,21 +13,22 @@ module Command
     class Workspaces < Base
       category "Workspace"
       description "Show workspaces."
+      MAX_WORKSPACE_INSPECT_LENGTH = 20
```
tompng: Could you also make the constant visibility private to avoid it being used as a public API, or let it just a local variable in this method?

```ruby
private_constant :MAX_WORKSPACE_INSPECT_LENGTH
```
or
```ruby
threshold = 20 # because only truncated_inspect use this value
if obj_inspection.size > threshold
  ...
end
```

GabrielLidenor: Sure will do it now

GabrielLidenor: @tompng all changes done. Thanks for the review.

tompng (approved):
Thank you :+1:

PR #1250, opened 2026-09-07
Refactor: Reduce duplicated code in measure.rb  

### LINKED ISSUE #1249: Refactor: Reduce duplicated code in measure.rb
### Why

-`lib/irb/command/measure.rb` - contains duplicate logic on lines 38 and 41 for displaying messages when a callback is added.

### Proposed solution

Extract the code into a private method that will only execute when a callback message is given.

GabrielLidenor:
closes #1249 

### Why

`puts "#{added[0]} is added."` was written twice in the code

### Solution

private method to encapsulate the logic and prevent duplication.


lib/irb/command/measure.rb
```diff
@@ -34,14 +34,22 @@ def execute_internal(type = nil, arg = nil)
             puts "- #{type_name}" + (arg_val ? "(#{arg_val.inspect})" : '')
           end
         when :on
-          added = IRB.set_measure_callback(arg)
-          puts "#{added[0]} is added." if added
+          measure_callback = IRB.set_measure_callback(arg)
```
tompng: It is added in `IRB.conf[:MEASURE_CALLBACKS]`, so it's an internal expression of type, callback, and measuring argument, but I'm not sure if the whole tuple is really called measure_callback.
Currently, the internal structure of set_measure_callback is only leaked to here. Passing internal structure to another method will increase the leak range.

How about something like this?
```ruby
# Leave unrelated part
added  = IRB.set_measure_callback(arg)
# display_added_message(type) needs type, will print a message,
# and nil-check is done in here, the caller site, not in display method.
display_added_message(added[0]) if added
```

GabrielLidenor: Thanks for the feedback. 

Good to know about the leak, that makes sense to me, thanks for spending time pointing that out.

I will address that now

GabrielLidenor: @tompng comment addressed. Thank you once again.

tompng (approved):

PR #1253, opened 2026-09-12
fix: irb_require matches similar names

### LINKED ISSUE #1252: Bug: irb_require incorrectly matches similarly named files
IRB's `irb_require` command can incorrectly treat a file as already loaded when another file with a similar name has been loaded.

For example, after loading `foo2.rb`, running:

```ruby
irb_require "foo"
```

### Expected behavior

`irb_require` "foo" should load `foo.rb` even when `foo2.rb` has already been loaded.

### Actual behavior

`irb_require` "foo" is treated as already loaded because `foo2.rb` is considered a match.

### Reproduction
1. Create foo.rb and foo2.rb.
2. Run irb_require "foo2".
3. Run irb_require "foo".
4. Observe that foo.rb is not loaded.

GabrielLidenor:
closes #1252 

Previously, requiring foo could be skipped if foo2.rb was already loaded. The filename matching now requires an exact match, and a regression test covers this case.


tompng (approved):

tompng:
Looks good. Thank you :+1:

In this repository and in most repositories in ruby org, you don't need to create a short-lived issue just to pair with a bug fix PR. Of course a standalone bug report issue or an issue meant for bigger discussion is always welcome :smile:
One exception: Bugfix PR to ruby/ruby repo requires bug report to bugs.ruby-lang.org for backport management.

GabrielLidenor:
> Looks good. Thank you 👍
> 
> In this repository and in most repositories in ruby org, you don't need to create a short-lived issue just to pair with a bug fix PR. Of course a standalone bug report issue or an issue meant for bigger discussion is always welcome 😄 One exception: Bugfix PR to ruby/ruby repo requires bug report to bugs.ruby-lang.org for backport management.

Oh, Ok

Thank you so much for this info. 😄

PR #1256, opened 2026-09-23
Stop aliasing load/require to ruby_load/ruby_require

### LINKED ISSUE #1255: `IRB::IrbLoader` aliases of `load`/`require` trigger the alias-fallback deprecation warning on Ruby master
### Steps to reproduce

```
docker run --rm rubylang/ruby:master ruby -w -e 'require "irb"'
```

The same warnings appear when starting `irb` itself with `ruby -w -S irb`.

### Expected behavior

`require "irb"` emits no warnings.

### Actual behavior

```
$ ruby -v
ruby 4.1.0dev (2026-09-20T00:17:23Z master 241ea171d1) +PRISM [x86_64-linux]
$ ruby -w -e 'require "irb"'
/usr/local/lib/ruby/gems/4.1.0+4/gems/irb-1.18.0/lib/irb/ext/loader.rb:16: warning: the fallback to Object for alias of 'load' in module 'IRB::IrbLoader' is deprecated and will be removed in Ruby 4.3
/usr/local/lib/ruby/gems/4.1.0+4/gems/irb-1.18.0/lib/irb/ext/loader.rb:17: warning: the fallback to Object for alias of 'require' in module 'IRB::IrbLoader' is deprecated and will be removed in Ruby 4.3
```

The warnings also appear in this repository's own CI against ruby head, for example in https://github.com/ruby/irb/actions/runs/35449203213/job/105913325835 (`rake test head`, ruby 4.1.0dev 97207692c2): `lib/irb/ext/loader.rb:16` and `:17` are reported 48 times across the test processes, although the job still passes because warnings are not turned into errors.

### Reference

`alias` in a module used to fall back to searching `Object` when the method is defined neither in the module nor in its ancestors. Ruby master now warns about this fallback, and the plan is to warn unconditionally in Ruby 4.2 and to raise `NameError` in Ruby 4.3, at which point `require "irb"` would fail.

- https://bugs.ruby-lang.org/issues/22276
- https://github.com/ruby/ruby/commit/1c0b1be076437e618a331726a5334d77c892bc6e

tompng:
Fixes #1255

Aliasing them in a module is deprecated in ruby-4.1.0dev, and aliasing wasn't necessary in that module.


lib/irb/command/load.rb
```diff
@@ -58,15 +58,15 @@ def execute_internal(file_name = nil)
...
+          return require(file_name)
         end
 
         begin
           irb_load(f = file_name + ".rb")
           $".push f
           return true
         rescue LoadError
-          return ruby_require(file_name)
+          return require(file_name)
```
st0012: nit: is this `return` and the one above necessary?

st0012 (approved):

